PR #31, opened 2022-11-14
Reduce number of properties selected on self join

AkshayN5:
Previously, we selected properties regardless of whether they were selected in the outer select clause for a self-join with multiple aliases. The fix is to select properties only if they are required in the converted SQL inner clause.


ColinKerr (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
```diff
@@ -513,14 +513,15 @@ void ECSqlExpPreparer::RemovePropertyRefs(ECSqlPrepareContext& ctx, ClassRefExp
...
+        PropertyNameExp const* propertyNameExp = parentPropertyExp->GetAsCP<PropertyNameExp>();
+        if (propertyNameExp->IsPropertyRef())
             continue;
-        if (propertyName->IsVirtualProperty())
+        if (propertyNameExp->IsVirtualProperty())
             break;
 
-        if (&classMap == &propertyName->GetPropertyMap().GetClassMap())
-            ctx.GetSelectionOptionsR().AddProperty(propertyName->GetPropertyMap());
+        const RangeClassRefExp* classRefExp = propertyNameExp->GetClassRefExp();
```
rschili, line 522: Can you quickly explain this piece? So we did not have to adjust that selected properties dictionary after all?

rschili (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
khanaffan: In `SELECT a.Prop FROM test.Foo a , test.Foo b ` we only like to select `a.Prop` but before this change` b.Prop` get selected as well as Prop is in both classes, and it was included in select column list. Instead of comparing alias qualified name a shortcut is to compare `a.Prop` resolved class ref which would be `test.Foo a` and so we just compare pointers.

rschili: Understood, sounds good, thanks!

PR #27, opened 2022-11-10
Improve testing of concurrent query manager

khanaffan:
* Splitting test into more atomic test.
* Fix a potential hang issue.

ITwinJs PR: https://github.com/iTwin/itwinjs-core/pull/4667

TS: 4682

PR #36, opened 2022-11-16
Add launch.json

pmconne:
Copied directly from old repo, removed Visualization tests.


.vscode/launch.json
```diff
@@ -0,0 +1,586 @@
...
+            "name": "(Windows) Attach",
+            "type": "cppvsdbg",
+            "request": "attach",
+            "processId": "${command:pickProcess}"
+        },
+        {
+            "name": "LNX Native BeHttpTest",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${env:OutRoot}LinuxX64/static/ProductCopy/BeHttp-Gtest/BeHttpTest",
```
calebmshafer, line 160: We should remove all of the dlls that no longer exist.

pmconne: Done.

calebmshafer (approved):

pmconne:
Linux failed with the following, nothing to do with this PR, something to do with @khanaffan's #27 which isn't even merged yet...
```
  1) concurrent query tests
       reset config:
     TypeError: conn.concurrentQueryResetConfig is not a function
      at Function.resetConfig (lib/test/ECDb.test.js:50:21)
      at Context.<anonymous> (lib/test/ECDb.test.js:65:51)

  2) concurrent query tests
       query timeout:
     TypeError: conn.concurrentQueryResetConfig is not a function
      at Function.resetConfig (lib/test/ECDb.test.js:50:21)
      at Context.<anonymous> (lib/test/ECDb.test.js:98:31)

  3) concurrent query tests
       query max memory:
     TypeError: conn.concurrentQueryResetConfig is not a function
      at Function.resetConfig (lib/test/ECDb.test.js:50:21)
      at Context.<anonymous> (lib/test/ECDb.test.js:108:31)

  4) concurrent query tests
       restart query:
     TypeError: conn.concurrentQueryResetConfig is not a function
      at Function.resetConfig (lib/test/ECDb.test.js:50:21)
      at Context.<anonymous> (lib/test/ECDb.test.js:118:31)
```

calebmshafer:
I can merge this but very concerned about how @khanaffan 's changes potentially made it into the build.

@chuckkir any idea?

chuckkir:
The JS build puts output into the source tree. The problem file is iModelJsNodeAddon\api_package\ts\lib\test\ECDb.test.js which is not part of the source so it doesn't get updated on a pull but also is not deleted. Is it possible to move it to the output tree? If not, I guess we need a pipeline step to purge the bits of source that aren't actually source (lib on down in this case).

PR #40, opened 2022-11-18
Presentation: Use related properties cache for `ContentInstancesOfSpecificClasses` specification

grigasp:
Resolves https://github.com/iTwin/itwinjs-backlog/issues/268.

When building the consolidated content descriptor, we generally start from some base class (e.g. `bis.Element`), then find all its concrete classes that have instances in the iModel, and finally create content for each of them separately. Looking up related property paths was being done as part of the last step, so we had to run the expensive related property paths lookup query as many times as there are concrete classes with instances of the given base class, which could be hundreds if not thousands. 

On the other hand, in most cases the related property paths are the same for most concrete classes, as presentation rules specifying those paths are usually applied on the base classes (`bis.Element`, `bis.GeometricElement3d`, etc.), so it makes no sense to re-run all those queries to get the same result. So instead of doing that, now we first find all related property paths that apply for given base class and then, when handling concrete classes, pick the paths that apply just for that concrete class. This saves us from running a bunch of queries and in most cases is way more performant.


grigasp:
As mentioned in https://github.com/iTwin/itwinjs-backlog/issues/268#issuecomment-1359066928, this change does improve performance in many cases, but also makes it much worse in a few others. That makes this change a no-go, I'm closing the PR.

PR #42, opened 2022-11-21
Add PRAGMA support in ECSQL and implement disqualify_type_filter and file_info pragmas

khanaffan:
* Add PRAGMA support in ECSQL
* Set PRAGMA disqualify_type_filter FOR Bis.ExternalSourceAspect


iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
```diff
@@ -1021,13 +1021,13 @@ double PragmaVal::GetDouble() const {
 //---------------------------------------------------------------------------------------
 bool PragmaVal::GetBool() const {
     if(IsBool()) {
-        return std::any_cast<bool>(m_val);
+        return m_integer;
```
pmconne, line 1024: Did you mean `m_bool`?

iModelCore/ECDb/CHANGES.md
```diff
@@ -0,0 +1,38 @@
+# Changes
+
+This document including important changes to syntax or file format.
+
+| Tables          | Version   |
+| --------------- | --------- |
+| Profile Version | `4.0.0.2` |
+| ECSQL Version   | `1.0.0.0` |
+
+## `11/22/2022`: Add `PRAGMA` support in `ECSQL`
```
pmconne, line 10: Can you add this info to NextVersion.md please so it's documented as part of a iTwin.js release?

iModelCore/ECDb/CHANGES.md
```diff
@@ -0,0 +1,38 @@
...
+
+    ```sql
+        -- Following is set by default in iModelPlatform when iModel is opened.
+        PRAGMA disqualify_type_filter=TRUE
+            FOR BisCore.ExternalSourceAspect
+    ```
+
+1. `PRAGMA file_info`: Allow to run basic information about file, ecsql, ecdb and sqlite.
+1. `PRAGMA help`: Display all the pragma supported.
+1. Added ECSQL version and set it to `1.0.0.0`. In future this version will be incremented as new feature in ECSQL is added or improved.
```
pmconne, line 38: Can you document what the four parts of this version string indicate?

khanaffan: I added documentation around each digit.

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
khanaffan: Thanks for pointing out. I have fixed this one.

iModelCore/ECDb/CHANGES.md
khanaffan: This document is intended for dev use to keep track of major changes internal or external and explain what we added or deleted and why.  Some of these cannot be put into iTwin.js release another can. I will not advertise PRAGMA stuff for external users right now. We might refine and add or delete some of these pragmas. Though there is a genuine use but it's internal.

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
```diff
@@ -0,0 +1,1092 @@
...
+};
+
+//================================================================================
+// @bsiclass PragmaFileInfo
+//================================================================================
+struct PragmaFileInfo : PragmaManager::GlobalHandler {
+    PragmaFileInfo():GlobalHandler("file_info","return file info"){}
+    ~PragmaFileInfo(){}
+    void AppendDbGuid(ECDbCR ecdb, BeJsValue row) {
+        row.appendValue() = "db_guid";
```
ColinKerr, line 161: From an iModel point of view this should be iModel Id but at the ECDb level just calling it the db guid is correct.  

This highlights that this ECSql pragma blurs the line between ECDb and iModel concepts.  Maybe it would be better to focus on ECDb file information here, I think that has the most value anyway.

For example:
- The bis schema version is a nice to have but you can already get that via ECDbMeta queries.  
- Type is an iModel interpretation of the briefcase id, any one who knows what 'standalone' and 'briefcase' mean should also be able to interpret the id to figure that out.


I think at least schema version and type should be removed.

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
```diff
@@ -0,0 +1,1092 @@
...
+        row.appendValue() =  ecdb.GetDbGuid().ToString();
+    }
+    void AppendProjectId(ECDbCR ecdb, BeJsValue row) {
+        row.appendValue() = "project_id";
```
ColinKerr, line 165: I don't particularly like exposing this as project id because it is not necessarily a projects id but I'm not sure we have a better name that fits in this scope.  iTwin id will be correct when used in the context of iTwin platform but all we can really say at the ECDb level is that this is the id of the larger context that the file belongs to.  I would suggest 'context' but we do not use 'context' else where in our API.

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
```diff
@@ -0,0 +1,1092 @@
...
+        rowSet->FreezeSchemaChanges();
+        return BE_SQLITE_READONLY;
+    }
+    static std::unique_ptr<PragmaManager::Handler> Create () { return std::make_unique<PragmaFileInfo>(); }
+};
+
+//================================================================================
+// @bsiclass PragmaECSchemaVersion
+//================================================================================
+struct PragmaECSchemaVersion : PragmaManager::SchemaHandler {
```
ColinKerr, line 333: Why do we need this?  You can just use ECDbMeta queries.  This syntax is a more compact but we don't need to make it more compact.

pmconne (approved):

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
khanaffan: i have remove the version pragma.

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
khanaffan: removed pragma for file_info for now until we figure out what useful information, we can show in it.

ColinKerr (approved):

PR #46, opened 2022-11-22
Add standard search paths after db so db has precedence

ColinKerr:


nick4598:
This PR fixes the new issues we were seeing in some of our itwinjs tests (I built Colin's branch and ran the tests locally). The problem was that I added new search paths, but they were used as locaters BEFORE the "final" locater which is to check the database itself for the requested schema. This changed the behavior of importing schemas. Colin's PR now checks the database itself for the requested schema before going on to the new search paths that we added.

c-macdonald (approved):

PR #66, opened 2022-12-09
Fix a crash in `Utf8PrintfString::CreateFromVaList` for long strings on unix 

grigasp:
Resolves https://github.com/iTwin/itwinjs-backlog/issues/559

Must make a copy of `va_list` for the second attempt, otherwise we get a crash on unix.


aruniverse:
@skirby1996 how do we backport this to the 3.5 branch?

PR #73, opened 2022-12-13
Add support for querying instance properties in ECSQL

khanaffan:
Allow access to instance property from base class.


iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
```diff
@@ -0,0 +1,270 @@
...
+            "   PROP_EXISTS(ec_classid('ts.Sub'),  'SubProp1'    ),"
+            "   PROP_EXISTS(ec_classid('ts.Sub'),  'SubProp2'    )";
+
+        ECSqlStatement stmt;   
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, sql));
+
+        ASSERT_EQ(stmt.Step(), BE_SQLITE_ROW);
+
+        int i = 0;
+        ASSERT_EQ(stmt.GetValueInt(i++), 1);
```
rschili, line 221: Could we use boolean instead of int for this? Or is it intentional?

rschili (commented):
Do we have tests for the new syntax? I only see some tests in InstanceReaderTest, but nothing for the new $ syntax?

khanaffan:
> Do we have tests for the new syntax? I only see some tests in InstanceReaderTest, but nothing for the new $ syntax?

Interesting it did write them and it seems I never pushed that last part to imodel02 branch instead i remember I stashed it to do something else and forgot. I am trying to find that git stash.

iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
khanaffan: It is the return value of SQLite function which is integer as SQLite has no boolean data type. So, my test captured return value of the function. Though the GetValueBoolean() will work as well.

khanaffan: i have added the test.

rschili (approved):

pmconne:
Are there any TypeScript changes?

I seem to recall asking before, but I do not see any comment to this effect on this PR: should this new syntax be documented in our public docs? You documented it nicely here, where no one is likely to find it.

khanaffan:
> Are there any TypeScript changes?
> 
> I seem to recall asking before, but I do not see any comment to this effect on this PR: should this new syntax be documented in our public docs? You documented it nicely here, where no one is likely to find it.

There are no typescript changes. It's just changed to ECSQL.
We do not want to advertise this publicly at this moment. Though we would like a few people to review it by using and give us feedback on it.

pmconne (approved):
Token approval for incidental changes to launch.json (consider not committing such changes).

PR #85, opened 2022-12-27
Relying on installed/pipeline node rather than repo

chuckkir:
This is the next step to removing dependence on node committed to repositories. Run the installed npm rather than the one in the repository. This only affects the building and running of the tests imodeljsnodeaddon. Also cleaned up some copyright headers.

### DIFF
--- a/iModelJsNodeAddon/api_package/ts/runTests.mke
+++ b/iModelJsNodeAddon/api_package/ts/runTests.mke
@@ -5,7 +5,7 @@
 %include mdl.mki
 
 baseDir = $(_MakeFilePath)
-npmCmd = $(BBPYTHONCMD) $(SrcRoot)thirdparty/nodejs_14_16_0/bentley/npm.py
+npmCmd = npm
 npmArg =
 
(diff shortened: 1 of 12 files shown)


iModelJsNodeAddon/api_package/ts/runTests.mke
```diff
@@ -16,7 +16,7 @@ npmArg =
         %warn *** Skipping tests as we're not running on an Apple Silicon processor ***
         %return
     %else
-        npmCmd = $(SrcRoot)thirdparty/nodejs_16_15_0/bentley/npm.sh
+        npmCmd = npm
```
pmconne, line 19: This is harmlessly redundant now.

pmconne (approved):

iModelJsNodeAddon/api_package/ts/runTests.mke
chuckkir: Turns out the npmArg is really only to npm.sh so can be removed entirely. I took a guess, but there may be some work that needs to be done when it's running on the ARM boxes.

ColinKerr (approved):

PR #86, opened 2022-12-29
Presentation: Evaluate node artifacts when taking node from cache

saskliutas:
After changes to not cache transient node data we now need to evaluate node artifacts when loading nodes from cache. Previously artifacts were evaluated only when querying nodes based on specification. This change inverts this logic to always try to load artifacts unless we now that there are no artifact rules for specification used to query nodes.

### DIFF
--- a/iModelCore/ECPresentation/Source/Hierarchies/NavNodeProviders.cpp
+++ b/iModelCore/ECPresentation/Source/Hierarchies/NavNodeProviders.cpp
@@ -230,7 +230,7 @@ void NavNodesProviderContext::Init()
     m_childNodeRule = nullptr;
     m_usedClassesListener = nullptr;
     m_pageOptions = nullptr;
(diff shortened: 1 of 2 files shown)


grigasp (commented):
Did you run perf tests with this change? I'm worried that checking node artifacts for all cached nodes might be expensive, we may need to cache this "may have artifacts" flag with the data provider / node  in such case.

saskliutas:
> Did you run perf tests with this change? I'm worried that checking node artifacts for all cached nodes might be expensive, we may need to cache this "may have artifacts" flag with the data provider / node in such case.

This should affect performance only if `ChildrenArtifacts` are used in ECExpression. Otherwise we do not evaluate children artifacts. I ran performance tests with a couple iModels using `iModelIndex` ruleset and didn't notice performance difference.

grigasp (approved):

PR #87, opened 2023-01-02
Create new iModels with WAL mode enabled

kabentley:


iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
```diff
@@ -665,100 +645,93 @@ DgnDbPtr DgnDb::OpenDgnDb(DbResult* outResult, BeFileNameCR fileName, OpenParams
...
+    }
 
-    DbResult rc = CreateNewDb(projectFile, params, params.GetGuid());
+    DbResult rc = CreateNewDb(iModelName, params, params.GetGuid());
     if (BE_SQLITE_OK != rc)
         return rc;
 
-    m_fileName = projectFile.GetNameUtf8();
+    // iModels should always use WAL journal mode
+    if (!memoryDb)
```
pmconne, line 683: Is this the only actual functional change? Way too much unrelated rename/formatting diff to sift through.

kabentley: yes, the call to EnableWalMode is new and the object of the PR. Several tests broke because they attempted to copy an open file which doesn't work in WAL mode unless you checkpoint first. That precipitated the renames so I could find all callers of create/open.

pmconne (approved):

PR #92, opened 2023-01-04
fix out-of-bounds array deref caught by asan

dassaf4:
This fix is also needed for iTwin 3.5 generation.

Details:
* Longstanding oob error caused by incomplete buffer -> bvector conversion
* ConvexHullXY closure point logic appended last point instead of first


MichaelBelousov (approved):
the repro no longer crashes after this

MichaelBelousov:
@MindaugasButkus can you confirm that the full agent run no longer hits this or does it hang because of https://github.com/iTwin/itwinjs-backlog/issues/581?

kabentley (approved):

MindaugasButkus:
> @MindaugasButkus can you confirm that the full agent run no longer hits this or does it hang because of [iTwin/itwinjs-backlog#581](https://github.com/iTwin/itwinjs-backlog/issues/581)?

The agent still hangs. However, with a specific subset of elements (to avoid the deadlock), the agent no longer crashes. It was crashing before this fix with the same subset of elements.

The https://github.com/iTwin/itwinjs-backlog/issues/581 issue remains.

PR #94, opened 2023-01-05
Schema import will fail if it requires data transform by default until flag is provided to override default behavior

khanaffan:
The default behavior previously was that schema import or profile upgrade internally *may* in some cases cause data transform in iModel. Which includes moving data from one column to another or adding new row to overflow table. With proper lock this kind of behavior may corrupt data in imodel.

Any schema change that requires any data to be transformed must happen exclusively by any one user and not by multiple users. 

itwin-core PR: https://github.com/iTwin/itwinjs-core/pull/4934


iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
```diff
@@ -519,10 +519,9 @@ enum DbResult
     BE_SQLITE_ERROR_SchemaUpgradeRequired   = (BE_SQLITE_IOERR | 15 << 24), //!< The schemas found in the database need to be upgraded.
     BE_SQLITE_ERROR_SchemaTooNew            = (BE_SQLITE_IOERR | 16 << 24), //!< The schemas found in the database are too new, and the application needs to be upgraded.
     BE_SQLITE_ERROR_SchemaTooOld            = (BE_SQLITE_IOERR | 17 << 24), //!< The schemas found in the database are too old, and the DgnDb needs to be recreated after extensive data transformations ("teleported").
-    BE_SQLITE_ERROR_SchemaLockFailed        = (BE_SQLITE_IOERR | 18 << 24), //!< Error acquiring a lock on the schemas before upgrade.
+    BE_SQLITE_ERROR_DataTransformRequired   = (BE_SQLITE_IOERR | 18 << 24), //!< Schema update need to update data.
```
pmconne, line 522: You've renamed one error code and removed another. Do you intend to update code like that in BeSQLite.ts in itwinjs-core to reflect these changes? `DbResult` is `@public`.

khanaffan: Those codes that was renamed or removed are not used anymore. Yes, I did that on itwin.js side changes.

Though i have a question, does this count as breaking change? Should i wait for 4.x?

pmconne: Please link to itwinjs-core PR that makes those changes in your PR description.
Yes, removing/renaming enum members is a breaking change.
IMO the "codes and locks" one can hang around unused for now.
Is your change in the semantics of the other one important enough to warrant renaming it?

khanaffan: No, i can add a new one though the IOERR sub flags are all used so I renamed one as they are not used. I will link the PR in minute.

khanaffan: Diff will be really bad but i did following
1. Sync BE_SQLITE_* with SQLITE_* errors and sort them. 
2. BE_SQLITE_ERROR_DataTransformRequired is assigned new value.

Made similar change to https://github.com/iTwin/itwinjs-core/pull/4934

iModelCore/ECDb/ECDb/SchemaImportContext.cpp
```diff
@@ -403,6 +413,341 @@ bool NoAdditionalForeignKeyConstraintsPolicy::IsException(ECN::NavigationECPrope
...
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+void SqlTypeDetector::Validate(RE2 const& re) {
+    if (!re.ok()) {
+        LOG.errorv("REGEX: %s\n", re.pattern().c_str());
+        LOG.errorv("ERROR: %s\n", re.error().c_str());
+    }
+    BeAssert(re.ok());
```
pmconne, line 483: Why are you both asserting and logging? Asserting means you are confident this will never ever happen, but logging suggests there's a chance it might happen.
(Triggering an assertion causes DEBUG builds to fail or hang, a perpetual annoyance).

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
```diff
@@ -609,7 +618,23 @@ BeSQLite::EC::DropSchemaResult DgnDb::DropSchema(Utf8StringCR name, bool logIssu
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-SchemaStatus DgnDb::ImportSchemas(bvector<ECSchemaCP> const& schemas)
+// SchemaStatus DgnDb::ImportSchemas(bvector<ECSchemaCP> const& schemas)
```
pmconne, line 621: delete?

pmconne (approved):

iModelCore/ECDb/ECDb/SchemaImportContext.cpp
khanaffan: The check was for only dev time check so i changed them to inline asserts. I have removed the validate method altogether.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
khanaffan: deleted

ColinKerr (approved):

rschili (approved):

aruniverse:
@iTwin/itwinjs-core-presentation @kabentley mind reviewing this?

grigasp (approved):

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -5598,6 +5602,7 @@ Utf8CP BeSQLiteLib::GetErrorName(DbResult code) {
 +---------------+---------------+---------------+---------------+---------------+------*/
 Utf8CP BeSQLiteLib::GetErrorString(DbResult rc) {
     switch(rc) {
+        case BE_SQLITE_ERROR_DataTransformRequired:         return "failed to import schema as it require data transform";
```
kabentley, line 5605: failed to import schema: data transformation required

pmconne:
```suggestion
        case BE_SQLITE_ERROR_DataTransformRequired:         return "failed to import schema: data transformation required";

```

pmconne: Please approve when you are satisfied.

iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
```diff
@@ -221,7 +221,8 @@ DbResult DgnDb::DisqualifyTypeIndexForBisCoreExternalSourceAspect()
 //--------------------------------------------------------------------------------------
 DbResult DgnDb::InitializeSchemas(Db::OpenParams const& params)
     {
-    SchemaUpgradeOptions const& schemaUpgradeOptions = ((DgnDb::OpenParams const&) params).GetSchemaUpgradeOptions();
+    const auto& inParams = ((DgnDb::OpenParams const&)params);
```
kabentley, line 224: auto const&

pmconne:
```suggestion
    auto const& inParams = ((DgnDb::OpenParams const&)params);

```

kabentley (changes requested):

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
```diff
@@ -546,8 +546,17 @@ ProfileState DgnDb::_CheckProfileVersion() const
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-BeSQLite::DbResult DgnDb::_OnBeforeProfileUpgrade()
+BeSQLite::DbResult DgnDb::_OnBeforeProfileUpgrade(Db::OpenParams const& params)
     {
+    if (GetBriefcaseId().IsBriefcase())
+        {
+        if (!((const DgnDb::OpenParams&)params).GetAllowDataTransformDuringSchemaUpdate())
```
kabentley, line 553: I now see that this subclass already existed - yuk.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
```diff
@@ -546,8 +546,17 @@ ProfileState DgnDb::_CheckProfileVersion() const
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-BeSQLite::DbResult DgnDb::_OnBeforeProfileUpgrade()
+BeSQLite::DbResult DgnDb::_OnBeforeProfileUpgrade(Db::OpenParams const& params)
     {
+    if (GetBriefcaseId().IsBriefcase())
```
kabentley, line 551: what you really want to test here is whether the iTwinId is valid - if not, that means it's a "standalone" iModel.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
```diff
@@ -557,7 +566,7 @@ BeSQLite::DbResult DgnDb::_OnBeforeProfileUpgrade()
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-BeSQLite::DbResult DgnDb::_OnAfterProfileUpgrade()
+BeSQLite::DbResult DgnDb::_OnAfterProfileUpgrade(Db::OpenParams const& params)
```
kabentley, line 569: you don't use params - why bother passing it?

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -521,8 +521,8 @@ export declare namespace IModelJsNative {
     public hasPendingTxns(): boolean;
     public hasUnsavedChanges(): boolean;
     public importFunctionalSchema(): DbResult;
-    public importSchemas(schemaFileNames: string[]): DbResult;
-    public importXmlSchemas(serializedXmlSchemas: string[]): DbResult;
+    public importSchemas(schemaFileNames: string[], allowDataTransformDuringSchemaUpdate: boolean): DbResult;
```
kabentley, line 524: if you're going to change the signature of these methods, why not a new "options" arg?

khanaffan: Good point. I will instead use options.

iModelCore/ECDb/ECDb/ECDb.mke
```diff
@@ -484,7 +484,8 @@ LINKER_LIBRARIES            = $(ContextSubpartsLibs)$(BeSQLiteLib) \
                               $(ContextSubpartsLibs)$(libprefix)iTwinXml$(libext) \
                               $(ContextSubpartsLibs)$(libprefix)BeLibxml2$(libext) \
                               $(ContextSubpartsLibs)$(libprefix)iTwinJsonCpp$(stlibext) \
-                              $(ContextSubpartsLibs)$(libprefix)iTwinGeomSerialization$(libext)
+                              $(ContextSubpartsLibs)$(libprefix)iTwinGeomSerialization$(libext) \
+                              $(ContextSubPartsLibs)$(stlibprefix)iTwin_google_re2$(stlibext)
```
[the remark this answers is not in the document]
khanaffan: I have debug code to detect if any code change to mapping has caused any data. I use REGEX library to differentiate between changes to ec_* table vs changes to data tables and then fail schema import. The check is only enabled during debug build.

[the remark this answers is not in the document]
kabentley: can you use [std::regex](https://en.cppreference.com/w/cpp/regex)? I would like lessen/remove our dependency on the google regex library.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
kabentley: actually, i think you should remove this test entirely. If someone wants to use standalone iModels, they should pass `true` for `schemaLockHeld`

iModelCore/ECDb/ECDb/ECDb.mke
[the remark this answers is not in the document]
khanaffan: Sure, I can try, `std::regex` seem to not support some basic things like multiline (^: start of line, $: end of line). But 
Looking at [doc ](https://en.cppreference.com/w/cpp/regex/syntax_option_type)now it seems c++17 did introduce it.  My case in ECDb is simple but use multiline so I can test it.

>multiline (C++17) | Specifies that ^ shall match the beginning of a line and $ shall match the end of a line if the ECMAScript engine is selected.

iModelCore/ECDb/PublicAPI/ECDb/ECDb.h
```diff
@@ -144,6 +144,39 @@ struct ECSqlConfig {
 struct EXPORT_VTABLE_ATTRIBUTE ECDb : Db
 {
 public:
+    //=======================================================================================
+    // @bsiclass
+    //=======================================================================================
```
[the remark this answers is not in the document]
khanaffan: I removed `ECDb::OpenParam` and added flag to `Db::OpenParam`.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
khanaffan: Fixed

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
khanaffan: I removed it.

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
```diff
@@ -2487,6 +2499,7 @@ struct EXPORT_VTABLE_ATTRIBUTE Db : NonCopyableClass
         ProfileUpgradeOptions m_profileUpgradeOptions = ProfileUpgradeOptions::None;
         bool m_rawSQLite = false;
         bool m_fromContainer = false;
+        bool m_allowDataTransformDuringSchemaUpdate = false;
```
kabentley, line 2502: this should be m_schemaLockHeld. If we could test for that fact in C++ it wouldn't be necessary, but this isn't a programmer choice, it's a fact that our TS layer has to set correctly. In other words you don't turn on this flag, you acquire the schema lock and *that* turns on this flag.

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/VersionCompare_Test.cpp
```diff
@@ -137,7 +137,7 @@ void VersionCompareTestFixture::SetUpTestCase()
     m_db->ResetBriefcaseId(BeBriefcaseId(2));
 
     // Import DgnPlatformTest schema containing the test element and such
-    SchemaStatus schemaStatus = DgnPlatformTestDomain::GetDomain().ImportSchema(*m_db);
+    SchemaStatus schemaStatus = DgnPlatformTestDomain::GetDomain().ImportSchema(*m_db, SchemaManager::SchemaImportOptions::AllowDataTransformDuringSchemaUpgrade);
```
kabentley, line 140: seems like you could make `SchemaManager::SchemaImportOptions::AllowDataTransformDuringSchemaUpgrade` the default and then you wouldn't have to change all these tests. The only call that really matters is the one from TS, and it won't use the default.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -428,6 +428,10 @@ export declare namespace IModelJsNative {
     readonly customAttributes?: Array<{ [value: string]: any }>;
   }
 
+  interface SchemaImportOptions {
+    readonly schemaLockHeld: boolean;
```
kabentley, line 432: should be optional, default to false if not present.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -521,8 +525,8 @@ export declare namespace IModelJsNative {
     public hasPendingTxns(): boolean;
     public hasUnsavedChanges(): boolean;
     public importFunctionalSchema(): DbResult;
-    public importSchemas(schemaFileNames: string[]): DbResult;
-    public importXmlSchemas(serializedXmlSchemas: string[]): DbResult;
+    public importSchemas(schemaFileNames: string[], options: SchemaImportOptions): DbResult;
```
kabentley, line 528: make `options` optional too

iModelCore/ECDb/ECDb/ECDb.mke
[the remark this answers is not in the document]
khanaffan: So `std::regex_constants::syntax_option_type::multiline` is still not supported in MSVC or CLANG. The doc i passed seems to be wrong.

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
khanaffan: fixed

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
khanaffan: I have changed the name in c++ layer.

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/VersionCompare_Test.cpp
khanaffan: I wanted to be not set by default in ECDb. I guess I will continue that in DgnDb. I have now set it by default.

kabentley:
I don't see any tests of `schemaLockHeld` logic in the TS tests here. Shouldn't there be some?

khanaffan:
> I don't see any tests of `schemaLockHeld` logic in the TS tests here. Shouldn't there be some?

I am renaming the test, but it's called "import schema" in `DgnDb.test.ts`.

kabentley (approved):

PR #95, opened 2023-01-05
cherry-pick: fix out-of-bounds array deref caught by asan

dassaf4:
Cherry-pick of #92 

Details:
* Longstanding oob error caused by incomplete buffer -> bvector conversion
* ConvexHullXY closure point logic appended last point instead of first

### DIFF
--- a/iModelCore/GeomLibs/geom/src/funcs/ConvexHull.cpp
+++ b/iModelCore/GeomLibs/geom/src/funcs/ConvexHull.cpp
@@ -205,7 +205,7 @@ bool includeClosurePoint
                 }
             if (includeClosurePoint && hullPoints.size() > 1)
                 {
-                DPoint3d xyz = hullPoints.back ();
+                DPoint3d xyz = hullPoints.front();
                 hullPoints.push_back (xyz);
                 }
             }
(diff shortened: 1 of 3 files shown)


MichaelBelousov (approved):

EarlinLutz (changes requested):
This code was probably ported from old-style pre-allocated arrays that were certain to have sufficient size.

I think xyz1.push_back (xyz1[0]) still has potential problems if the size and capacity are already equal -- xyz1[0] should be copied into a local DPoint3d before being pushed.  But I'll be delighted if push_back is smart enough to not deallocate the input-time memory block before copying to a newly allocated block.

dassaf4:
> I think xyz1.push_back (xyz1[0]) still has potential problems if the size and capacity are already equal...

Great point! I changed the push_back to use a local copy of the point, and also added a test to confirm that the compiler appends a valid point on push_back realloc just in case, as I'm sure I've seen (and used!) the anti-pattern elsewhere.

EarlinLutz (approved):

MichaelBelousov:
latest changes still pass the reported repro

PR #99, opened 2023-01-09
Validate that CodeScope is the ElementId of an existing element

kabentley:
The value of `CodeScope` *must* be a valid ElementId, but we did not enforce that. Throw an exception if the supplied value is illegal.

On insert and update, the value supplied from JavaScript may be a `FederationgGuid`. If so, look up the element and use its ElementId.

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
@@ -557,7 +557,11 @@ void DgnCode::ToJson(BeJsValue val) const {
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
 DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
-    DgnCode val;
+    DgnCode val = CreateEmpty();
+    val.m_value = value[json_value()].asString();
+    if (val.m_value.empty())
+        return val; // if the value is empty, don't use any of the other parameters
+
     auto specJson = value[json_spec()];
     val.m_specId = CodeSpecId(specJson.asUInt64());
     if (!val.m_specId.IsValid()) {
@@ -566,7 +570,26 @@ DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
         auto specPtr = db.CodeSpecs().GetCodeSpec(specJson.asCString());
         val.m_specId = specPtr.IsValid() ? specPtr->GetCodeSpecId() : CodeSpec::GetNullCodeSpecId();
     }
-    val.m_scope = value[json_scope()].asString();
-    val.m_value = value[json_value()].asString();
+
+    // the code scope must refer to an existing element. The json may have an ElementId or FederationGuid
+    auto scopeJson = value[json_scope()];
+    auto scopeStr = scopeJson.asCString();
+
+    // Note: this is necessary since GetId64 parses strings that start with number successfully and can get confused with Guids.
+    // ElementIds in JSON must start with "0x"
+    auto validId = scopeStr && (scopeStr[0] == '0' && (scopeStr[1] == 'X' || scopeStr[1] == 'x'));
+
+    auto scopeId = scopeJson.GetId64<DgnElementId>();
+    if (!validId || !db.Elements().ElementExists(scopeId)) {
+        BeGuid scopeFederationGuid;
+        DgnElementCPtr scopeEl;
+        if (BentleyStatus::SUCCESS == scopeFederationGuid.FromString(scopeJson.asCString()))
+            scopeEl = db.Elements().QueryElementByFederationGuid(scopeFederationGuid);
+        if (!scopeEl.IsValid())
+            throw std::invalid_argument("invalid code scope");// the value didn't refer to an element either by ElementId or FederationGuid
+
+        scopeId = scopeEl->GetElementId();
+    }
+    val.m_scope = scopeId.ToHexStr();
     return val;
 }
--- a/iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
@@ -553,12 +553,16 @@ void DgnCode::ToJson(BeJsValue val) const {
     val[json_value()] = m_value.GetUtf8();
 }
 
-/*---------------------------------------------------------------------------------**//**
-* @bsimethod
-+---------------+---------------+---------------+---------------+---------------+------*/
-DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
-    DgnCode val = CreateEmpty();
+/**
+ * Convert a JSON CodeScopeProps to a DgnCode.
+ * @note the specId may be supplied as either an ElementId or a classname.
+ * @note if `validateScope` is true, the scopeId value may either be an ElementId or FederationGuid of the scope element.
+ * If it is not valid, or does not refer to an existing element, an exception is thrown. If `validateScope` is false, the scopeId value is saved verbatim.
+ */
+DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db, bool validateScope) {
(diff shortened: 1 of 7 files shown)


iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
```diff
@@ -566,7 +568,26 @@ DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
...
+    // the code scope must refer to an existing element. The json may have an ElementId or FederationGuid
+    auto scopeJson = value[json_scope()];
+    auto scopeStr = scopeJson.asCString();
+
+    // Note: this is necessary since GetId64 parses strings that start with number successfully and can get confused with Guids.
+    // ElementIds in JSON must start with "0x"
+    auto validId = scopeStr && (scopeStr[0] == '0' && (scopeStr[1] == 'X' || scopeStr[1] == 'x'));
+
+    auto scopeId = scopeJson.GetId64<DgnElementId>();
+    if (!validId || !db.Elements().ElementExists(scopeId)) {
```
pmconne, line 581: Did you decide not to special-case "0x1" after all?

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
```diff
@@ -566,7 +568,26 @@ DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
...
-    val.m_scope = value[json_scope()].asString();
-    val.m_value = value[json_value()].asString();
+
+    // the code scope must refer to an existing element. The json may have an ElementId or FederationGuid
+    auto scopeJson = value[json_scope()];
+    auto scopeStr = scopeJson.asCString();
+
+    // Note: this is necessary since GetId64 parses strings that start with number successfully and can get confused with Guids.
+    // ElementIds in JSON must start with "0x"
+    auto validId = scopeStr && (scopeStr[0] == '0' && (scopeStr[1] == 'X' || scopeStr[1] == 'x'));
```
pmconne, line 578: You are permitting uppercase X in Ids? We forbid it elsewhere.

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
```diff
@@ -566,7 +568,26 @@ DgnCode DgnCode::FromJson(BeJsConst value, DgnDbCR db) {
...
+
+    auto scopeId = scopeJson.GetId64<DgnElementId>();
+    if (!validId || !db.Elements().ElementExists(scopeId)) {
+        BeGuid scopeFederationGuid;
+        DgnElementCPtr scopeEl;
+        if (BentleyStatus::SUCCESS == scopeFederationGuid.FromString(scopeJson.asCString()))
+            scopeEl = db.Elements().QueryElementByFederationGuid(scopeFederationGuid);
```
pmconne, line 585: You expressed some concerns about additional overhead imposed by verifying scope element exists. Are you ok with:
- Querying for and loading entire element vs just its federation Guid; and
- If caller supplies a valid but non-existent element Id, attempting fruitlessly to look up an element that has that Id as its federation Guid?

pmconne (commented):
Are you concerned about any iModels that may exist with invalid code scopes beginning to throw exceptions in certain contexts making them unusable?
Would it be safer to only throw from `DgnCode::FromJson` during insert and update?

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
kabentley: That code comes from BeJsConst::GetUInt64. Should we not allow it there?

kabentley:
> Are you concerned about any iModels that may exist with invalid code scopes beginning to throw exceptions in certain contexts making them unusable? Would it be safer to only throw from `DgnCode::FromJson` during insert and update?

Currently `DgnCode::FromJson` is only called from insert/update. But this logic should apply to any new use case of it too.

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
pmconne: Meh, I don't know. BeInt64Id uses BeStringUtilities::HasHexPrefix which allows uppercase too.
TypeScript only permits lowercase.

kabentley: that's why i used `scopeId.ToHexStr()` so that it will always be "correct" in the iModel.

pmconne:
> No, `DgnCode::FromJson` is not called from GetElement/GetModel.

[Yes, it is](https://github.com/iTwin/imodel-native/blob/137a73e655d7a1336f181a89c3f93a216b3143b7/iModelJsNodeAddon/JsInteropDgnDb.cpp#L317).

[This caller](https://github.com/iTwin/itwinjs-core/blob/b99c65cdddf8b8e4975cff218c42b813864ab48e/core/backend/src/IModelDb.ts#L1578) expects any error thrown by the addon to have an `errorNumber`.

kabentley:
> > No, `DgnCode::FromJson` is not called from GetElement/GetModel.
> 
> [Yes, it is](https://github.com/iTwin/imodel-native/blob/137a73e655d7a1336f181a89c3f93a216b3143b7/iModelJsNodeAddon/JsInteropDgnDb.cpp#L317).
> 
> [This caller](https://github.com/iTwin/itwinjs-core/blob/b99c65cdddf8b8e4975cff218c42b813864ab48e/core/backend/src/IModelDb.ts#L1578) expects any error thrown by the addon to have an `errorNumber`.

I thought you were talking about `DgnElements::GetElement`  (that doesn't/can't use JSON and would be bad to throw). 

If you attempt to load an element/model from JavaScript via Code and you supply an invalid scope, you will now get an exception that tells you that's what went wrong. I guess i should put some value (NotFound?) for errorNumber.

pmconne:
> If you attempt to load an element/model from JavaScript via Code and you supply an invalid scope, you will now get an exception

That is why I'm concerned. We know that iModels can exist that contain elements with invalid scopes, and even after this PR it will still be possible to create new ones (by, e.g., deleting the scope element). So, it will be possible, for example, for an app to query the iIModel to obtain an invalid code, then attempt to obtain the properties of the element with that code and either get back no properties or produce an exception, depending on which API it calls.

Other ways to create invalid code scopes, I think, include running the transformer and omitting the code scope element, and connectors using the native API.

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
kabentley: > Querying for and loading entire element vs just its federation Guid;
This is rare and not worth worrying about.

>If caller supplies a valid but non-existent element Id, attempting fruitlessly to look up an element that has that Id as its federation Guid?
`scopeFederationGuid.FromString` will fail first.

kabentley:
> That is why I'm concerned. We know that iModels can exist that contain elements with invalid scopes, and even after this PR it will still be possible to create new ones (by, e.g., deleting the scope element). So, it will be possible, for example, for an app to query the iIModel to obtain an invalid code, then attempt to obtain the properties of the element with that code and either get back no properties or produce an exception, depending on which API it calls.
> 
> Other ways to create invalid code scopes, I think, include running the transformer and omitting the code scope element, and connectors using the native API.

That's what I thought you were talking about originally. Nothing calls DgnCode::FromJson using data read from an iModel. Do I misunderstand your concern?

iModelCore/iModelPlatform/DgnCore/CodeSpec.cpp
kabentley: i added it to `ElementExits`

kabentley:
> IMO validation should happen on insert/update. It should not happen when constructing a code from JSON for other purposes.
I agree. Done.

pmconne (approved):

pmconne:
@kabentley what's this waiting on?

PR #103, opened 2023-01-10
don't use getenv_s in conditional debug code since it doesn't exist on linux

MichaelBelousov:
- affects optional debug only code
- fixes debug building on linux (`getenv_s` is a microsoft extension)
- assumes the security benefits of `getenv_s` aren't necessary since this code doesn't exist in release/NDEBUG versions

### DIFF
--- a/iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
+++ b/iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
@@ -1007,18 +1007,13 @@ BentleyStatus MainSchemaManager::ImportSchemas(SchemaImportContext& ctx, bvector
     In Debug builds, the environment variable can be set to a directory path to
     dump existing and incoming schemas to for every ImportSchemas call.
     */
-    Utf8CP envVarName = "ECDB_SCHEMAIMPORT_DUMP_TO";
-    size_t requiredSize;
-    if (getenv_s(&requiredSize, NULL, 0, envVarName) == 0 && requiredSize != 0)
+    BeFileName dumpSchemaDir;
+    const char* const schemaImportDumpTo = getenv("ECDB_SCHEMAIMPORT_DUMP_TO");
+    if (schemaImportDumpTo != NULL)
         {
-        BeFileName dumpSchemaDir;
-        std::vector<char> chars(requiredSize);
-        if (getenv_s(&requiredSize, chars.data(), requiredSize, envVarName) == 0)
-            {
-            dumpSchemaDir.AssignUtf8(chars.data());
-            DumpSchemasToFile(dumpSchemaDir, m_ecdb.Schemas().GetSchemas(true), "existing");
-            DumpSchemasToFile(dumpSchemaDir, schemas, "incoming");
-            }
+        dumpSchemaDir.AssignUtf8(schemaImportDumpTo);
+        DumpSchemasToFile(dumpSchemaDir, m_ecdb.Schemas().GetSchemas(true), "existing");
+        DumpSchemasToFile(dumpSchemaDir, schemas, "incoming");
         }
     #endif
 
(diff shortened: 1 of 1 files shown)


khanaffan (approved):

rschili (approved):
So, the original reason I went with getenv_s was that some warning popped up during my build, saying getenv is not to be used, as it's not safe.
I agree it's a non-issue here, so I'm fine moving to plain getenv().
However, I know those _s methods usually were introduced in microsoft extensions, but a whole bunch of them made it into the official c++ standard and are supported on any platform. According to cppreference.com, getenv_s should be part of the c++ 11 standard. https://en.cppreference.com/w/c/program/getenv
So I really don't understand why our builds on other platforms do not support it. I was told about this and was going to include a fix with a different PR where I basically wrapped this in a #ifdef WINDOWS, but your fix is fine, too.

MichaelBelousov:
> However, I know those _s methods usually were introduced in microsoft extensions, but a whole bunch of them made it into the official c++ standard and are supported on any platform. According to cppreference.com, getenv_s should be part of the c++ 11 standard. https://en.cppreference.com/w/c/program/getenv

I saw the same thing, and I struggled researching this. But it looks like that's erroneous on cppreference's part. I don't see getenv_s listed in any other source... And it's definitely not on my linux machine's `/usr/include/stdlib.h`. There is a `secure_getenv` however, but with a different meaning. I'll do additional research

MichaelBelousov:
ok here is an update. It's in the standard behind a special extension preprocessor declaration that must be made. See page 624 of this free draft: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf (the actual standard is behind a paywall.
I believe I saw this last week when I first encountered this issue and researched. It isn't mentioned in a lot of places, and it's interesting that I only really see cppreference showing it. I still think we should just use regular `getenv` in this case

rschili:
Agreed. I wonder what the appropriate way to do this would be if we needed it in release builds. We do have some spots where we use env vars. The code is inconsistent there, we don't have a central place in a Be library which gets an env var, we should probably put something.

I noticed I was the first to pick getenv_s, but that cppreference page made me feel quite confident about it.

MichaelBelousov:
hmmm perhaps I should switch back to using getenv_s and just define the necessary preprocessor macro before includes in this translation unit, as an example for when we do eventually use `getenv_s` for something else... I will close this unless some disagrees and make a new PR just adding the necessary header and making sure it works on my linux machine

MichaelBelousov:
back again... sorry. Looks like GNU libc does not support the `__STDC_LIB_EXT1__` which explains why `grep -R getenv_s /usr/include` didn't find it for me. More details on why are here: https://stackoverflow.com/questions/47867130/stdc-lib-ext1-availability-in-gcc-and-clang

So in practice it's non-portable. I will revive using `getenv` instead with this PR and then and we can decide when we write a generic `getenv` bentley utility whether we want to use `getenv_s` on windows only imo.

MichaelBelousov:
@rschili going to wait on your review even though the checks are passing, in case you disagree with my reasoning

rschili:
> @rschili going to wait on your review even though the checks are passing, in case you disagree with my reasoning

@MichaelBelousov I'm good with the decision. And I'm grateful you took care of this. Thanks a lot.

pmconne:
@rschili why don't you simply add (to be uncommented when you are doing this kind of debugging)
`// #define ECDB_SCHEMAIMPORT_DUMP_TO "c:\rob-schilli\whatever\"`
instead of reading the environment.

rschili:
It was meant to be used with connectors where rebuilding is a pain.
This change has been live for months, but Mike Belousov changed it from getenv_s into getenv yesterday.

-Rob
________________________________

rschili:
I‘m tempted to just drop this whole thing. I was trying to do some good, saving people time while diagnosing import issues, but I dont want this to turn into a time consuming burden. I‘m already in bed, maybe Mike can push another update today, if not I‘ll take care of it tomorrow.

MichaelBelousov:
hey sorry I was also out, got back as soon as possible to fix this. #108 should be a good enough short-term answer imo.

PR #104, opened 2023-01-11
use IMMEDIATE mode for the defaultTxn in writeable WAL mode

kabentley:


khanaffan (approved):

pmconne:
Did you intentionally alter test.bim and if so in what way?

kabentley:
> Did you intentionally alter test.bim and if so in what way?

Yes, I had to manually enable WAL mode.

pmconne (approved):

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnStyles_Test.cpp
```diff
@@ -437,6 +437,7 @@ TEST_F(DgnLineStyleTest, InsertLineCodeComponentAsJson)
 
     ASSERT_TRUE(DbResult::BE_SQLITE_OK == m_db->SaveChanges());
     seedFilePath = m_db->GetDbFileName();
+    m_db->CloseDb();
```
kabentley, line 440: FYI: All of these were places where we inadvertently forgot to close the (writeable) db. We then open it again later in the test. That used to "succeed" because the first writeable connection saved its changes and therefore did not have a txn open, so the second writer "stole" its lock. Now, the second open fails.

grigasp (approved):

PR #105, opened 2023-01-11
fix broken custom attribute property serialization in `getECClassMetaData`

MichaelBelousov:
the baseaccess string argument to `JsonECInstanceWriter::WritePrimitiveValue` was being erroneously used always, when it's only meant for structs. This meant properties were being accessed like `PropName.PropName`, when they should have just been `PropName`, so nothing was written to the JSON.

closes https://github.com/iTwin/itwinjs-core/issues/4624

Misc:
also `chmod +x installnativeplatformLinux.sh` so I don't need to keep typing `bash` everytime I use it


iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -323,4 +323,27 @@ describe("basic tests", () => {
       }
     });
   });
+
+  it.only("testGetECClassMetaData custom attributes", async () => {
```
pmconne, line 327: Remove `.only`

pmconne (changes requested):
Looks right to me but @khanaffan please review.

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
MichaelBelousov: That's a face palm. Will do

MichaelBelousov:
I just started running backend js tests and am running into a number of issues. already hit both a hang and a separate abort on iModel close, but I don't know if it's related to my small change. I will revive this issue after I have confirmed whether those occur on master and submitted issues to the backlog for them if so

MichaelBelousov:
I still have issues with WAL mode (https://github.com/iTwin/itwinjs-backlog/issues/607), but rebasing this change on the pre-wal mode commit (with a cherry-pick to fix a crash), I pass the same tests as without these changes. That is, I fail all the geographical projection related tests and the objectvault test.
I don't believe this change has anything to do with any of that so I'm reopening

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
MichaelBelousov: done

khanaffan (approved):

pmconne (approved):

PR #106, opened 2023-01-12
Add codeSpecKind Json property. Add Update support to CodeSpecs.

diegoalexdiaz:


kabentley:
I don't think this PR is necessary. All of this should be done in TypeScript.

kabentley:
@diegoalexdiaz I think you should abandon this PR.

PR #108, opened 2023-01-12
make ecdb_schemaimport_dump env code opt-in

MichaelBelousov:
- make ecdb schemaimport dump env code opt-in, and add note about how to handle it cross-platform in the future
- fixes issue with debug builds on windows caused by #103


pmconne (commented):
Shouldn't you also revert #103? @rschili said this is only useful for connectors, which generally only run on Windows, where `getenv` produces an error.

MichaelBelousov:
> Shouldn't you also revert #103? @rschili said this is only useful for connectors, which generally only run on Windows, where `getenv` produces an error.

Will do

MichaelBelousov:
this is now a pure revert commit followed by the same `#define` change

rschili (approved):

rschili:
I approved and re-queued the build, looks like we may have a flaky test in there.
Actually, all of this is really half-baked. Connectors will typically produce a lot of assertion hits when running on an imodel-native debug build (which we should fix at some point, but the reality is they do produce them), so the dump feature has limited usefulness right now... I still wanted to keep it around, because I already re-implemented and then deleted it like 5 times in the past :-)

pmconne:
> Connectors will typically produce a lot of assertion hits when running on an imodel-native debug build

Why not remove the check on `NDEBUG` then?

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -1002,18 +1002,24 @@ void DumpSchemasToFile(BeFileName const& parentDirectory, bvector<ECSchemaCP> co
 //+---------------+---------------+---------------+---------------+---------------+------
 BentleyStatus MainSchemaManager::ImportSchemas(SchemaImportContext& ctx, bvector<ECSchemaCP> const& schemas, SchemaImportToken const* schemaImportToken) const
     {
-    #ifndef NDEBUG
+    //#define ALLOW_ECDB_SCHEMAIMPORT_DUMP
+    #if !defined(NDEBUG) && defined(ALLOW_ECDB_SCHEMAIMPORT_DUMP)
```
pmconne, line 1006:
```suggestion
    #if defined(ALLOW_ECDB_SCHEMAIMPORT_DUMP)
```
Now you can enable this feature without having to deal with slower debug performance and annoying assertions.

rschili: Yeah, it's better this way, let's do as Paul suggests.

MichaelBelousov:
> I approved and re-queued the build, looks like we may have a flaky test in there. 

I added a flaky test issue for that test last night

PR #109, opened 2023-01-13
Normal maps in tiles

pmconne:
[itwinjs-core PR](https://github.com/iTwin/itwinjs-core/pull/4922)
imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/45


pmconne:
@MichaelBelousov do you understand why [RenderingAsset::Relocate](https://github.com/iTwin/imodel-native/blob/6280148644cbb7db8294c44287f8cbf1f7539e45/iModelCore/iModelPlatform/DgnCore/RenderMaterial.cpp#L239) returns `ERROR` if there's no pattern map?
A pattern map is not required. Does `ERROR` mean something to the caller? Is it ignored?

markschlosseratbentley (approved):

MichaelBelousov:
> @MichaelBelousov do you understand why [RenderingAsset::Relocate](https://github.com/iTwin/imodel-native/blob/6280148644cbb7db8294c44287f8cbf1f7539e45/iModelCore/iModelPlatform/DgnCore/RenderMaterial.cpp#L239) returns `ERROR` if there's no pattern map? A pattern map is not required. Does `ERROR` mean something to the caller? Is it ignored?

I have no idea. I don't believe we cover that in the core tests.

PR #125, opened 2023-01-23
Presentation: Should not support filtering nodes built with `parent` symbol

grigasp:
Note: this likely requires regenerating presentation-full-stack-test snapshots. Will do that a bit later, will then associate this PR with snapshot update PR.


saskliutas (approved):

grigasp:
Confirmed no related `itwinjs-core` changes are needed.

PR #126, opened 2023-01-24
Fixed ECSQL scanner mem leak detected by Valgrind 

khanaffan:
Valgrind found a memory leak in ECSQL parser which was caused due to some`OSQLParseNode` not freed.

Tested with `CLANG_SANTIZER=leak` with `CLANG13 `which seems more accurate and performant than valgrind.


aruniverse:
Hoping to pull this into the next addon. Can we get this reviewed?
@kabentley @rschili @calebmshafer @ColinKerr

ColinKerr (approved):

PR #136, opened 2023-01-31
Added support for Greatest and Least functions

RohitPtnkr1996:


iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -22,6 +23,18 @@ std::unique_ptr<Exp> ECSqlParser::Parse(ECDbCR ecdb, Utf8CP ecsql, IssueDataSour
 
     if (parseTree == nullptr || !error.empty()) {
         Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSQL, "Failed to parse ECSQL '%s': %s", ecsql, error.c_str());
+
+        // Check if Min/Max function with multiple args was called. 
+        std::cmatch maxRegexMatches;
+        std::regex_search(ecsql, maxRegexMatches, std::regex(R"rx([(=,\s]MAX\(\d+,.*\))rx", std::regex_constants::icase));
```
khanaffan, line 29: What about if I do `SELECT 'MAX(2,3)' AS str`  ? Or a comment in ECSQL container similar string? Also if ecsql look like this
```sql
select MAX
( 3, 
4)
FROM test.Goo
```

The solution is flawed. 

Use parse tree to find how many parameter MAX or MIN have.

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparedStatement.cpp
```diff
@@ -639,6 +658,10 @@ ECSqlStatus ECSqlInsertPreparedStatement::PrepareLeafStatements(PrepareInfo& pre
                       valuesECSqlClause.c_str());
 
         ECSqlParser parser;
+        // Check if Greatest/Least was changed to Max/Min in the un-prepared statement
+        // If so, it needs to be reset to Greatest/Least in the prepared statement to succeed parsing
+        ResetMaxOrMinFunctionInSql(ecsql);
```
khanaffan, line 663: Do not do that. Read my previous comment.

iModelCore/ECDb/Tests/NonPublished/ECDbTests.cpp
```diff
@@ -561,5 +561,159 @@ TEST_F(ECDbTestFixture, NewFileECDbProfileVersion)
...
+
+        { __LINE__, "SELECT LEAST()", ECSqlStatus::InvalidECSql, { 0 }, "Preparing the ECSQL 'SELECT LEAST()' failed. Underlying SQLite statement failed to prepare: BE_SQLITE_ERROR wrong number of arguments to function MIN() (BE_SQLITE_ERROR) [SQL: SELECT MIN()]" },
+        { __LINE__, "SELECT LEAST(4)", ECSqlStatus::Success, { 4 }, "" },
+        { __LINE__, "SELECT LEAST(4, 2, 1, 5, 3)", ECSqlStatus::Success, { 1 }, "" },
+        { __LINE__, "SELECT LEAST(4, 2, 1, 5, 3)", ECSqlStatus::Success, { 1 }, "" },
+
+        { __LINE__, "SELECT MAX()", ECSqlStatus::InvalidECSql, { 0 }, "Failed to parse ECSQL 'SELECT MAX()': syntax error" },
+        { __LINE__, "SELECT MAX(4)", ECSqlStatus::Success, { 4 }, "" },
+        { __LINE__, "SELECT MAX(4, 2, 1, 5, 3)", ECSqlStatus::InvalidECSql, { 0 }, "MAX function with multiple arguments isn't supported. Please use GREATEST instead." },
+        { __LINE__, "SELECT MAX(4, 2, 1, 5, 3)", ECSqlStatus::InvalidECSql, { 0 }, "MAX function with multiple arguments isn't supported. Please use GREATEST instead." },
```
khanaffan, line 597: Use More complex examples with types and actual column and compute expressions in both SELECT and WHERE clause.

```
SELECT GREATEST(
       'a', 
       'b',
       'c')
```

```
SELECT LEAST(
       'a', 
       'b',
       'c')
```

```
SELECT LEAST(
       'a', 
       'b',
       'c')
```

```
SELECT MAX(MAX(3), MAX(4))
```

khanaffan (changes requested):

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
RohitPtnkr1996: The Min/Max functions have a grammar rule defined only for single args. So, if I call them with multiple args, the parse tree generation itself fails giving a syntax error.
That's why I had to resort to regex.

khanaffan: You can remove rules for Max/Min from grammar. It will be then detected as regular function with n param.

khanaffan (approved):

PR #137, opened 2023-01-31
allow custom schema export file destinations and import locators (prepare for transform long schema names)

MichaelBelousov:
**linked PRs**
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/5011/files

I plan on backporting this to 3.6 as a bugfix for the transformation service to consume

## What

- do not use an `BeXmlWriter` that could be null if an i/o error occurs
- add the option to supply a custom file name when exporting a schema
- add the option to supply a custom schema locator (using SchemaLocaterCallback and `ECSchemaXmlContext`) during calls to `nativeDb.importSchemas()` which can use that locater to resolve references to schemas.

## Why

For handling long schema names on the transformer.

linux + mac common file systems usually have a 255 byte limit to path segments and therefore to filenames. This means that exporting a schema with a name of length > `255 - '.ecschema.xml'.length` could fail with an error on those operating systems. Transformation often exports schemas from the source and imports them into the target, which means certain transformations of models with certain schemas fail on linux agents.

This is especially noticeable in certain dgn user-defined schemas if generated from a non-ascii    label, a hex encoded utf16 string with `____` between every wide character is used to construct the schema name. This gets too long for 255 bytes very quickly. That is a summary to the best of my knowledge but @ColinKerr might know more about what/why for this. 

In the accompanying JavaScript (itwinjs-core PR](https://github.com/iTwin/itwinjs-core/pull/5011/files, these two additions are used to write a shorter named file while exporting schemas during a transformation, and when the target imports them, it can still find the renamed schemas.


MichaelBelousov:
closing to investigate transforming schemas without touching the disk

khanaffan (approved):

ColinKerr (approved):

PR #145, opened 2023-02-06
Consume New Snapshots in iModel Evolution Tests

naveedkhan8067:
As part of iModel evolution tests, adding the test runner to verify new snapshots which were generated after the recent profile upgrade.


pmconne (commented):
I assume this is fine. @ColinKerr?

pmconne (approved):

PR #150, opened 2023-02-09
Presentation: Fix hierarchy level filtering with related instances

grigasp:
When creating the filter definition, we only set either target class or relationship alias, leaving the other one empty. The query builder expects both to be set, so we have to make sure that's the case.


saskliutas (approved):

grigasp:
@mergifyio backport release/3.x

PR #154, opened 2023-02-10
Fix property remapping for added base class

rschili:
The existing statement to detect properties that need remapping when moving a class within the hierarchy was flawed.
- It didn't correctly detect mixins props on the migrating class
- It remapped too many properties as it didn't exclude columns from moved classes on the resident classes side

I recreated the statement from scratch and added tests for the missing scenarios.
Also I left some diagnostic code that can be enabled by a #define which is disabled by default.


iModelCore/ECDb/Tests/NonPublished/SchemaRemapTests.cpp
```diff
@@ -6724,4 +6724,521 @@ TEST_F(SchemaRemapTestFixture, InjectBaseClass4_Simplified)
     ASSERT_EQ(JsonValue(R"json([{"ifcCompositionType":"A"}])json"), result);
     }
     }
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+TEST_F(SchemaRemapTestFixture, RevitStoryScenario)
```
ColinKerr, line 6731: These tests are too large to exist without a description of the test case they are covering.  Either text or a diagram.  

Also it would be ideal if the differences between the new and old schemas was highlighted in the xml or the changes were inserted in the xml with some print f string.  As it is you have to copy them out and diff to see the difference.  Another option would be to save them as test data external to the test.  None of these options are great but we have to do better for our future selves that will have to read and understand them.

rschili: I've used printf and added an explanation to the schemas.

khanaffan (approved):

PR #160, opened 2023-02-13
Split geocoord tests into optional extensive tests used when require

AlainRobertAtBentley:
Limited list of representative GCS and Datum to a minimum in normal tests and moved tests that test GCS definition properties to extensive tests that can be called manually when the dictionary is updated.


kabentley (approved):

pmconne:
Is the motivation to reduce CI build times? If so, what percentage of time are you saving?

kabentley:
> Is the motivation to reduce CI build times? If so, what percentage of time are you saving?

Yes, I asked him to make these changes since the GeoCoord tests were taking the longest time of all of our tests (@AlainRobertAtBentley can give the before/after times). They were testing the data against the conversion library that only ever changes once in a blue moon, and only explicitly. Those tests only need to be run by whomever updates the data, on the rare occasion it happens, not every build.

AlainRobertAtBentley:
The unit tests are now down to around 2 minutes. The other extensive tests take 3 and half minutes so this is about what we will save on every build. Maybe I can shorten unit tests even more but it will take some more effort that would be the subject of future changes

kabentley:
So 5:30 vs 2:00. But mainly it is silly to test something that can't possibly change since the inputs to the tests (both the code and the data) are not versioned in this repo.

AlainRobertAtBentley:
Tests related to GCS definitions are limited to an extreme minimum. Only 118 GCS are used and they are selected so that one for each projection method/unit/selected datum are tested. The bulk of what remains is related to parsing WKT fragments, OSGeo/Autodesk XML fragments and matching EPSG codes. The WKT parse tests are still too long and that is what I will try to minimize later on.

PR #167, opened 2023-02-16
reverted change and added comment to explain the reason

RohitPtnkr1996:


pmconne:
Do your changes to ignore_list.txt in #146 need to be reverted?

khanaffan (approved):

RohitPtnkr1996:
@pmconne, I chose not to revert it.
There was only one test in the list, un-related to any of this.

PR #169, opened 2023-02-21
Added new schema merge flag to overwrite display labels

RohitPtnkr1996:


iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -237,8 +239,11 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
                     result.Issues().ReportV(IssueSeverity::Fatal, IssueCategory::BusinessProperties, IssueType::ECSchema, "Schema '%s' from %s side failed to be copied.", schema->GetFullSchemaName().c_str(), side);
                     failedToFillSchemas = true;
                     }
-                copiedSchema->SetOriginalECXmlVersion(schema->GetOriginalECXmlVersionMajor(), schema->GetOriginalECXmlVersionMinor());
-                result.GetSchemaCache().AddSchema(*copiedSchema);
+                if (copiedSchema.IsValid())
```
rschili, line 242: Nice catch, but better reverse this check, log an error and return ERROR in case the copied schema is not valid.

rschili: Also, can we hit this with a test?

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -58,8 +58,10 @@ BentleyStatus SchemaMerger::MergePrimitive(PrimitiveChange<T>& change, TParent*
     if(opCode == ECChange::OpCode::Deleted)
         return BentleyStatus::SUCCESS;
 
-    //if we prefer left values, and there is a valid old value, we keep it.
-    if(preferLeftValue && opCode == ECChange::OpCode::Modified && change.GetOld().IsValid())
+    // If the change type we're looking at is DisplayLabel and the "OverwriteDisplayLabel" flag is set in the merge options, we want to set the display label with the rightmost label value.
+    // Else, if we prefer left values or "OverwriteDisplayLabel" flag is unset, and there is a valid old value, we keep it.
+    if(!(change.GetType() == ECChange::Type::DisplayLabel && options.GetOverwriteDisplayLabel())
```
rschili, line 63: While logically correct and working, this condition will be very hard to read for our future selves.
I'd rather keep this spot here as is and change the callers which call MergePrimitive() for display labels provide the "PreferLeftValue" boolean parameter.
To get rid of the negation I'd rename the flag and make it default to true, like this:

`MergePrimitive(schemaChange->DisplayLabel(), left, &ECSchema::SetDisplayLabel, left->GetName().c_str(), result, options)`

Turns into
`MergePrimitive(schemaChange->DisplayLabel(), left, &ECSchema::SetDisplayLabel, left->GetName().c_str(), result, options, options.GetKeepDisplayLabels())`

That's overall more lines of code changed, but it keeps logic regarding display labels in a clear and easy to find location.

iModelCore/ecobjects/src/SchemaMerger.cpp
RohitPtnkr1996: The thing is, we're already doing that after we return from the lambda function.
We're checking if "failedToFillSchemas" got set to true and if it has, we're returning Error.
Also the error is being logged on line 239.
So, I just added the check.

And yes it can be hit from a test if the xml hasn't been read correctly. For eg, invalid schema name etc

iModelCore/ecobjects/src/SchemaMerger.cpp
RohitPtnkr1996: Yeah I quite like that.
But wouldn't the GetKeepDisplayLabels() make it more confusing ? Keeping display labels as they are is what we're doing by default anyway. The overwriting is something new we've added that we need to check.

Personally, I think having the negation won't be that bad, especially once we take a look at the function signature.
MergePrimitive(schemaChange->DisplayLabel(), left, &ECSchema::SetDisplayLabel, left->GetName().c_str(), result, options, !options.GetOverwriteDisplayLabel()) ?

Is this okay ? Or would you rather I change the getter like you've mentioned ?

rschili: Both are justifiable... Yeah I think your suggestion is fine, having the flag default to false seems more intuitive, just need to make sure we get the negation right on every call.

Naming in the merger has always been difficult. It's using the "left" "right" terms now consistently, to indicate that both sides are considered equal, none is the "existing" while the other is the "incoming". So calling it "Overwrite" isn't super clear on what gets overwritten with what. I'd call it "PreferRightSideDisplayLabels" or something similar.

rschili (approved):

ColinKerr (approved):

PR #173, opened 2023-02-22
Remove static schema holder classes

AkshayN5:


iModelCore/ecobjects/PublicApi/ECObjects/StandardCustomAttributeHelper.h
```diff
@@ -60,6 +62,9 @@ struct CoreCustomAttributeHelper final
     CoreCustomAttributeHelper();
     ~CoreCustomAttributeHelper();
 
+    static ECSchemaPtr m_schema;
```
pmconne, line 65: Prefix static members `s_`; `m_` is for non-static members.

iModelCore/ecobjects/src/StandardCustomAttributeHelper.cpp
```diff
@@ -129,155 +36,68 @@ ECObjectsStatus StandardCustomAttributeHelper::GetDateTimeInfo(DateTime::Info& d
...
+ECSchemaPtr CoreCustomAttributeHelper::m_schema = nullptr;
+//*********************** CoreCustomAttributeHelper *************************************
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+//static
+ECSchemaPtr CoreCustomAttributeHelper::_GetSchema()
     {
-    return GetHolder()->_CreateCustomAttributeInstance(attribute);
+    if (m_schema == nullptr)
```
pmconne, line 92: Race condition on a static variable.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchemaConverter.h
```diff
@@ -175,10 +175,9 @@ struct ECSchemaConverter
     //! @param[in] schema   The schema to traverse
     //! @param[in] context  A context to use if a particular converter needs to locate a schema which is not currently a schema reference.
     //! @param[in] doValidate Flag saying whether to validate the schema or not.  This is used by the DgnV8Converter to disable validation until it has had a chance to fix the schemas
-    static bool Convert(ECSchemaR schema, ECSchemaReadContextP context = nullptr, bool doValidate = true)
+    static bool Convert(ECSchemaR schema, ECSchemaReadContextR context, bool doValidate = true)
         {
-        if (nullptr != context)
-            GetSingleton()->AddSchemaReadContext(*context);
+        GetSingleton()->AddSchemaReadContext(context);
```
rschili, line 180: Mmh, this looks like it could be problematic. We have a global read context in a singleton which comes from "somewhere" and gets used "somewhere". I don't think read context was intended to be used like that. But I'm not sure if it can cause a problem when used like this.

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -834,7 +834,7 @@ ECObjectsStatus ECSchema::CreateEntityClass (ECEntityClassP& pClass, Utf8StringC
 //---------------------------------------------------------------------------------------
 // @bsimethod
 //---------------+---------------+---------------+---------------+---------------+-------
-ECObjectsStatus ECSchema::CreateMixinClass (ECEntityClassP& pClass, Utf8StringCR name, ECEntityClassCR appliesTo)
+ECObjectsStatus ECSchema::CreateMixinClass (ECEntityClassP& pClass, Utf8StringCR name, ECEntityClassCR appliesTo, ECSchemaReadContextR schemaContext)
```
rschili, line 837: This is one spot that could potentially come back and bite us. I believe the method is commonly used in many spots, I don't feel good about changing its signature.
As an alternative, the method could require the schema to have a reference to CoreCustomAttributes and fail if the reference is missing. We could then add this overload alongside the method. So we don't break anybody. However, this could lead to runtime errors which is even worse :-) So I guess I'm fine but not happy with this...

iModelCore/ecobjects/test/NonPublished/CustomAttributeTests.cpp
```diff
@@ -613,7 +613,7 @@ TEST_F (CustomAttributeTest, SerializeSchemaToXmlUtfString)
     caClass->CreatePrimitiveProperty (stringProp, caUtf8StringPropName, PRIMITIVETYPE_String);
 
         // *** WIP_PORTABILITY: Use an escape such as \u here. Don't try to use extended ascii directly
-    WCharCP caPropValueString = L"����ᳵ";
+    WCharCP caPropValueString = L"����ᳵ";
```
rschili, line 616: was this on purpose? Looks to me like the unicode in the file broke?

rschili (commented):
@ColinKerr Akshay pretty much implemented what we discussed and I had in mind. Seeing it, I wonder about an alternative.
Making ReadContext the "lifeline" of schemas in memory like we do now makes memory management transparent, so I like it. But it's also a hassle because up until now nobody had to care when it came to these helpers.
We now have a lot of potential  bugs and crashes because schemas could be deleted and later accessed, our test coverage cannot ensure callers do the right thing.
As an alternative, so people can continue to "not care", why don't we turn the list of referenced schemas in ECSchema from a vector of ECSchemaP into a vector of ECSchemaPtr? We never want referenced schemas to die before a schema that references them does. Do you know the reason why we didn't implement it like this?

iModelCore/ecobjects/test/NonPublished/CustomAttributeTests.cpp
AkshayN5: It appears to be broken, and I didn't tamper with this section of the code.

rschili: Can you try and revert it? In this comment view, this looks like nothing changed, but actually, the characters changed from german special chars into some invalid stuff.
chars

rschili (approved):

pmconne (changes requested):
Do not resolve other people's comments on your PRs.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchemaConverter.h
khanaffan: I do not even see the point of this PR if we going to keep a static schema cache. I thought the point was to remove the global static cache. This PR only changes how its initialized rather than removing it.

ColinKerr: The singleton pattern was used here to keep the original static version of the API available when we added a configurable instance version.  We can probably get rid of the static API and move entirely to an instance version so long as we update the connectors and keep their behavior consistent

khanaffan: @AkshayN5 explain that to me. But my primary concern was` ~ECSchema()`dependency on global BeMutex. Having `SchemaReadContext` cached has that dependency that can cause a crash when the application exit when BeMutex is destroyed before the static schema cache. C++ does not ensure in which order in which global objects are destroyed and we saw the issue on MacOS/Linux where the test suit crashed on exit.

@AkshayN5 seems to have resolved my concern in his recent push by not caching read the context in a static member.

AkshayN5: @khanaffan Yes, I removed the read context from the static member in the recent commit. Schema context is now passed to the required functions as a parameter.

ColinKerr (approved):

pmconne:
@khanaffan are you satisfied?

khanaffan:
> 

yes its good.

khanaffan (approved):

pmconne (approved):

PR #176, opened 2023-02-27
Fixed issue where base class compare fails for class name with underscore

RohitPtnkr1996:
This PR is for the ticket https://dev.azure.com/bentleycs/iModelTechnologies/_sprints/taskboard/EC/iModelTechnologies/Iteration%2097?workitem=1091617


iModelCore/ecobjects/src/SchemaComparer.cpp
```diff
@@ -916,7 +916,7 @@ BentleyStatus SchemaComparer::CompareBaseClasses(BaseClassChanges& changes, ECBa
         }
     
     // now compare mixins by full name
-    bset<Utf8String, CompareIUtf8Ascii> oldMixinFullNames, newMixinFullNames;
```
rschili, line 919: So, what happened when there is an underscore in the set?
I'm worried because we use this Comparer all over the place (quick search shows 105 callers), it relies on BeStringUtilities::StricmpAscii().
@ColinKerr for discussion.

RohitPtnkr1996: This issue we're facing seems to be a one off. It'll only happen when the two schemas being compared have base classes that match exactly upto the first underscore.
The function std::set_difference needs the argument containers to be pre-sorted.

For example, let's say we have a common "BaseClass" on both left and right, and a new class "Base_Class" is being added on the right.
In order for set_diff to work, the two containers need to be <BaseClass> and <BaseClass, Base_Class> since the uppercase 'C' has a lower ASCII value than '_'.
However, the CompareIUtf8Ascii converts all chars to lowercase and then compares them one-by-one.
So now, when "base_class" is being compared with "baseclass", the char '_' gets compared with 'c' and since '_' has a lower ASCII value than the lowercase C, the container gets re-sorted to <base_class, baseclass> which causes set_diff to output incorrect values.
This then causes the wrong class to get marked for deletion and we get the error reported on the ticket.

rschili: Good explanation. I'm now convinced, we have more problems like this in our codebase. I'm approving this PR, but we should put a story on the backlog to fix the general scenario.

rschili (approved):

RohitPtnkr1996:
I've added a new test in BeStringUtilities that is testing the core reason for this bug.
It's currently failing so I've added it to the ignore list with the appropriate explanation.

iModelCore/ecobjects/src/SchemaComparer.cpp
ColinKerr: This confuses me.  If the problem is sort order then wouldn't the following fail with either comparison algorithm:

Old:
TestClass
  B

New:
TestClass
  B
  A

because A will be sorted before B.  If not how is it different than base classes BaseClass and base_class?

RohitPtnkr1996: No that's not quite right. Your example will work correctly since the correct sort order **should** be "A, B".
However, in the New side, if you changed the 3rd class to "_B", you'll get the error.

Like I said, this is more of a corner case which will only happen for special characters that lie between uppercase and lowercase alphabets in the ASCII table are compared with uppercase characters.
The default sorter in bset is already doing everything correctly.
But, ComapreIUtf8ASCii converts everything to lowercase, so the alphabets being compared always end up greater than the special chars.
The test case I've added to BeStringUtilities_test shows this clearly.

RohitPtnkr1996: Also, this whole problem with correct sorting being so necessary only happens because we're using std::set_difference to figure out which classes have been added/removed.
We wouldn't care about the sort order if we were doing it any other way, easiest one being a simple for loop with find() function.

RohitPtnkr1996: Had a discussion with Colin.
The fix has been modified where instead of removing ComapreIUtf8ASCii from the bset, I'm instead passing the ComapreIUtf8ASCii comparison logic to the comparer in std::set_difference.

ColinKerr (approved):

PR #179, opened 2023-02-27
Add cloudcache comments

DanRod1999:
Issue: https://github.com/iTwin/itwinjs-backlog/issues/579

added comments to all Cloudcontainer functions to clarify a cloudcache needs to be connected in order to use them.


iModelCore/BeSQLite/CloudSqlite.cpp
```diff
@@ -244,6 +244,7 @@ int CloudCache::FindToken(Utf8CP storage, Utf8CP accessName, Utf8CP containerNam
 
 /**
  * Attempt to connect this container to a CloudCache and its cloud container
+ * @note this function requires the cloudcache to be connected to the cloud container
```
nick4598, line 247: Sorry I mentioned everything that includes a cloudcache, but this one actually instantiates the cloudcache

DanRod1999:
Don't know if there is a better way to add these comments, but I noticed the `@note` tag was used in other places. If that isn't the proper use case I can change it.

iModelCore/BeSQLite/CloudSqlite.cpp
```diff
@@ -271,6 +272,7 @@ CloudResult CloudContainer::Connect(CloudCache& cache) {
 
 /**
  * Disconnect this container from the CloudCache. If the containerDb is open, close it first.
+ * @note this function requires the cloudcache to be connected to the cloud container
```
nick4598, line 275: Could we add to the message, "See CloudContainer::Connect"

nick4598:
Just adding these comments in because iModelHub team ran into some trouble using some of the functions on CloudContainer not realizing they had to connect a cloudcache. People probably won't run into this again, but just incase.

nick4598 (approved):

iModelCore/BeSQLite/CloudSqlite.cpp
```diff
@@ -271,6 +271,7 @@ CloudResult CloudContainer::Connect(CloudCache& cache) {
 
 /**
  * Disconnect this container from the CloudCache. If the containerDb is open, close it first.
+ * @note This funciton does nothing if there is no cloudcache connected to the cloud container. See CloudContainer::Connect
```
kabentley, line 274: this wording is backwards. You connect a CloudContainer to its CloudCache, not the other way around (multiple containers may connect to a single cache.)

kabentley (changes requested):

kabentley:
Just to be clear, this whole api is internal. Using @note is fine, but only people looking at the code will see any of this. All the *real* documentation should be on the TypeScript layer. Please read that and see if it is clear about connect/disconnect. 

If the iModelHub team is using this outside of the TypeScript api, that can only be temporary from a fork.

iModelCore/BeSQLite/CloudSqlite.cpp
kabentley: also, "funciton"

PR #180, opened 2023-02-28
unicode aware trim for Code Values

MichaelBelousov:
Use unicode aware functions from icu4c to implement a unicode-aware trim function. Use it when create `DgnCodeValue` objects.

- There are some iModels in production that contain at least untrimmed non breaking spaces in their code values.
- unicode's own trim routine does not trim `00a0`/`&nbsp;`, the non-breaking space, which is a convention created by Java, that `&nbps` semantically should not participate in trimming.
- I investigated using node's internal icu4c dependency, and have [asked them](https://github.com/nodejs/node/discussions/46885) (in multiple places) if there's a way we can at least drop bundling our own icu4c data and use theirs.
- here's [a microbenchmark](https://github.com/MichaelBelousov/ascii-vs-icu-trim-benchmark), TLDR utf8 aware impl is about 2-3x slower, but we're only using it on code values right now.
- this needs to be ported for consumption by the connectors


pmconne (commented):
Can you add a test to Utf8String_test.cpp?

kabentley:
Presumably this won't be much slower for the vast majority of cases that don't have spaces at the start or end? Did you test that?

iModelCore/Bentley/Bentley/WString.cpp
```diff
@@ -374,6 +374,55 @@ Utf8StringR Utf8String::Trim ()
...
+// https://unicode-org.github.io/icu/userguide/icu4c/build.html#recommended-build-options
+#define U_NO_DEFAULT_INCLUDE_UTF_HEADERS 1
+#include <unicode/utypes.h>
+#include <unicode/utf8.h>
+#include <unicode/uchar.h>
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+Utf8StringR Utf8String::TrimUtf8 ()
```
kabentley, line 387: move open brace to end of previous line: `() {

MichaelBelousov:
> Presumably this won't be much slower for the vast majority of cases that don't have spaces at the start or end? Did you test that?

I tested and it seems to be roughly 1.4-2x slower for those cases. I tested in the linked benchmark by settings `#define TESTSPC ""`. My benchmark isn't great, I should probably read a much larger set of strings from some corpus at init time.

iModelCore/Bentley/Bentley/WString.cpp
MichaelBelousov: done

MichaelBelousov:
> Can you add a test to Utf8String_test.cpp?

done. Sorry for failing to find that file in the first place. Please make any test case suggestions.

pmconne (approved):

MichaelBelousov:
opened port: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel02/pullrequest/317842

MichaelBelousov:
I just realized that this actually makes it impossible for the transformer to preserve code values in the iModels that already have `&nbsp;`. I think I will have to add an API for use by the transformer to preserve them during element insert... to prevent people forking an iModel, getting some code values from master using raw ECSQL (without wrapping them in a `Code` instance which would fix that) and trying to find them in a branch which had its code values "fixed" by this change.

PR #182, opened 2023-03-01
Check return values of prepare and step in IsWalMode()

nick4598:
This function segfaults if prepare / step on pragma journal_mode don't succeed when we call stmt.getValueText. 

Not sure how the file I'm using was created, but I've asked about that. This file has a valid SQLITE db header, so it succeeds on isValidDbFile(), but then fails with SQLITE_NOTADB in the prepare for pragma journal_mode. Previously, isWalMode did not exist, so attempting to open this db would fail (as long as rawSQLite is set to false) on 
>rc = statement.TryPrepare(*this, "SELECT Namespace,Name,Id,SubId,TxnMode,RawSize,Data,StrData FROM " BEDB_TABLE_Property);

which takes place a little while later in the function.

Opening this database with rawSQLite set to true allows us to open the database successfully with my change, and also before IsWalMode() was introduced I'd imagine. I assume any queries on it would fail though.

This seems sort of wrong to me... should we maybe do some more work to figure that out in the open? Maybe TryPrepare(pragma journal mode) ? I tried that out and the cloudsqlite tests were failing, so maybe we should do it only if params.m_skipFileCheck is set to false. @kabentley @khanaffan any thoughts there?


kabentley:
I made a slight change to your change. I think this is fine (attempting to open an invalid file with raw mode will fail later, that's OK).

It'd be nice to add a test for this. Do you have a little corrupt file we can add to the tests?

nick4598:
> I made a slight change to your change. I think this is fine (attempting to open an invalid file with raw mode will fail later, that's OK).
> 
> It'd be nice to add a test for this. Do you have a little corrupt file we can add to the tests?

Actually I don't think opening the file fails. My simple test in TS succeeds as long as we don't check for the presence of the be_prop. 
The test right now is just: "db2.openDb(path.join(__dirname, "InvalidFile.bim"), {rawSQLite: true, openMode: OpenMode.Readonly});"

kabentley:
> Unless you mean we wont fail when opening but will just fail later and that part you are okay with

Yes, that's what I mean. That's the normal SQLite behavior - you don't fail until you attempt to read something from the file.

kabentley (approved):

khanaffan:
It's less known but you can use pragma as a table value function.
`select null from pragma_journal_mode() WHERE journal_mode ='wal'`
and then just `return stmt.Step() == BE_SQLITE_ROW`

khanaffan (approved):

nick4598:
Added test. This test would segfault before the change.

nick4598:
@mergifyio backport release/3.x

PR #188, opened 2023-03-06
PrgOutputDir is defined

naveedkhan8067:
The attribute is required for `bb taglist` command. 

**Note:** The build has started failing with error: `PrgOutputDir does not exist for part RunIModelEvolutionTests. It is needed for the name of the tag file`

### DIFF
--- a/iModelCore/iModelPlatform/iModelPlatform.PartFile.xml
+++ b/iModelCore/iModelPlatform/iModelPlatform.PartFile.xml
@@ -143,7 +143,7 @@
          ********************************************************************************************** -->
 
     <!-- Entry point part -->
-    <Part Name="RunIModelEvolutionTests">
+    <Part Name="RunIModelEvolutionTests" PrgOutputDir="IModelEvolutionTests">
         <SubPart PartName="IModelEvolutionRunAllTestRunners"/>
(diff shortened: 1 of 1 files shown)


iModelCore/iModelPlatform/iModelPlatform.PartFile.xml
```diff
@@ -143,7 +143,7 @@
          ********************************************************************************************** -->
 
     <!-- Entry point part -->
-    <Part Name="RunIModelEvolutionTests">
+    <Part Name="RunIModelEvolutionTests" PrgOutputDir="IModelEvolutionTests">
         <SubPart PartName="IModelEvolutionRunAllTestRunners"/>
```
chuckkir, line 147: I wouldn't have thought this necessary unless posting BDFs?

naveedkhan8067: Not sure, in past `bb taglist` command was working fine in the build. Now its failing with the error described above. That's why I added it.

chuckkir: Ok. I'll have to ponder that, but it's not a bad idea to have the taglist in the artifacts.

chuckkir (approved):

pmconne (approved):

PR #191, opened 2023-03-08
Move `BeJsValue.h` away from `jsoncpp`

grigasp:
Related to https://github.com/iTwin/itwinjs-backlog/issues/444#issuecomment-1454712836.

This PR inverts the dependency between `BeRapidJson` and `BeJsonCpp` parts - previously `BeJsonCpp` had a `BeRapidJson` dependency and now it's the opposite. 

I initially tried to avoid making `BeRapidJson` depend on `BeJsonCpp`, but that required removing jsoncpp-related implicit constructors from `BeJsConst` and `BeJsValue`. That broke tons of code across the whole repository in places where `jsoncpp` is still used... I believe we need to first get rid of `jsoncpp` usages before removing this dependency.

### DIFF
--- a/iModelCore/BeSQLite/BeSQLite.PartFile.xml
+++ b/iModelCore/BeSQLite/BeSQLite.PartFile.xml
@@ -4,7 +4,7 @@
 
     <!-- Bind my own PublicAPI and VendorAPI into BuildContext. Note that this part does not cause any sub-part PublicAPIs to be bound. Also deliver my notices. -->
     <Part Name="PublicAPI" BentleyBuildMakeFile="BeSQLite.prewire.mke">
-        <SubPart PartName="PublicAPI" PartFile="iModelCore/libsrc/jsoncpp/BeJsonCpp" />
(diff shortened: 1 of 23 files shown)


pmconne:
Did you plan to update imodel-native-internal? I would have expected changes to be needed in Visualization, but the CI jobs appear to be happy...

grigasp:
> Did you plan to update imodel-native-internal? I would have expected changes to be needed in Visualization, but the CI jobs appear to be happy...

I didn't find any `BeJsValue.h` includes in `imodel-native-internal`. The partfiles don't specify `BeRapidJson` as a dependency, but it comes indirectly from other dependencies.

pmconne:
@grigasp thanks, I see Visualization and PSBRepGeometry are still using BeJsonCpp for all of their JSON needs. We'll have to follow up with a separate PR.

pmconne (approved):

kabentley (approved):

dassaf4 (approved):

ColinKerr (approved):

saskliutas (approved):

PR #192, opened 2023-03-08
Add `model` argument to Element and ElementAspect callbacks

kabentley:
This is required for channel enforcement.


pmconne:
@DanRod1999 why is testing against itwinjs-core optional? The whole point is that the native PR should fail unless itwinjs-core builds successfully.

> If you have changes only in native, but wish to test against itwinjs-core master or imodel02
>
>    Create your PR in imodel-native and place the link for the itwinjs-branch as follows:
>        imodel-native: https://github.com/iTwin/imodel-native/tree/master
>        Replace master with imodel02 if that is the branch you wish to use
>        Note: testing against imodel02 tends to give problems with Rush Audit

DanRod1999:
> @DanRod1999 why is testing against itwinjs-core optional? The whole point is that the native PR should fail unless itwinjs-core builds successfully.
> 
> > If you have changes only in native, but wish to test against itwinjs-core master or imodel02
> > Create your PR in imodel-native and place the link for the itwinjs-branch as follows:
> > imodel-native: https://github.com/iTwin/imodel-native/tree/master
> > Replace master with imodel02 if that is the branch you wish to use
> > Note: testing against imodel02 tends to give problems with Rush Audit

This can easily be set (and was the initial intention). However, I had a lot of problems with rush audit failing because it wasn't up to date. I left it out because I didn't want to hold up peoples PRs for no reason. This might require a change to the itwinjs-core build steps. Maybe make audit optional so it can skip on PR val builds. This is something I could add. @calebmshafer or @aruniverse does this sound like an appropriate solution?

aruniverse:
> Maybe make audit optional so it can skip on PR val builds

@DanRod1999 
Yea, I would make the `audit` and `lint` tasks optional for pr-val. I'm debating whether or not the `rush change -v` task should be optional, but there might just be a bug with this step. I was under the impression that if the currentBranch === targetBranch, we would skip this check, [but this says otherwise](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2004155&view=logs&jobId=bbf0d238-6f46-513d-50ac-f6207b4e71d6&j=bbf0d238-6f46-513d-50ac-f6207b4e71d6&t=294867a9-5cd0-5c54-e408-b626d5a99664)

pmconne:
> lint tasks optional

Why? Will lint fail for some reason other than the PR author overlooked something that whoever publishes the next addon will have to fix?

aruniverse:
> > lint tasks optional

> Why?

The code will need to be linted prior to merging into master so at some point it will be fixed, this would just defer when.
Failing pr val due to a linter error on the js side would honestly annoy me. I personally would care more that im actually able to build the js and tests pass. Sitting around 25-30 mins for the native pr to build successfully, then fail immediately for something that should be trivial to fix like a linter error and not actually test the native changes is just a waste of time imo

pmconne:
> Failing pr val due to a linter error on the js side would honestly annoy me
> just a waste of time

It sounds like you have not wasted days of your life fixing other people's oversights while trying to get an addon out the door.
Lint errors are not always trivial.
If the PR author wishes to avoid wasting their own time due to lint errors during CI, they can run lint locally before submitting their PR.

aruniverse:
> Lint errors are not always trivial.

> If the PR author wishes to avoid wasting their own time due to lint errors during CI, they can run lint locally before submitting their PR.

Fair points. I guess my concerns for "wasted" native build times can be alleviated with some further build caching, maybe something we will look into further later. 

I guess for the time being @DanRod1999 , lets keep linting required

pmconne (approved):

PR #211, opened 2023-03-17
Let clients of the daemon work with readonly perms to cloudcache. Update SQLite to 3.41.1 from 3.41.0

nick4598:
New daemon will work with old clients. New clients won't work with old daemon. Need to update daemon first but shouldn't be a problem.

Also includes update to 3.41.1, I could revert this but this is what is on trunk for CloudSQLite at the moment


nick4598:
@khanaffan  I noticed you added SQLITE_CHANGESETAPPLY_IGNORENOOP to the sqlite3.h file, is that the only bentley specific change? Maybe I should revert the SQLite update I'm doing here?

kabentley:
@nick4598 this is not a Bentley specific change. Dan somehow lost this change that he made for @khanaffan on the main branch.

khanaffan:
Do not update sqlite like this, I have setup a [process](https://dev.azure.com/bentleycs/iModelTechnologies/_wiki/wikis/iModelTechnologies.wiki/28770/Guide-to-updating-SQLite-on-imodel-native) for it so we can pick Bentley specific changes while being consuming release version of sqlite and precisely know which changes went into which version of itwinjs. 

Updating from master has not proven well in past due to unexpected performance issues and confusing version as master has next version. We had issue backporting a sqlite security patch because we did not knew which version of sqlite code went into 2.19.x. Updating to latest caused ECPresentation to experience performance issues.

This is why it's important not to take sqlite from master or use a amalgamated version from cloud sqlite repo.

Point me to changes Dan made for you and I will pull that in and update sqlite.

kabentley:
@khanaffan I don't think Dan made any changes to SQLite for this change (he can't on this branch). He just periodically updates the copy of SQLite he uses to build/test it. I think @nick4598 should just remove the changes to sqlite.h and sqlite.c from this PR.

kabentley (approved):

PR #213, opened 2023-03-21
 Treat numeric INF/NAN as NULL in IModels

khanaffan:
* ECSQL statement will treat inf/nan as NULL when binding to prevent them from entering an IModel
* For existing cases of INF as a property value we patch RapidJson to render INF/NAN as NULL when generating JSON string.


iModelCore/libsrc/rapidjson/vendor/include/rapidjson/writer.h
```diff
@@ -547,7 +547,9 @@ inline bool Writer<StringBuffer>::WriteDouble(double d) {
     if (internal::Double(d).IsNanOrInf()) {
         // Note: This code path can only be reached if (RAPIDJSON_WRITE_DEFAULT_FLAGS & kWriteNanAndInfFlag).
         if (!(kWriteDefaultFlags & kWriteNanAndInfFlag))
```
kabentley, line 549: this code is in this file in 2 places. You should make this change in both of them.

kabentley (approved):

kabentley (approved):

PR #216, opened 2023-03-27
Use server time rather than local computer time for write lock expiration

kabentley:
Write locks for CloudContainers expire after a duration (usually an hour). Previously the times were dependent on local computer times, leading to the dependence that they all be relatively in sync. 

This PR includes a change from SQLite.org to use the time from the server, so expiration duration is not ambiguous and does not use local computer times at all.


kabentley:
@wgoehrig this change eliminates the potential problem with time skew between computers we discussed. I hope we don't waste time on token expiration.

nick4598 (approved):

PR #223, opened 2023-03-29
Profile upgrade fails due to a data transformation is detected where there is none

khanaffan:
The issue happens if iModel has an overflow table. Schema import always runs instance upgrade on all classes that are mapped overflow table.


pmconne (approved):

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -4,6 +4,7 @@
 *--------------------------------------------------------------------------------------------*/
 import { DbResult, Id64Array, Id64String, IModelStatus, OpenMode } from "@itwin/core-bentley";
 import { BlobRange, DbBlobRequest, DbBlobResponse, DbQueryRequest, DbQueryResponse, DbRequestKind, DbResponseStatus, ProfileOptions } from "@itwin/core-common";
+import { DomainOptions } from "@itwin/core-common/lib/cjs/BriefcaseTypes";
```
aruniverse, line 7: is this meant to be imported from cjs like this? outside the barrel file?

khanaffan: I did not realize it was imported like this. I think VSCode automatically did that when I used a type. I will fix it.

ColinKerr (approved):

PR #232, opened 2023-04-07
Make application fail when SQLite hits a fatal error

khanaffan:
This change is to avoid corrupt changeset being pushed to Hub. Currently, we have undefined behavior where the application after hitting critical error continues to run and push an invalid changeset to the hub. 

* throw `std::runtime exception` when SQLite hits a fatal error.
* `DgnDbNative.triggerAutoCommitFailure()` allow triggering auto commit for testing and verifying we throw a c++ exception.

### DIFF
--- a/iModelCore/BeSQLite/BeSQLite.cpp
+++ b/iModelCore/BeSQLite/BeSQLite.cpp
@@ -770,7 +770,6 @@ BeGuid   Db::GetDbGuid() const {return m_dbFile->m_dbGuid;}
 int32_t  Db::GetCurrentSavepointDepth() const {return (int32_t) m_dbFile->m_txns.size();}
 Utf8String Db::GetLastError(DbResult* lastResult) const {return IsDbOpen() ? m_dbFile->GetLastError(lastResult) : "Not opened";}
 void Db::Interrupt() const {return sqlite3_interrupt(GetSqlDb());}
-
 int64_t  Db::GetLastInsertRowId() const {return sqlite3_last_insert_rowid(GetSqlDb());}
 int      Db::GetModifiedRowCount() const {return sqlite3_changes(GetSqlDb());}
 int      Db::GetTotalModifiedRowCount() const { return sqlite3_total_changes(GetSqlDb()); }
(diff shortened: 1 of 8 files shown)


iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -933,29 +932,21 @@ static Utf8CP getStartTxnSql(BeSQLiteTxnMode mode)
...
     // Sqlite initiate auto rollback in case of SQLITE_FULL, SQLITE_IOERR, SQLITE_NOMEM, SQLITE_BUSY, and SQLITE_INTERRUPT
     // We force a crash if COMMIT/ROLLBACK is called outside BeSQLite api.
     if (sqlite3_get_autocommit(m_sqlDb) != 0 ) {
         LOG.error("Sqlite initiated autocommit due to a fatal error.");
-        if (BeSQLiteLib::s_throwExceptionOnUnexpectedAutoCommit) {
-             LOG.error("Runtime debug option to throw exception on unexpected autocommit is set to *true*. Caller must handle exception and then delete the briefcase/db afterword.");
-            throw std::runtime_error("sqlite initiated autocommit due to a fatal error");
-        }
-        return 0;
+        throw std::runtime_error("sqlite initiated autocommit due to a fatal error");
```
pmconne, line 948: Do we really want to crash the backend unconditionally?
What is frontend supposed to do in response?
Can this be handled more gracefully?

khanaffan: Yes. There is no other way around it. This can happen during any call to sqlite. The safest thing is to fail. The other way around it is to add try{} catch{} to all DgnDbNative calls and catch this exception or check autocommit status and call some callback into JS.

khanaffan: It is rare for this to happen but when it happens, we end up pushing an invalid changeset where a change is locally rollbacked but its presence in the changeset. The flag to throw it as an exception was for connectors which also avoids pushing the invalid changeset but failing rather than continuing.

kabentley: you should at least include some indication of the SQLite error. Is that available?

pmconne: > What is frontend supposed to do in response? How will it even know?

> What is the user supposed to think when the app just crashes? How will they know, for example, that they need to free up disk space?

khanaffan: @pmconne that was when `saveChanges()` fails. It can fail for a bunch of reasons as well. But once it fails there is nothing you can do either. As my test shows `BeSQliteTestFixture.HandleSqliteFullFailure` if you hit SQLITE_FULL even abandonChanges() will fail. But it does not cause SQLite to call auto-commit. 

The auto-commit case is different. It is invoked by SQLite to last ditch attempt to recover by rollbacking everything. But given it can happen at any point in time during the execution of a SQL statement. It can even happen in COMMIT or ROLLBACK.

khanaffan: @kabentley yes I tried to get the last error and it seems it's not set.

khanaffan:
We do not need this as is. I am implementing an alternative solution.

PR #239, opened 2023-04-12
Expose virtual table api via BeSQLite

khanaffan:
This is to allow certain experiment on ECDb level.


kabentley:
i think you should create new cpp/h files for this stuff

khanaffan:
> i think you should create new cpp/h files for this stuff

done.

kabentley (approved):

PR #248, opened 2023-04-19
Fix pugixml whitespace parsing in instances

rschili:
properties that contain only whitespaces were not parsed correctly as pugi considers the whitespace insignificant. An option can be used to change the default behavior.


iModelCore/ecobjects/src/Format.cpp
```diff
@@ -312,7 +312,7 @@ SchemaReadStatus ECFormat::ReadXml(pugi::xml_node unitFormatNode, ECSchemaReadCo
 
     SetNumericSpec(spec);
     pugi::xml_node child = unitFormatNode.first_child();
-    if (child)
+    if (child && child.type() == pugi::xml_node_type::node_element)
```
ColinKerr, line 315: Do we need some more tests to ensure that we have updated the deserialization code to take these new empty string nodes into account?

rschili: There is an existing test that covers the problem, but it's not the intended purpose of that test. I'll add some dedicated testing for this

ColinKerr (approved):

PR #252, opened 2023-04-24
Lazy init prop_exists func due to hit to startup performance

khanaffan:


rschili (approved):

kabentley:
Please explain what you observed to prompt this change and when it started. When will the "lazy" function now be hit?

iModelCore/ECDb/ECDb/ECDbSqlFunctions.h
```diff
@@ -219,7 +219,7 @@ struct PropExistsFunc final : ScalarFunction , ECDb::IECDbCacheClearListener {
     public:
         explicit PropExistsFunc(ECDbCR ecdb) : ScalarFunction(SQLFUNC_PropExists, 2, DbValueType::IntegerVal), m_ecdb(ecdb) {
             const_cast<ECDbR>(m_ecdb).AddECDbCacheClearListener(*this);
-            m_propMap.Build(m_ecdb, true);
+            // m_propMap.Build(m_ecdb, true);
```
kabentley, line 222: remove this commented out code and replace it with an explanation that it can't be done here and why.

kabentley (changes requested):

khanaffan:
> Please explain what you observed to prompt this change and when it started. When will the "lazy" function now be hit?

This was reported by @nick4598 .  According to him it's added startup time to the backend and if it can be avoided. The code in question runs a query that results in the download of multiple 4Mb blocks for v2 a checkpoint. Which can be avoided so my change is regarding that. The query itself is not slow; the download blocks add time.

iModelCore/ECDb/ECDb/ECDbSqlFunctions.cpp
```diff
@@ -21,7 +21,9 @@ void PropExistsFunc::_ComputeScalar(Context& ctx, int nArgs, DbValue* args) {
         ctx.SetResultError("prop_exists(I,S]) expect two args");
         return;
     }
-
+    if (m_propMap.Empty()) {
```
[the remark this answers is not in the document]
khanaffan: It's not expensive, the query takes like < 1s for on-disk files. It's a v2 downloading block that is slow and can happen to any query executed when blocks are not in the cache.

iModelCore/ECDb/ECDb/ECDbSqlFunctions.h
khanaffan: done

iModelCore/ECDb/ECDb/ECDbSqlFunctions.cpp
```diff
@@ -22,6 +22,7 @@ void PropExistsFunc::_ComputeScalar(Context& ctx, int nArgs, DbValue* args) {
         return;
     }
 
+    if (m_propMap.Empty()) m_propMap.Build(m_ecdb, true);
```
kabentley, line 25: put this on 2 lines and explain why it is lazy created rather than where it was before

kabentley (changes requested):

iModelCore/ECDb/ECDb/ECDbSqlFunctions.cpp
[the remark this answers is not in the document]
kabentley: don't explain it to me, explain it in the code so if anyone sees it in the future they understand why it is here

iModelCore/ECDb/ECDb/ECDbSqlFunctions.cpp
pmconne, line 25: Can two threads call `_ComputeScalar` on the same `PropExistsFunc` object simultaneously?

khanaffan: There is a case where you can copy a SqlFunc to two different connections and in that case `_ComputeScalar` will be called by two different threads.  But we do not do that anywhere in ECDb. ECPresentation does that by registering the same function across multiple connections.

kabentley (approved):

PR #258, opened 2023-04-27
exportGraphics: new option to export raw uv-parameters

dassaf4:
Honor the new `exportRawParameters` option in `JsInterop::ExportGraphics` and `JsInterop::ExportPartGraphics`.

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/da4/export-graphics-raw-uv

### DIFF
--- a/iModelJsNodeAddon/JsInteropExportGraphics.cpp
+++ b/iModelJsNodeAddon/JsInteropExportGraphics.cpp
@@ -215,10 +215,11 @@ public:
         };
     bvector<LineStyleGeometryCacheEntry>    m_lineStyleGeometryCache;
     int32_t                                 m_drawingStyledCurveVectorStack;
+    bool                                    m_exportRawParameters;
 
-    ExportGraphicsProcessor(DgnDbR db, IFacetOptionsR facetOptions, double decimationTolerance, bool generateLines, double minLineStyleComponentSize) :
+    ExportGraphicsProcessor(DgnDbR db, IFacetOptionsR facetOptions, double decimationTolerance, bool generateLines, double minLineStyleComponentSize, bool exportRawParameters = false) :
         m_db(db), m_facetOptions(facetOptions), m_gotBadPolyface(false), m_decimationTolerance(decimationTolerance),
-        m_generateLines(generateLines), m_minLineStyleComponentSize(minLineStyleComponentSize), m_drawingStyledCurveVectorStack(0) { }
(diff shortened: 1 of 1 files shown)


pmconne:
This seems wrong. The pattern map should correctly specify how to compute the UV params.
Does the texture look correct when viewing the iModel in a viewer like display-test-app? Are they only wrong when viewing the output of `exportGraphics`? If so, shouldn't we fix `exportGraphics`?
It's possible `exportGraphics` is attempting to work around [this](https://www.puredevsoftware.com/blog/2018/03/17/texture-coordinates-d3d-vs-opengl/) with glTF (following OpenGL conventions) being the target, based on this comment:
```
// TextureMapping::Trans2x3 inverts V so that the frontend doesn't need to flip textures
// on load. We need to invert back to match interchange conventions.
```

dassaf4:
> Does the texture look correct when viewing the iModel in a viewer like display-test-app?

Yes. DTA displays the user's testcase the same way as the texture is seen in Blender. The testcase is a bim file containing a single triangle with a patternMap texture with scale [1,-1] and Units.Meters.

> Does it only look wrong when viewing the output of `exportGraphics`?

Yes. The user states the output of `exportGraphics` on their testcase doesn't look right in their viewer (based on HOOPS engine). 
However, they state that on house_model.bim (sample from cra-template-desktop-viewer), `exportGraphics` produces meshes that _do_ look right in their viewer. I'd like to view the output of `exportGraphics` on the user's bim file in a gltf viewer via test-apps\export-gltf, but I need help with that.

> If so, shouldn't we fix `exportGraphics`? 

Yes. If there is a problem with `exportGraphics`, it might be specific to the type of texture in the user's testcase. I don't know enough about texture parameters to know what `exportGraphics` should be returning for this texture and geometry. Neither does the user, but they suggested they might have better luck with raw parameters returned from `exportGraphics`. This PR doesn't change existing behavior; rather it provides a new option to `exportGraphics` for the user to try.

pmconne:
> I'd like to view the output of exportGraphics on the user's bim file in a gltf viewer via test-apps\export-gltf, but I need help with that.

You can drag and drop glTF files into Paint3d, which comes installed on WIndows.

dassaf4:
> You can drag and drop glTF files into Paint3d, which comes installed on WIndows.

Cool! But my immediate problem is how to generate the glTF file using `exportGraphics`. Our test-app\export-gltf apparently does this, but how do I "run" it on the user's bim file?

pmconne:
Ask user to supply the iModel, ideally as a file, or at least as a link to it on the hub (and grant you access to it).
Then I guess ping me.

PR #270, opened 2023-05-08
Use `baseUri` rather than `accessName` for cloud containers

kabentley:
itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/azure-baseuri


kabentley:
@nick4598 this will require changes for the checkpoint creation agent.

pmconne (approved):

khanaffan (approved):

PR #271, opened 2023-05-08
Ignore subcategory for non-element geometry

bbastings:
As the geometry stream is created from json as if it were a part, in order to not require a category, and parts don't support subcategory changes, it could result in decoration geometry that wasn't snappable. The supplier of the non-element geometry should only include geometry for the appropriate subcategory according to the HitDetail so there isn't any need for filtering on the backend.


pmconne (approved):

a-gagnon:
Thanks!!
Am I right in assuming the `main` branch is 4.x? Would it be possible to backport to the latest 3.7.x if that's the case?

bbastings:
Yeah, should be 4.0…

imodel-native wasn’t used in 3.7 (or any release yet I think)…so I don’t know if this can be cherry picked. Feel free to make the change to 3.7 if Caleb says it’s cool and someone will make a new addon. I deleted my imodel02 stuff yesterday since I could no longer build it due to newer versions of rush/npm…

-B

PR #278, opened 2023-05-11
Backport openssl 3.1 

nick4598:
backport of https://github.com/iTwin/imodel-native/pull/276


pmconne:
Explanation why 846 files changes in this PR vs 1036 in #276?

nick4598:
> Explanation why 846 files changes in this PR vs 1036 in #276?

Dan had already updated to some 3.1.0.dev version of openssl and not an official release in main, but that change (3.1.0.dev version) never made it to to release/3.x branch. I think that would explain why diff # of files changed. Do you think that is a sufficient explanation? Or do you have other recommendations for how I could be more confident?

pmconne:
> Do you think that is a sufficient explanation?

If it compiles, yep. I wasn't going to compare 1000+ files.

nick4598:
> > Do you think that is a sufficient explanation?
> 
> If it compiles, yep. I wasn't going to compare 1000+ files.

Nevermind, I did do something wrong. A file was missing according to bb so that was a red flag to me so I did it again locally to see where I may have gone wrong, 
1) git cherry-pick {Dan's SHA}
2) git checkout --theirs . (first had to manually delete one file drbg_lib.c that didn't have a theirs)
3) git add .
4) git cherry-pick --continue since there were lots of merge conflicts in the first cherry-pick (I didn't do this for this branch, but it doesn't seem to make a difference in the # of files changed for my other branch I tried it in) and I'm still missing asn1_dsa.c which does exist in main.

Well while typing this I just realized the reason this is happening is because I don't have dan's original PR to add the 3.1.0-dev version.. maybe I should also cherry-pick that.

nick4598:
new PR: https://github.com/iTwin/imodel-native/pull/279

PR #282, opened 2023-05-16
Convert EC2 Xml schemas to EC 3.2 Xml schemas for TypeScript

akshay05n:
Create an API that can convert EC 2 schemas to EC 3.2 schemas

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/aks/convert-ec2-schemas


iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -4136,6 +4150,9 @@ friend struct SchemaJsonWriter;
     //!Loops through a schema's classes and properties and removes control characters from their descriptions, display labels, and role labels.
     //! @param[in]  schema  pointer to the schema which will be looped through
     ECOBJECTS_EXPORT static void RemoveInvalidDisplayCharacters(ECSchemaR schema);
+
+    //! The schemaXml will be read to obtain the SchemaKey, ECXML Major and Minor versions.
+    ECOBJECTS_EXPORT static SchemaReadStatus ReadSchemaStub(Utf8StringR schemaXml, SchemaKey& schemaKey, uint32_t& ecXmlMajorVersion, uint32_t& ecXmlMinorVersion);
```
rschili, line 4155: On second thought, SchemaStub is a concept we better keep inside this project. What information do you need exactly? If it's only the SchemaKey, then better call it "ReadSchemaKey"

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -2929,6 +2929,19 @@ struct SearchPathSchemaFileLocater : IECSchemaLocater, RefCountedBase, NonCopyab
     ECOBJECTS_EXPORT static SearchPathSchemaFileLocaterPtr CreateSearchPathSchemaFileLocater(bvector<WString> const& searchPaths, bool includeFilesWithNoVerExt=false);
 };
 
+//=======================================================================================
+//! Reads schemas from a given set of Xml strings
+//=======================================================================================
+struct EXPORT_VTABLE_ATTRIBUTE StringSchemaLocater : IECSchemaLocater, NonCopyableClass
```
rschili, line 2935: Now that this is here we should remove the one in TestFixture.h and use this one instead

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3211,6 +3211,18 @@ ECSchemaPtr SearchPathSchemaFileLocater::_LocateSchema(SchemaKeyR key, SchemaMat
...
 
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+ECSchemaPtr StringSchemaLocater::_LocateSchema(SchemaKeyR key, SchemaMatchType matchType, ECSchemaReadContextR schemaContext)
+    {
+    ECSchemaPtr schema;
+    auto schemaXml = m_schemaXml.find(key.m_schemaName);
+    if (schemaXml != m_schemaXml.end())
+        ECSchema::ReadFromXmlString(schema, schemaXml->second.c_str(), schemaContext);
```
rschili, line 3222: This method returns a status code. Now that this is no longer test but production code, let's evaluate the return code and log an error message if it's not success.

iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -1788,6 +1788,33 @@ struct NativeDgnDb : BeObjectWrap<NativeDgnDb>, SQLiteOps
...
 
+    Napi::Value ConvertEC2XmlSchemas(NapiInfoCR info)
+        {
+        RequireDbIsOpen(info);
+        REQUIRE_ARGUMENT_STRING_ARRAY(0, ec2XmlStrings);
+        REQUIRE_ARGUMENT_ANY_OBJ(1, jsOpts);
+
+        JsInterop::SchemaImportOptions options;
+        options.m_schemaLockHeld = jsOpts.Get(JsInterop::json_schemaLockHeld()).ToBoolean();
+        const auto& maybeEcSchemaContextVal = jsOpts.Get(JsInterop::json_ecSchemaXmlContext());
```
rschili, line 1799: I don't see anybody calling this with a context, what's the scenario, try to cover it in a test

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -510,4 +510,23 @@ describe("basic tests", () => {
...
+        <ECClass typeName="TestEntityClass" isDomainClass="true" />
+      </ECSchema>`;
+
+    const refSchema = `<?xml version="1.0" encoding="UTF-8"?>
+      <ECSchema schemaName="RefSchema" nameSpacePrefix="rs" version="1.0" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.2.0">
+        <ECClass typeName="TestStructClass" isStruct="true" />
+      </ECSchema>`;
+
+    assert.isTrue(dgndb.isOpen());
+    const ec3Schemas: string[] = dgndb.convertEC2XmlSchemas([schemaXml, refSchema], { schemaLockHeld: false });
```
rschili, line 527: I assume that's only a dummy test still. Verify the full schema contents, make sure it's EC3 etc..

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
akshay05n: For the current problem, I only need to obtain the SchemaKey from the XML.

iModelCore/ecobjects/src/ECSchema.cpp
akshay05n: Error messages are already logged by the ECSchema::ReadFromXmlString() method if the schema XML string cannot be deserialized or a duplicate schema is detected.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
akshay05n: Removed StringSchemaLocater from TestFixture.h

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
akshay05n: Updated method for extracting only the SchemaKey from the schema XML.

iModelJsNodeAddon/IModelJsNative.cpp
akshay05n: We decided to exclude the ECSchemaReadContext from the parameter list since the caller no longer needs to provide it. Instead, the conversion logic creates its own schema context and uses it for the schema conversion process.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -922,6 +923,70 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
...
+        if (SchemaReadStatus::Success != schemaStatus)
+            return BE_SQLITE_ERROR;
+
+        schemaKeyPairs[i].second = schema;
+        }
+
+    if (0 == schemaKeyPairs.size())
+        return BE_SQLITE_ERROR;
+
+    for (bpair<SchemaKey, ECSchemaPtr>& schemaPair : schemaKeyPairs)
```
ColinKerr, line 972: I think you need to order by dependency order before you convert.  That way base property definitions from a referenced schema will be converted before the derived ones

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -510,4 +510,25 @@ describe("basic tests", () => {
       }],
     });
   });
+
+  it("testConvertEC2XmlSchemas", async () => {
```
ColinKerr, line 514: This is not enough of a test.  It doesn't test any actual conversions that are done by the schema converter and does not test any actual dependencies between schemas.  See existing schema converter tests for an example of a schema that actually has something to convert.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -922,6 +923,70 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
...
+        {
+        if (schemaPair.second != nullptr)
+            {
+            bool conversionStatus = ECSchemaConverter::Convert(*(schemaPair.second), *schemaContext);
```
ColinKerr, line 976: Is this the same API that the dgn connector calls?  I know it's the same class but I cannot remember if they use the static method or if they construct a converter.  Please check and confirm we're doing it the same way as the dgn connector

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -2929,6 +2929,19 @@ struct SearchPathSchemaFileLocater : IECSchemaLocater, RefCountedBase, NonCopyab
...
+//! Reads schemas from a given set of Xml strings
+//=======================================================================================
+struct EXPORT_VTABLE_ATTRIBUTE StringSchemaLocater : IECSchemaLocater, NonCopyableClass
+{
+private:
+    bmap<Utf8String, Utf8String> m_schemaXml;
+protected:
+    ECOBJECTS_EXPORT ECSchemaPtr _LocateSchema(SchemaKeyR key, SchemaMatchType matchType, ECSchemaReadContextR schemaContext) override;
+public:
+    ECOBJECTS_EXPORT void AddSchemaString(Utf8String schemaName, Utf8String schemaXml) {m_schemaXml[schemaName] = schemaXml;}
```
rschili, line 2942: Using Utf8String here is problematic. It means the value gets passed and stored by value, which means the huge schema string gets copied multiple times.
We should probably store this as Utf8String anyways because the caller may delete the string before our locater gets to reading it, or we may make that a documented limitation of the StringSchemaLocater.
In any case, you should change the parameters of this method into either Utf8StringR or Utf8CP

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3211,6 +3211,18 @@ ECSchemaPtr SearchPathSchemaFileLocater::_LocateSchema(SchemaKeyR key, SchemaMat
     return schemaOut;
     }
 
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+ECSchemaPtr StringSchemaLocater::_LocateSchema(SchemaKeyR key, SchemaMatchType matchType, ECSchemaReadContextR schemaContext)
+    {
+    ECSchemaPtr schema;
+    auto schemaXml = m_schemaXml.find(key.m_schemaName);
```
rschili, line 3220: This is hacky, it works for tests, but for published code we should improve this. You have to check the schema key of the schema is actually valid for the request. Check the other schema locater implementations on how to do that.

Basically, use SchemaMatchType to compare the requested key against what you have and then decide to return the schema or not.

We should cover this with tests, too.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -922,6 +931,80 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
     return dgndb.SaveChanges();
     }
 
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+DbResult JsInterop::ConvertEC2XmlSchemas(bvector<Utf8String> const& ec2XmlStrings, bvector<Utf8String>& ec3XmlStrings, ECSchemaReadContextPtr schemaContext)
```
rschili, line 937: DbResult is not a good return code here. The method is independent of ECDb.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -922,6 +931,80 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
...
+
+    for (int i = 0; i < ec2XmlStrings.size(); i++)
+        {
+        Utf8String ec2XmlString = ec2XmlStrings[i];
+        ECSchemaPtr schema;
+        SchemaReadStatus schemaStatus = ECSchema::ReadFromXmlString(schema, ec2XmlString.c_str(), *schemaContext);
+
+        if (SchemaReadStatus::DuplicateSchema == schemaStatus && schemaKeyPairs[i].second == nullptr)
+        {
+            schema = ECSchema::LocateSchema(schemaKeyPairs[i].first, *schemaContext);
```
rschili, line 968: I don't think this is necessary. The incoming schemas are not in dependency order, so loading one of them may automatically load its referenced schemas into the context, and when you call ReadFromXmlString() again, you will run into this DuplicateSchema problem and use the one from the context instead.

Instead of doing this, you could just always call schemaContext.LocateSchema() and not call ReadFromString at all. It should boil down to the same logic and you would not need this additional block here.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -922,6 +931,80 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
...
+
+    bvector<bpair<SchemaKey, ECSchemaPtr>> schemaKeyPairs;
+    StringSchemaLocater locater;
+    for (Utf8String ec2XmlString : ec2XmlStrings)
+        {
+        SchemaKey key;
+        SchemaReadStatus status = ECSchema::ReadSchemaKey(ec2XmlString, key);
+        if (SchemaReadStatus::Success != status)
+            return BE_SQLITE_ERROR;
+        locater.AddSchemaString(key.GetName(), ec2XmlString);
```
rschili, line 955: I mentioned on the locater that it should be aware of the schema keys and not just match the name, so better provide the key here, you can internally still use a map keyed on name in the locater but still remember the key, as you already have it here.
Hmm on second thought, the StringSchemaLocater probably needs to support just names as well, because that's what all our test code will be calling. It should probably have an overload which takes the key and xml, and another which takes the name and xml, but that second internally would call ReadSchemaKey() and remember it.
The locater could then expose a list of its known schema keys, which you could use in the for loop below instead of filling schemaKeyPairs. That way you only have to fill and touch the schemaKeyPairs vector once.

rschili (changes requested):

iModelJsNodeAddon/JsInterop.cpp
akshay05n: Added code to sort the schemas in dependency order prior to conversion

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
akshay05n: Updated parameters of AddSchemaString() method as SchemaKeyCR and Utf8StringCR

iModelJsNodeAddon/JsInterop.cpp
akshay05n: The return code has been changed to BentleyStatus type

iModelJsNodeAddon/JsInterop.cpp
akshay05n: Updated the code to use the ECSchema::LocateSchema() method. The schema is located using SchemaMatchType::LatestWriteCompatible as the match type criteria.

iModelCore/ecobjects/src/ECSchema.cpp
akshay05n: Added check to compare the requested key using `expectedSchemaKey.Matches(key, matchType)`. I also added tests in SchemaStringLocaterTests.cpp file.

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -510,4 +510,202 @@ describe("basic tests", () => {
...
+    assert.equal(ec3Schemas.length, 2);
+    // converted EC3 schemas are always sorted in dependancy order: RefSchema-> TestSchema
+    const ec3RefSchema = ec3Schemas[0];
+    const ec3SchemaXml = ec3Schemas[1];
+
+    assert.isTrue(ec3RefSchema.includes("http://www.bentley.com/schemas/Bentley.ECXML.3.2"));
+    assert.isTrue(ec3SchemaXml.includes("http://www.bentley.com/schemas/Bentley.ECXML.3.2"));
+  });
+
+  it("testConvertEC2XmlSchemas enumeration in ref schema", async () => {
```
rschili, line 536: Adding comment to track. Like I mentioned in teams, I think it would be a good test to put schema things in there which have reserved names, so the converter renames them, and then later the test verifies the things have been renamed (the reserved name thing should be gone). This should be a reliable method to verify the conversion worked successfully.

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
akshay05n: Added additional tests to verify that schema metadata is also converted

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -510,4 +510,23 @@ describe("basic tests", () => {
...
+    const refSchema = `<?xml version="1.0" encoding="UTF-8"?>
+      <ECSchema schemaName="RefSchema" nameSpacePrefix="rs" version="1.0" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.2.0">
+        <ECClass typeName="TestStructClass" isStruct="true" />
+      </ECSchema>`;
+
+    assert.isTrue(dgndb.isOpen());
+    const ec3Schemas: string[] = dgndb.convertEC2XmlSchemas([schemaXml, refSchema], { schemaLockHeld: false });
+    assert.equal(ec3Schemas.length, 2);
+    dgndb.saveChanges();
+    dgndb.closeIModel();
```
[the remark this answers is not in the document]
akshay05n: I did not include ConversionSchema in the schema context for the conversion because I did not see it used in the DgnV8 connector source code during schema conversion.

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3216,11 +3216,25 @@ ECSchemaPtr SearchPathSchemaFileLocater::_LocateSchema(SchemaKeyR key, SchemaMat
...
+        return nullptr;
+    if (expectedECSchema.IsValid())
+        return expectedECSchema;
+
+    // Read schema from Xml string if the cached schema is invalid
+    ECSchemaPtr schemaOut;
+    if (SchemaReadStatus::Success != ECSchema::ReadFromXmlString(schemaOut, expectedSchemaXml.c_str(), schemaContext))
+        return nullptr;
+    m_knownSchemas[key.GetName()] = std::make_tuple(expectedSchemaXml, expectedSchemaKey, schemaOut);
+    return schemaOut;
```
ColinKerr, line 3237: Schema locaters do not need to cache the schemas it loads, the SchemaContext the locater is added to does the caching.  So this doesn't need to be so complex

iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -2612,6 +2612,89 @@ struct NativeRevisionUtility : BeObjectWrap<NativeRevisionUtility>
...
+struct NativeSchemaUtility : BeObjectWrap<NativeSchemaUtility>
+    {
+    private:
+        DEFINE_CONSTRUCTOR
+
+    public:
+        NativeSchemaUtility(NapiInfoCR info) : BeObjectWrap<NativeSchemaUtility>(info) {}
+        ~NativeSchemaUtility() {SetInDestructor();}
+
+    static Napi::Value ConvertCustomAttributes(NapiInfoCR info)
```
ColinKerr, line 2628: Both methods here and in JsInterop are 95% the same.  You should be able to refactor them to share code.  For example make a one method that takes a flag to control if CAs are converted in both classes then have the two exposed Napi methods call one common method with different flags.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -2929,6 +2929,20 @@ struct SearchPathSchemaFileLocater : IECSchemaLocater, RefCountedBase, NonCopyab
...
 };
 
+//=======================================================================================
+//! Reads schemas from a given set of Xml strings
+//=======================================================================================
+struct EXPORT_VTABLE_ATTRIBUTE StringSchemaLocater : IECSchemaLocater, NonCopyableClass
+{
+private:
+    bmap<bpair<SchemaKey, SchemaMatchType>, ECSchemaPtr> m_knownSchemas;
+    bmap<Utf8String, Utf8String> m_schemaStrings;
```
ColinKerr, line 2939: You should be able to get by with one variable that holds the schema strings by their key.  As noted schema locaters should not cache the deserialized schema, that is the schema contexts job.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -3106,6 +3120,7 @@ struct ECSchema : RefCountedBase, IECCustomAttributeContainer
 friend struct ECSchemaDownConverter;
 
 friend struct SearchPathSchemaFileLocater;
+friend struct StringSchemaLocater;
```
ColinKerr, line 3123: Why does StringSchemaLocater need to be a friend of ECSchema?

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -2935,12 +2935,11 @@ struct SearchPathSchemaFileLocater : IECSchemaLocater, RefCountedBase, NonCopyab
 struct EXPORT_VTABLE_ATTRIBUTE StringSchemaLocater : IECSchemaLocater, NonCopyableClass
 {
 private:
-    bmap<bpair<SchemaKey, SchemaMatchType>, ECSchemaPtr> m_knownSchemas;
-    bmap<Utf8String, Utf8String> m_schemaStrings;
+    bmap<Utf8String, bpair<SchemaKey, Utf8String>> m_schemaStrings;
```
ColinKerr, line 2938: Why not make it a map of SchemaKey and schema xml string?  For example ECSchemaCache uses a SchemaMap which has a SchemaKey as the key

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3211,6 +3211,37 @@ ECSchemaPtr SearchPathSchemaFileLocater::_LocateSchema(SchemaKeyR key, SchemaMat
...
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+ECSchemaPtr StringSchemaLocater::_LocateSchema(SchemaKeyR key, SchemaMatchType matchType, ECSchemaReadContextR schemaContext)
+    {
+    // Get cached version of the schema
+    ECSchemaPtr schemaOut = schemaContext.GetFoundSchema(key, matchType);;
+    if (schemaOut.IsValid())
+        {
+        return schemaOut;
+        }
```
ColinKerr, line 3224: The schema locater should not look in the context.  It should just load the schema on request.  The schema context will check it's cache before calling the locaters registered to the context

iModelCore/ecobjects/test/NonPublished/SchemaStringLocaterTests.cpp
```diff
@@ -0,0 +1,87 @@
...
+    ASSERT_TRUE(schema.IsValid());
+
+    auto fruit = schema->GetClassCP("Fruit");
+    ASSERT_NE(nullptr, fruit);
+    }
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+TEST_F(SchemaStringLocaterTests, LocatingSchemaContext_SchemaXmlIsDeleted)
```
ColinKerr, line 65: What is the point of this test?

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
rschili: I suggested this approach in a call we had today because when keying on SchemaKey, we cannot use the map's find method to obtain matchtypes other than "Exact".
We'd have to iterate and check every single key, which would kill the purpose of a map. However, now I wonder how SchemaCache does it.

iModelCore/ecobjects/src/ECSchema.cpp
rschili: This call is wrong, however, the other locaters check the context before they attempt to load a schema. Like, if Schema XY version 1.0.0 matchType "Latest" is requested, and our locater can provide version 3.0.0, it can use GetFoundSchema(<for 3.0.0>) before actually attempting to load that one.
I also first thought this is unnecessary, but the other locaters seem to do this as well.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
ColinKerr: To be honest I didn't check to see what SchemaMap is doing, they could have an exact match and using find_if when looking at match type.  But you can also construct the bmap with a SchemaKeyLessThan that has the match type you want to use:
```cpp
bmap<ECN::SchemaKey, ECN::ECSchemaCP, SchemaKeyLessThan<SchemaMatchType::Latest>> map;
```

iModelCore/ecobjects/src/ECSchema.cpp
ColinKerr: The SearchPathSchemaFileLocater is overly complex. 

If version 3.0.0 was already in the cache then it would have been loaded from the cache before this locater was called.  I think it's possible that this code exists in the SearchPathSchemaFileLocater but is no longer necessary.  This code has gone through a lot of permutations.  I went back a bit in the history and this has been in the code since very early, I didn't dig into it enough to understand how things worked at that point but it was a lot different.  Let's try without it.

iModelCore/ecobjects/test/NonPublished/SchemaStringLocaterTests.cpp
akshay05n: This test ensures that schema XML strings are retained in memory so they can be accessed even if the original source strings are deleted.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
akshay05n: I originally included StringSchemaLocater as a friend of ECSchema because SearchPathSchemaFileLocater was also a friend of ECSchema. However, this is not necessary, so I have removed StringSchemaLocater from the list of friends of ECSchema.

iModelCore/ecobjects/src/ECSchema.cpp
akshay05n: The updated schema locator no longer makes a call to the schema context method.

ColinKerr (approved):

pmconne (approved):

PR #288, opened 2023-05-18
Add `baseUri` and `storageType` members to CloudContainer

kabentley:
Also, `blockSize` should not be optional on `intializeContainer`.

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/default-block-size


iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -410,11 +412,11 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
         JsCloudUtil handle;
         auto result = handle.Init(*this);
         if (result.IsSuccess()) {
-            int blockSize = 0;
+            int blockSize = 64 * 1024;
```
nick4598, line 415: Maybe a nitpick but previously, passing a blocksize of 0 to handle.initializeContainer and by extension, sqlite3_bcv_create would mean we fell back to BCV_DEFAULT_BLOCKSIZE. I guess it doesn't really matter anymore since blockSize is no longer optional (and I guess we should never hit this as a result?) in your itwinjs-core branch, but would 4 * 1024 * 1024 (current value of BCV_DEFAULT_BLOCKSIZE) be more appropriate?

nick4598:
@kabentley Just curious, do you plan to use storageType somehow in the future since you separated it from the customuri&sas params? Or is this more because it just seems like a property that the CloudContainer should have available?

iModelJsNodeAddon/JsCloudSqlite.cpp
kabentley: The sqlite default, BCV_DEFAULT_BLOCKSIZE of 4Mb is really only appropriate for large and rarely changing databases. It works well for checkpoints and workspaces. But for most other uses of CloudSqilte we've found that a much smaller blocksize works better. 

I have now made it so that blockSize is required and there is no default. People should have to pick the appropriate blockSize (though I think 64K and 4Mb should be the only choices because CloudCaches cannot mix blockSizes.)

kabentley:
> @kabentley Just curious, do you plan to use storageType somehow in the future since you separated it from the customuri&sas params? Or is this more because it just seems like a property that the CloudContainer should have available?

I just wanted to make it a property so you can test it at runtime. Combining it with "uriParams" (which is only relevant for Azure) is confusing.

nick4598 (approved):

pmconne (approved):

PR #294, opened 2023-05-25
Fix clang ubuntu build

pmconne:
Using clang 13, 14, or 15 on ubuntu 22.04 to build locally. I observe errors where code is trying to declare a class member of type `vector<T>` where `T` is an incomplete (forward-declared) type. AFAICT this is a correctly-reported error per C++20.
Fix is easy, though I question why an alleged "C API" uses `vector` at all.

@MichaelBelousov observes the same. @markschlosseratbentley for some reason does not.
CI does not. Unclear why. It's using clang 13. I'd be interested to know what version of libstdc++ it's using.


markschlosseratbentley (approved):

MichaelBelousov (approved):
builds locally

dassaf4 (approved):
FWIW, "capi" is just the name we gave to what started out as the MDL API (fdf files). This API eventually took on C++ features when everything started getting compiled as C++.

PR #297, opened 2023-05-26
Merge ares 1.19.1 for CVE

chuckkir:


tcobbs-bentley:
@chuckkir @nick4598 Is the fact that this isn't merged possibly causing SSL errors on Android?

nick4598:
> @chuckkir @nick4598 Is the fact that this isn't merged possibly causing SSL errors on Android? (Android checkpoint downloads are failing with an SSL error.)

It shouldn't since ares is not compiled on android.

chuckkir:
Also it doesn't run on the platforms it is supposed to build on so not merged is currently the preferred state until I figure out what is wrong.

pmconne (approved):

PR #298, opened 2023-05-26
Ensure FederationGuids and delete-prevention on auto-created elements.

diegoalexdiaz:
The RootSubject, DictionaryModel and RealityDataSourcesModel are the only elements in an iModel currently not being assigned a FederationGuid. That is because they are created earlier than when handlers are initialized. This PR is meant to address that inconsistency. We discussed it with the transformation and iModelHub teams and they don't foresee any issues. They will test with the code from the branch of this PR before merging it into master (that's why this PR is created initially as a draft).

While updating the corresponding test-case, I noticed that iModelPlatform doesn't prevent the deletion of the RealityDataSourcesModel, which it does for the RootSubject/RepositoryModel and DictionaryModel. That has been like that since the RealityDataSourcesModel was introduced in bim02 (2016). However, the RealityDataSourcesModel, like the other two, is not meant to be deleted. 

Addressing that issue in the same way delete-prevention was implemented for the other two classes would require introducing a new class+handler for RealityDataSources in BisCore/iModelPlatform, plus data-update-code for existing iModels. The simpler way (implementation wise) proposed in this PR consists of just reusing the existing LinkModel handler (which is the one currently instantiated for the RealityDataSourcesModel) and check for that case on its _OnDelete callback via the hard-coded ElementId in iModelPlatform for the RealityDataSourcesModel.


iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -511,6 +511,12 @@ DgnElementCPtr DgnElements::InsertElement(DgnElementR element, DgnDbStatus* outS
 
     DgnDbStatus ALLOW_NULL_OUTPUT(stat,outStat);
 
+    if ( !element.m_federationGuid.IsValid())
```
DanRod1999, line 514: @pmconne, when you said we don't check if a guid is valid before inserting, was this what you had in mind when suggesting we should check? Based on what I found while looking around, the `fromString()` function does seem to check that the guid is valid or else it fails, so I don't think we can insert a bad guid that way.

pmconne: I was referring to [this](https://github.com/iTwin/imodel-native/blob/c11f767346b77bdae40811793599a8be758dfc00/iModelJsNodeAddon/JsInteropDgnDb.cpp#L375)

```
        // if no federationGuid was supplied, create one for the element before we add it.
        if (!inJson.isStringMember(json_federationGuid()))
            el->SetFederationGuid(BeGuid(true));
```

If the TypeScript caller does, e.g., `insertElement({ ...props, federationGuid: "not a valid guid" })`, then the code above will not assign a new, valid guid to the element.

DanRod1999: going to make a separate PR for this then since this change would not be needed to BP to imodel02, but the rest of these changes would

nick4598:
Regarding the description of this PR:  "We discussed it with the transformation and iModelHub teams and they don't foresee any issues. They will test with the code from the branch of this PR before merging it into master (that's why this PR is created initially as a draft)." , I believe it can now be merged since Mindaugas tested it a few weeks back. 

The email that Mindaugas sent was: 

> "I ran the transformation with iModel where I have added federation guids to RootSubject, DictionaryModel and RealityDataSources and transformation successfully passed. I’ve tested on both currently used imodeljs native which we use in production (version 3.6.1), our migration branch to itwinjs 4.x (version 4.1.1) and using native build from Diego’s branch.
> 
> The only thing notable this is that core transformer skips processing of those three elements as there are expected differences that can be present in those elements like CodeValue of RootSubject. So we might need to consider transferring federation guids for these three elements during the transformations, but adding a manual step to update federation guids didn’t cause any problems.
> 
> So, it seems that everything is good regarding to adding federation guids to RootSubject, DictionaryModel and RealityDataSources from transformations side. They don’t cause any issues in the current version and don’t cause any problems with Diego’s changes."

@mindaugasdirg Is my understanding correct that core-transformer still needs a change on its side (separate from this PR) to properly handle fedguids on RootSubject, DictionaryModel and RealityDataSources? Or has that since been handled? When you say that you've added a manual step to update federation guids and it didn't cause any problems, where did you add that step? Is that in the transformation service itself? Could you point me to that code?

nick4598 (approved):

iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp
```diff
@@ -268,10 +270,11 @@ DbResult DgnDb::CreateRootSubject(CreateDgnDbParams const& params)
     DgnModelId modelId = DgnModel::RepositoryModelId();
     CodeSpecId codeSpecId = CodeSpecs().QueryCodeSpecId(BIS_CODESPEC_Subject);
     DgnCode elementCode = DgnCode(codeSpecId, elementId, params.m_rootSubjectName, m_codeValueBehavior);
+    BeGuid federationGuid(true);
```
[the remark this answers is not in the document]
diegoalexdiaz: The question was raised and discussed at a BWG meeting a couple of times. The decision was not to keep them in-sync as they are really two different things: 1) The Identifier of an iModel (i.e. a BIS repository) in iModelHub vs. 2) The Identifier of the element (subject) in that iModel that is meant to describe what the BIS repository is about.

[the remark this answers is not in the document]
MichaelBelousov: sounds good to me, thanks

mindaugasdirg:
> @mindaugasdirg Is my understanding correct that core-transformer still needs a change on its side (separate from this PR) to properly handle fedguids on RootSubject, DictionaryModel and RealityDataSources? It seems like a change that should exist in core-transformer itself.. but I don't know. Or has that since been handled? When you say that you've added a manual step to update federation guids and it didn't cause any problems, where did you add that step? Is that in the transformation service itself? Could you point me to that code?

@nick4598,  I agree, in my quick testing, I updated federation guids manually after the transformation finished processing everything, but the proper way would be to update transformer to not skip those elements. The skipping of these default elements are specified here. https://github.com/iTwin/imodel-transformer/blob/093be1ade668f683562baf302d2aae866de3032b/packages/transformer/src/IModelImporter.ts#L105C27-L105C27. Although the proper way of updating these elements would be to update only some of the properties, since CodeValue of RootSubject would be accidently updated to the old iModel's name

ColinKerr (approved):

PR #307, opened 2023-06-06
wrap import error of native addon with suggestion to install prerequisites

### LINKED ISSUE #287: Wrap import error of native addon on windows with suggestion to install MSVC runtime
Occasionally on windows desktops people seem to run into an issue where node (or electron) throws an error when loading the native addon, with the windows error message `The specified module could not be found`.

image
https://github.com/iTwin/imodel-native/blob/78720da4c7fd133a891d14a27bb1a013bbdeb564/iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts#L94

This happens because when node.js/electron requires the native addon, windows fails to load the addon DLL due to a missing dependency of the DLL, frequently this is just the [Microsoft Visual C++ Runtime](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170) not being installed on the system. I have not identified any other required DLLs dependencies that windows 10/11 shouldn't already have installed.

The require occurs [here](https://github.com/iTwin/imodel-native/blob/78720da4c7fd133a891d14a27bb1a013bbdeb564/iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts#L94)

Could it be wrapped in a clearer platform-specific warning? like so:

```ts
try {
  this._nativeLib = require(`./${NativeLibrary.archName}/${NativeLibrary.nodeAddonName}`) as typeof IModelJsNative;
} catch (err) {
  if (/The specified module could not be found/.test(err.message) && process.platform === 'win32')
    err.message += "\niTwin.js requires that Microsoft Visual C++ Runtime be installed."
      + "\nThe latest version is available at Microsoft's website."
      + "\nhttps://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist";
  throw err;
}
```

MichaelBelousov:
closes #287

### DIFF
--- a/iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
+++ b/iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
@@ -91,7 +91,15 @@ export class NativeLibrary {
   public static get nativeLib() { return this.load(); }
   public static load() {
     if (!this._nativeLib) {
-      this._nativeLib = require(`./${NativeLibrary.archName}/${NativeLibrary.nodeAddonName}`) as typeof IModelJsNative; // eslint-disable-line @typescript-eslint/no-var-requires
+      try {
+        this._nativeLib = require(`./${NativeLibrary.archName}/${NativeLibrary.nodeAddonName}`) as typeof IModelJsNative; // eslint-disable-line @typescript-eslint/no-var-requires
+      } catch (err: any) {
+        if (/The specified module could not be found/.test(err.message) && process.platform === "win32")
+          err.message += "\nThis error may occur when trying to run an iTwin.js backend without"
+            + " having installed the prerequisites. See the following link for all prerequisites:"
+            + "\nhttps://www.itwinjs.org/learning/supportedplatforms/#backend-prerequisites";
+        throw err;
+      }
       if (this.isDevBuild)
         // eslint-disable-next-line no-console
         console.log("\x1b[36m", `using dev build from ${__dirname}\n`, "\x1b[0m");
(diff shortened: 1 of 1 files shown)


pmconne:
Why only if windows?

MichaelBelousov:
> Why only if windows?

I believe the message tested against is win32 API specific and figured if we're checking against it, we should restrict to windows only. But I'm open to removing the platform check and just attaching this message on all platforms when the addon fails to load.

pmconne (approved):

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -91,7 +91,14 @@ export class NativeLibrary {
   public static get nativeLib() { return this.load(); }
   public static load() {
     if (!this._nativeLib) {
-      this._nativeLib = require(`./${NativeLibrary.archName}/${NativeLibrary.nodeAddonName}`) as typeof IModelJsNative; // eslint-disable-line @typescript-eslint/no-var-requires
+      try {
+        this._nativeLib = require(`./${NativeLibrary.archName}/${NativeLibrary.nodeAddonName}`) as typeof IModelJsNative; // eslint-disable-line @typescript-eslint/no-var-requires
+      } catch (err: any) {
+        err.message += "\nThis error may occur when trying to run an iTwin.js backend without"
+          + " having installed the prerequisites. See the following link for all prerequisites:"
```
aruniverse, line 98:
```suggestion
          + "\nhaving installed the prerequisites. See the following link for all prerequisites:"

```

aruniverse: no worries, just a nit. i assume you broke this error msg string up for formatting

kabentley (approved):

MichaelBelousov:
possibly flaky tile admin test: https://github.com/iTwin/itwinjs-backlog/issues/753

PR #314, opened 2023-06-12
Linear sweep produces excessive facets

pmconne:
Test exhibiting potential cause of https://github.com/iTwin/itwinjs-backlog/issues/758.
Given a chord tolerance of 100.0, 0.1, or 0.01, the same number of points are produced.
@dassaf4 does this look to you like a bug in the facetter, or am I doing something wrong or have incorrect expectations?


dassaf4:
I'd say incorrect expectations.

Usually in cases like these, it's some other facet option that is dominating max chord length, because the `FacetOptions` constraint that yields the most strokes wins. As I mentioned to @Josh-Schifter and @DStradley back in April, setting other constraints to zero/max can ensure they don't dominate (e.g., angleTol = 0 or 2pi).

But in this case, the constraint that wins is B-spline specific. Each curve primitive can specify a min stroke count, and a Bezier curve's min stroke count is hardcoded at `degree` segments. Each of the holes in the extrusion solid's base parity region is represented by two semicircular cubic B-spline curves with ~34 poles, each of which translates to 31 Bezier curves, each of which gets stroked by degree=3 chords. That's ~200 stroke points minimum per hole. The range of max chord lengths tested do not exceed this minimum stroke count; you'd have to specify chord tol < 8.11e-05 to get more facets for this data.

If the holes were instead specified by true arcs, their stroke counts would be controllable by a larger range of max chord lengths (for sufficiently large values of angleTol and maxEdgeLength) since an arc's min stroke count is 4.

PR #327, opened 2023-06-22
Configurations to Consume OpenSSL in iTwinNativeThirdParty 

naveedkhan8067:
This upgrade allow us to consume relevant configurations of OpenSSL component in iTwinNativeThirdParty. The reason of doing this upgrade is that now we want to consume the duplicate/similar components from a central location.

Following statement is not added to the `BeOpenSSL.mke`:
    `$(o)rsa_sup_mul$(oext) : $(vendorSrc)crypto/bn/rsa_sup_mul.c ${MultiCompileDepends}`
The reason is that `rsa_sup_mul.c` file is not present at `./openssl/vendor/crypto/bn/` in imodel-native. It **could** be due to the newer version.


nick4598:
The missing rsa_sup_mul.c file shouldn't be a problem, that file was deleted in 3.1.1.

chuckkir (approved):

nick4598 (approved):

aruniverse (approved):

iModelCore/libsrc/openssl/BeOpenSSL.mke
```diff
@@ -29,9 +29,15 @@ BUILD_NO_STACK_CHECK = 1   # Don't want to modify third party code, but hopefull
 #----------------------------------------------------------
 %include mdl.mki
 
-appName = iTwinOpenSSL
 vendorSrc = $(_MakeFilePath)vendor/
-o = $(OutputRootDir)build/iTwinOpenSSL/
+
+%if defined(iTwinNativeThirdParty)
+  appName = BeOpenSSL
```
aruniverse, line 35: just want to make sure this is right and not swapped

chuckkir: Thanks for checking. The iTwinNativeThirdParty builds the NuGets that are Be*.dll. Imodel-native builds iTwin*.dll. So if iTwinNativeThirdParty is defined, then it builds Be*.dll. Also, the build would fail if the DLL was incorrectly named.

aruniverse: the iTwin prefix is not confusing at all haha; thx for explaining

PR #329, opened 2023-06-22
Hide curl/cares symbols so that we don't segfault depending on our node version

nick4598:


aruniverse (approved):

nick4598:

PR #330, opened 2023-06-22
prevent heap corruption in mdlBspline_validateCurveKnots

dassaf4:
Fixes iTwin/imodel-native-internal#112

Add new test to verify fix (fails with old code).

### DIFF
--- a/iModelCore/GeomLibs/geom/src/bspline/bspmisc.cpp
+++ b/iModelCore/GeomLibs/geom/src/bspline/bspmisc.cpp
@@ -217,7 +217,7 @@ BsplineParam*   pParams         // <=> potentially modified only if poles given
             if (pKnots[i] - *pStart >= tol)
                 break;
 
-        for (i = numKnots - degree, nExcess1 = 0; i >= 0; i--, nExcess1++)
+        for (i = numKnots - pParams->order - 1, nExcess1 = 0; i >= 0; i--, nExcess1++)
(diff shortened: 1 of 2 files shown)


iModelCore/GeomLibs/geom/src/bspline/bspmisc.cpp
```diff
@@ -217,7 +217,7 @@ BsplineParam*   pParams         // <=> potentially modified only if poles given
             if (pKnots[i] - *pStart >= tol)
                 break;
 
-        for (i = numKnots - degree, nExcess1 = 0; i >= 0; i--, nExcess1++)
+        for (i = numKnots - pParams->order - 1, nExcess1 = 0; i >= 0; i--, nExcess1++)
```
saeeedtorabi, line 220: optional:
If `numKnots - pParams->order - 1` have a meaning, you may create a variable for it.

saeeedtorabi (approved):

iModelCore/GeomLibs/geom/src/bspline/bspmisc.cpp
dassaf4: Good advice; this is the index of the penultimate knot, but not traditionally named, so I'll leave it.  My problem was seeing order-1 and replacing that with degree reflexively. But in fact, the quantity is actually -order-1 = -(degree+1)-1 = -degree-2.  So the regression was off by two!

PR #332, opened 2023-06-27
Add support for ECSQL views

khanaffan:
# General info
This PR combined 2 big features. Since it requires a profile update, and we have one other PR which also needs a profile update, we decided to roll it out like this:

- This PR will only make code changes and no profile update (so Views and custom attributes will not be exposed yet)
- The other PR will do the same
- We will later after at least 1 addon build create another PR which combines the necessary profile changes, and all the documentation. This is the point where the functionality added will become available to be used.

That means merging this PR will include many changes, but the new features will remain unavailable until we roll out the combined profile update.

# ECSQL Views

1. Allow the user to define an abstract class with a custom attribute `ECDbMap.View`. The class must not have any derived classes.
2. In `View` custom attribute user can provide an ECSQL query.
3. Query property and type should match the view class definition.
4. System properties e.g. `ECInstnaceId`, `ECClassId` can be returned by view query but is not required to be defined in case. They can be accessed via view class.
5. If view query does not return system properties, then the view class will not have it. `SELECT * FROm ts.ViewClass` will only return data properties.
6. All types of properties and computed expressions can be returned by view query if class definition defines those properties and their types correctly.
7. The PR adds a new system schema `ClassView.01.00.00` and this **PR will need profile version change for ecdb**.

Following is an example of ECSQL view definition
```xml
    <ECSchema
            schemaName="test_schema"
            alias="ts"
            version="1.0.0"
            xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.2">
        <ECSchemaReference name='ECDbMap' version='02.00.03' alias='ecdbmap' />
        <ECEntityClass typeName="SchemaClassesView" description="" displayLabel="" modifier="Abstract">
            <ECCustomAttributes>
                <View xmlns="ECDbMap.02.00.03">
                    <Query>
                        SELECT
                            cd.ECInstanceId,
                            cd.ECClassId,
                            sc.Name SchemaName,
                            cd.Name ClassName
                        FROM meta.ECSchemaDef sc
                            JOIN meta.ECClassDef cd ON cd.Schema.Id=sc.ECInstanceId
                        GROUP BY sc.Name, cd.Name
                        LIMIT 10
                    </Query>
                </View>
           </ECCustomAttributes>
            <ECProperty propertyName="SchemaName" typeName="string" />
            <ECProperty propertyName="ClassName"  typeName="string"/>
        </ECEntityClass>
    </ECSchema>
```
    
We will roll out this feature hidden at first, because the PR has a huge amount of changes. Final release will require a profile upgrade which will update ECDbMap and ECDbMeta. We will do this in a separate step afterwards.

# ECSQL Custom Attributes

There are now classes in ECDbMeta which allow querying custom attributes in ECSql. The classes utilize the new ECSql View mechanism.

Example, imagine a schema with this custom attribute:
```xml
        <ECCustomAttributeClass typeName="ClassAlias" modifier="Sealed" appliesTo="EntityClass">
            <ECProperty propertyName="Alias" typeName="string" />
        </ECCustomAttributeClass>
        <ECEntityClass typeName="Foo">
            <ECCustomAttributes>
                <ClassAlias>
                    <Alias>F</Alias>
                </ClassAlias>
            </ECCustomAttributes>
        </ECEntityClass>
``` 

The following sql can be used to obtain the first ECClass which has this custom attribute and the Alias "F":

```sql
SELECT e.Name
  FROM meta.ClassCustomAttribute ca
  JOIN meta.ECClassDef e ON ca.Class.Id = e.ECInstanceId
  WHERE ca.CustomAttributeClass.Id IS (TestSchema.ClassAlias) AND json_extract(ca.ValueJson, '$.ClassAlias.Alias')='F'
```


ColinKerr:
What if we required that the class be sealed and have the ClassMap applied with MapStrategy NotMapped.  

If we did this then EC would enforce that a view class could not have children.  It would also be clear that the class was not mapped to a table.

One downside is that the class is mapped just to a view.  We cannot add new MapStrategy options because older versions of ECDb will not allow unknown values.  How could we  modify ECDb to allow more changes in the future?  For example what if we allowed unknown MapStrategy values if the profile was schema import incompatible.  Selecting a class with an unknown map strategy would have to have some well defined behavior.  So the file could be opened and read but in a limited capacity.

iModelCore/ECDb/ECDb/ClassViews.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return false;
+    }
+    return invalidViews == 0;
+}
+//======================================================================================
+// @bsimethod
+//======================================================================================
+bool ClassViews::IsValid(ECN::ECClassCR viewClass, ECDbCR conn){
+    auto viewClassName = viewClass.GetFullName();
+    if (!viewClass.IsEntityClass() && !viewClass.IsRelationshipClass()) {
```
ColinKerr, line 100: you allow the custom attribute to be applied to a relationship class but do not have a test when applied to a relationship class

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
```diff
@@ -0,0 +1,631 @@
...
+                    <Query>
+                        SELECT
+                            cd.ECInstanceId,
+                            cd.ECClassId,
+                            sc.Name SchemaName,
+                            cd.Name ClassName
+                        FROM meta.ECSchemaDef sc
+                            JOIN meta.ECClassDef cd ON cd.Schema.Id=sc.ECInstanceId
+                        GROUP BY sc.Name, cd.Name
+                        LIMIT 10
```
ColinKerr, line 55: What is a LIMIT going to do to the view?   Will it limit to 10 rows?

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
```diff
@@ -0,0 +1,631 @@
+/*---------------------------------------------------------------------------------------------
```
ColinKerr, line 1: I'd like to see tests for:

- Return data and cover all primitive types.
- structs and arrays (should fail as they are not supported)
- Navigation props (are they supported)
- Are any implicit conversions supported if there is a mismatch between view property and selected property?
- Relationship view class
- Metadata mismatch like different persistence units
- instance query (is it supported?)

rschili:
> What if we required that the class be sealed and have the ClassMap applied with MapStrategy NotMapped.

I like that idea more than having it abstract, as it would reduce the unwanted things people could get away with when importing this using an older engine.

rschili:
@ColinKerr  Thinking about this, I realize we are actively avoiding introducing new concepts to EC.
Most of our questions and problems would not apply, if this were a new first class ECSchema item concept, independent of EntityClass.
In that case, older ECDbs would not pick it up, not offer it in UI, but still be able to load the schema, as unknown things are being ignored.
And for ECDb it would be easier to handle as well, because we would not have to filter all classes by a custom attribute to obtain views, instead we would have an ec_View table as a real first class citizen.

I'm not saying we should go for this. But would that not be the straightforward approach? Putting aside the additional development work for a moment.

iModelCore/ECDb/ECDb/ClassViews.cpp
khanaffan: Yes, tests for relationships and a bunch of other stuff are still pending. Since now we allow struct/struct array etc we also need to test for those. We also do not support CTE in the query, but I have no check or test for it. We either must fix that or fail at it. We also missed the test of nonprimitive data return by one view that is referenced by another view.

I have not gotten to them yet. But here is definitely a lot of cases that are not covered yet.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
khanaffan: Yes. It's just a SELECT query with all its capabilities. You can have UNION and LIMIT. It could be even a CTE or recursive query.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
khanaffan: Sure, anything you can do with ECSQL SELECT can be done in View.  View Query is substituted for View Class in query and all type resolutions will happen as in any sub query. We need tests for all above 

My only concerns are edge cases e.g. Conflict between an "Alias" used in "View sub query" that is also used in "query that references the view" and the user might get an error where he does not see a problem with the query but its failing when the view is substituted. To sandbox the view query, we need to change all the aliases used in the subquery to add prefixes, so they do not collide with the primary query.

khanaffan:
> 

From your post, i reminded me of the following missing check for a valid view.
1. Class must be sealed 
2. Class cannot be used as the endpoint of a persistent relationship. It can be the endpoint of a relationship view.
3. If the Class is defined as a Relationship. Then its endpoint can only be a view class.
4. There cannot be any other System custom attribute applied to a Class that has View CustomAttribute.

About your other point about `ECClassId`. Our `ECInstnaceId` only unique w.r.t `ECClassId`. If we return ViewClass `ECClassId` and return `ECInstanceId` from the view query we are breaking that ECInstanceId/ECClassId rule. This will lead to a lot of problems where `ECClassId` is important to the user. e.g. `InstanceProp` will simply not work. You can take any case where the user uses ECClassId to query the full instance of a row in the view. View themselves are just an alias to a query and they should have no other persisted or identifiable data. You should not be able to tell if an EC instance came from a query, view, or entity class. Only the "Persisted view" has its own unique Rowid, ClassId and you can create indexes and constraints on it. 

The view might not align with the definition of an `EntityClass` or `ECRelationshipClass` in that its System columns are not its own but that is what non-persisting views are. Even SQL view does not have `RowId` you can derive one from underlying tables but the whole concept of "Primary key" is really associated with "Persistence" as it's a way to find a row on disk.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
```diff
@@ -0,0 +1,631 @@
...
+        ASSERT_EQ(stmt.GetColumnCount(), 3) << "system prop is not returned as its not selected ";
+        stmt.Finalize();
+        m_ecdb.AbandonChanges();
+    }
+
+}
+// }
+//         listener.Dump("listener");
+
+// // Add issue listener
```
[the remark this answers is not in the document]
khanaffan: fixed

iModelCore/ECDb/ECDb/ClassViews.cpp
khanaffan: added linktable and nav property test.

iModelCore/ECDb/ECDb/ECSql/ECSqlFieldFactory.cpp
```diff
@@ -111,6 +111,38 @@ ECSqlColumnInfo ECSqlFieldFactory::CreateColumnInfoForProperty(ECSqlPrepareConte
...
+        ECTypeDescriptor typeDescriptor = DetermineDataType(dateTimeInfo, structType, ctx.Issues(), *ecProperty);
+        ECSqlPropertyPath propertyPath;
+        if (isGenerated)
+            {
+            propertyPath.AddEntry(*generatedProperty);
+            rootClass = &generatedProperty->GetClass();
+            }
+        else
+            propertyPath.AddEntry(*ecProperty);
+        return ECSqlColumnInfo(
```
rschili, line 147: @khanaffan can you review this piece? This is another parent parent parent situation, I hope there is a more robust way to do this.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -4163,13 +4345,39 @@ ECSqlParseContext::ParseArg const* ECSqlParseContext::CurrentArg() const
 //+---------------+---------------+---------------+---------------+---------------+--------
 void ECSqlParseContext::PopArg() { m_finalizeParseArgs.pop_back(); }
 
+//Helper for TryResolveClass to check if te current node is a table node inside a type predicate
+//If the class is a custom attribute class,we allow it to be used despite what the policy says
+//Example syntax in ECSQL: "IS (mySchema.MyCustomAttribute)"
+bool IsTableNameInTypePredicateSpecialCase(ClassMap const& classMap, OSQLParseNode const& node)
```
rschili: @khanaffan I recall you had a comment about this section, but we never got to talk about it. Do you have a better way in mind?

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
```diff
@@ -1317,217 +1341,248 @@ ECSqlStatus ECSqlExpPreparer::PrepareRelationshipJoinExp(ECSqlPrepareContext& ct
...
-    if (toRelatedIdSqlVisitor.GetResultSet().size() != 1)
-        {
-        BeAssert(false);
-        return ECSqlStatus::Error;
-        }
-    sql.Append(toRelatedIdSqlVisitor.GetResultSet().front().GetSqlBuilder().GetSql());
-    }
+    sql.Append(toNativeSqlSnippets.front());
+
+    // ///Resolve direction of the relationship
```
rschili, line 1377: @khanaffan large commented piece of code. Do we still need this?

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
```diff
@@ -2781,6 +2838,104 @@ ECSqlStatus ECSqlExpPreparer::GenerateECClassIdFilter(Utf8StringR filterSqlExpre
     return ECSqlStatus::Success;
     }
 
+//-----------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+//static
+ECSqlStatus ECSqlExpPreparer::PrepareNavValueCreationFuncExp(NativeSqlBuilder::List& nativeSqlSnippets, ECSqlPrepareContext& ctx, NavValueCreationFuncExp const& exp)
+    {
+    if (!ctx.GetECDb().Schemas().GetClass(exp.GetClassRefExp()->GetAs<ClassNameExp>().GetSchemaName(), exp.GetClassRefExp()->GetAs<ClassNameExp>().GetClassName())->GetPropertyP(exp.GetColumnRefExp()->GetAs<DerivedPropertyExp>().GetExpression()->GetAs<PropertyNameExp>().GetPropertyPath()[0].GetName())->GetIsNavigation())
```
rschili, line 2847: This looks like it can use some cleanup. It is long and hard to read and making several assumptions about the types of Exp objects which I'm not sure will always be correct.

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
```diff
@@ -269,6 +274,8 @@ BentleyStatus PropertyNameExp::ResolveLocalRef(ECSqlParseContext& ctx) {
 //+---------------+---------------+---------------+---------------+---------------+--------
 BentleyStatus PropertyNameExp::ResolveColumnRef(ECSqlParseContext& ctx)
     {
+    if (m_propertyPath.Size() == 3)
```
rschili, line 277: leftover debug code, remove before merge

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
```diff
@@ -381,6 +388,15 @@ BentleyStatus PropertyNameExp::ResolveColumnRef(ECSqlParseContext& ctx)
      }
 
     if (matchProps.empty()) {
+        // if it's RelECClassId argument of NAV function then assume that it's a static variable
+        if (GetParent()->GetType() == Exp::Type::NavValueCreationFunc && GetParent()->GetAs<NavValueCreationFuncExp>().GetRelECClassIdExp() == this)
```
rschili, line 392: I do not think we are making sure property path has as many elements as we require here.

iModelCore/ECDb/ECDb/BuiltInFuncs.h
```diff
@@ -17,6 +17,7 @@ BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
 #define SQLFUNC_ClassName "ec_classname"
 #define SQLFUNC_ClassId "ec_classid"
 #define SQLFUNC_InstanceOf "ec_instanceof"
+#define SQLFUNC_XmlCAToJson "XmlCAToJson"
```
rschili, line 20: Affan suggests we rename all the ToSomtehing methods into underscore ones and keep aliases for the pre existing ones so it doesn't break. New name would be like "guid_str_to_binary"

iModelCore/ECDb/ECDb/ClassViews.h
```diff
@@ -0,0 +1,26 @@
...
+#include "ECDbInternalTypes.h"
+#include "ClassMap.h"
+#include <functional>
+
+BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
+//======================================================================================
+// @bsiclass
+//======================================================================================
+struct ClassViews final {
+    constexpr static char ViewSchemaName[] = "ECDbMap";
```
rschili, line 15: constexpr char[] seems inconsistent, @khanaffan can we change this code to Utf8CP instead?

iModelCore/ECDb/ECDb/ECSql/ECSqlFieldFactory.cpp
simasjar: We don't need this piece of code, removed

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
simasjar: Refactored this place a bit

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
simasjar: We don't need this piece of code, removed

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
simasjar: removed

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
```diff
@@ -0,0 +1,1553 @@
...
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::InvalidECSql, stmt.Prepare(m_ecdb, "update ts.SchemaClassesView set schemaname = ?"));
+    }
+}
+
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+TEST_F(ClassViewsFixture, linktable_relationship_view) {
+    NativeLogging::Logging::SetLogger(&NativeLogging::ConsoleLogger::GetLogger());
```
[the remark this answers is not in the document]
rschili: NativeLogging needs to be removed before pushing the changes

iModelCore/ECDb/ECDb/ViewGenerator.cpp
```diff
@@ -871,6 +870,44 @@ BentleyStatus ViewGenerator::RenderRelationshipClassEndTableMap(NativeSqlBuilder
             sqlBuilder.Append(" AS INTEGER)");
         };
 
+    if (ClassViews::IsViewClass(*relationMap.GetRelationshipClass().GetTarget().GetAbstractConstraint()))
```
rschili, line 873: @khanaffan this is another section to take a closer look. It makes it so the relationship classes in ECDbMeta do not need a view query. Instead this goes and generates a view query dynamically by searching for a navigation property. It seems to work fine and I like this mechanism a lot. But maybe you see some flaw with the approach.

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
```diff
@@ -98,7 +98,7 @@ void ECDbMetaSchemaECSqlTestFixture::AssertSchemaDef(ECSchemaCR expectedSchema,
             ASSERT_EQ(m_ecdb.Schemas().GetClass("ECDbMeta", "ECSchemaDef")->GetId(), val.GetId<ECClassId>()) << "ECSchemaDef.ECClassId";
             ASSERT_STREQ("ClassId", colInfo.GetProperty()->GetAsPrimitiveProperty()->GetExtendedTypeName().c_str()) << "ECSchemaDef.ECClassId";
             }
-        else if (colName.EqualsI("Name"))
+        else if (colName.EqualsI("Name") || colName.EqualsI("Name_1"))
```
rschili, line 101: @simasjar can you elaborate on the scenario where the colName turns into Name_1 here?
How does it happen since no views are involved?

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
```diff
@@ -2000,4 +2000,576 @@ TEST_F(ECDbMetaSchemaECSqlTestFixture, PropertyOverrides)
...
+
+    {
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, "SELECT * FROM meta.CustomAttribute"));
+        //printf("%s\n", stmt.GetNativeSql());
+        ASSERT_EQ(stmt.Step(), BE_SQLITE_ROW);
+    }
+
+    /*{
+        ECSqlStatement stmt;
```
rschili, line 2032: Before we push we should enable this and make sure it works.

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
```diff
@@ -2000,4 +2000,576 @@ TEST_F(ECDbMetaSchemaECSqlTestFixture, PropertyOverrides)
...
+            }
+        //ASSERT_EQ(stmt.GetValueInt(0), 10);
+    }*/
+
+    {
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, "SELECT c.Name, XmlCAToJson(ca.Class.Id, ca.Instance) FROM meta.ECClassDef c JOIN meta.CustomAttribute ca ON ca.Class.Id=c.ECInstanceId"));
+        while (BE_SQLITE_ROW == stmt.Step())
+            {
+            printf("%s -> %s\n", stmt.GetValueText(0),stmt.GetValueText(1));
```
rschili, line 2046: Note to myself: needs to actually test instead of doing printf

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
```diff
@@ -2000,4 +2000,576 @@ TEST_F(ECDbMetaSchemaECSqlTestFixture, PropertyOverrides)
...
+                JOIN meta.ECPropertyDef pDef ON ca.Property.Id = pDef.ECInstanceId
+                JOIN meta.ECClassDef cDef ON ca.CustomAttributeClass.Id = cDef.ECInstanceId
+             WHERE cDef.Name='CAClass'
+            )stmt"));
+        ASSERT_EQ(stmt.Step(), BE_SQLITE_ROW);
+        ASSERT_STREQ("Bar", stmt.GetValueText(1));
+    }
+
+   {
+        // ECSqlStatement stmt;
```
rschili, line 2129: Try and enable and make it work

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
```diff
@@ -2000,4 +2000,576 @@ TEST_F(ECDbMetaSchemaECSqlTestFixture, PropertyOverrides)
...
+                JOIN meta.ECSchemaDef s ON ca.Schema.Id = s.ECInstanceId
+                JOIN meta.ECClassDef cDef ON ca.CustomAttributeClass.Id = cDef.ECInstanceId
+             WHERE cDef.Name='CAClass'
+            )stmt"));
+        ASSERT_EQ(stmt.Step(), BE_SQLITE_ROW);
+        ASSERT_STREQ("TestSchema", stmt.GetValueText(1));
+    }
+
+   {
+        // ECSqlStatement stmt;
```
rschili, line 2247: Here as well, make it work before merge

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
simasjar: It doesn't turn into Name_1 with the current implementation, removed those checks

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
simasjar: uncommented

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
simasjar: uncommented

iModelCore/ECDb/Tests/NonPublished/ECDbMetaSchemaECSqlTestFixture.cpp
simasjar: uncommented

iModelCore/ECDb/ECDb/ECSql/ValueCreationFuncExp.h
```diff
@@ -0,0 +1,75 @@
...
+struct NavValueCreationFuncExp final : ValueCreationFuncExp
+    {
+    private:
+        size_t m_idArgExpIndex = UNSET_CHILDINDEX;
+        size_t m_relECClassIdArgExp = UNSET_CHILDINDEX;
+        FinalizeParseStatus _FinalizeParsing(ECSqlParseContext&, FinalizeParseMode) override { return FinalizeParseStatus::Completed; }
+        bool _TryDetermineParameterExpType(ECSqlParseContext& ctx, ParameterExp& parameterExp) const override { parameterExp.SetTypeInfo(ECSqlTypeInfo::CreatePrimitive(ECN::PRIMITIVETYPE_Long)); return true; }
+        void _ToECSql(ECSqlRenderContext&) const override;
+        void _ToJson(BeJsValue, JsonFormat const&) const {}
+        Utf8String _ToString() const override { return "NavValueCreationFuncExp"; }
```
rschili, line 54: @simasjar This _ToJson was missing the override keyword. It still needs an implementation plus a test.

iModelCore/ECDb/ECDb/ECSql/ValueCreationFuncExp.h
```diff
@@ -0,0 +1,75 @@
...
+//! @bsiclass
+//+===============+===============+===============+===============+===============+======
+struct ValueCreationFuncExp : ValueExp
+    {
+    private:
+        size_t m_columnRefExpIndex = UNSET_CHILDINDEX;
+        size_t m_classNameExpIndex = UNSET_CHILDINDEX;
+
+    public:
+        ValueCreationFuncExp(std::unique_ptr<DerivedPropertyExp> columnRefExp, std::unique_ptr<ClassRefExp> classRefExp, Type type) : ValueExp(type)
```
rschili, line 22: I believe this needs to go protected because we don't want to directly create ValueCreationFuncExp, no? It also does not implement all virtual members, so it would not make sense to create an object of this class. I'm surprised the compiler does not complain about this.

simasjar: changed ctor access modifier to protected

iModelCore/ECDb/ECDb/ECSql/ValueCreationFuncExp.h
simasjar: implemented

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+        <ECCustomAttributes>
+            <ClassMap xmlns="ECDbMap.02.00.00">
+                <MapStrategy>ExistingTable</MapStrategy>
+                <TableName>ec_CustomAttribute</TableName>
+            </ClassMap>
+        </ECCustomAttributes>
+        <ECProperty propertyName="ContainerId"  typeName="int"/>
+        <ECProperty propertyName="ContainerType"  typeName="int"/>
+        <ECProperty propertyName="Ordinal"  typeName="int"/>
+        <ECProperty propertyName="Instance"  typeName="string"/>
```
ColinKerr, line 549: should have an extended type identifying this as ECInstanceXml or at least Xml

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+        <ECEnumerator name="Point2d" value="1793" displayLabel="Point2d" />
+        <ECEnumerator name="Point3d" value="2049" displayLabel="Point3d" />
+        <ECEnumerator name="String" value="2305" displayLabel="String" />
+        <ECEnumerator name="IGeometry" value="2561" displayLabel="IGeometry" />
+    </ECEnumeration>
+
+    <!-- Flat list of Custom Attributes without relationships to their container -->
+    <ECEntityClass typeName="CustomAttribute" description="" displayLabel="" modifier="Sealed">
+        <ECCustomAttributes>
+            <ClassMap xmlns="ECDbMap.02.00.00">
```
ColinKerr, line 541: use consistent version for referenced schema

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+                        XmlCAToJson([ca].[Class].[Id], [ca].[Instance]) [ValueJson]
+                    FROM [meta].[CustomAttribute] [ca]
+                    WHERE [ca].[ContainerType] = 992
+                    ORDER BY [ca].[Ordinal]
+                </Query>
+            </View>
+        </ECCustomAttributes>
+        <ECNavigationProperty propertyName="Property" relationshipName="PropertyHasCustomAttribute" direction="backward"/>
+        <ECNavigationProperty propertyName="CustomAttributeClass" relationshipName="CustomAttributeClassHasInstanceOnProperty" direction="backward"/>
+        <ECProperty propertyName="ValueJson"  typeName="string"/>
```
ColinKerr, line 579: Should have extended type name of json, maybe then the property name can loose the Json postfix.

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+    <ECRelationshipClass typeName="CustomAttributeClassHasInstanceOnClass" modifier="Abstract" strength="referencing">
+        <Source multiplicity="(1..1)" roleLabel="has instance on" polymorphic="false">
+            <Class class="ECClassDef"/>
+        </Source>
+        <Target multiplicity="(0..*)" roleLabel="is instance of" polymorphic="false">
+            <Class class="ClassCustomAttribute"/>
+        </Target>
+    </ECRelationshipClass>
+
+       <ECEntityClass typeName="SchemaCustomAttribute" description="" displayLabel="" modifier="Abstract">
```
ColinKerr, line 634: incorrect indentation

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+        </Target>
+    </ECRelationshipClass>
+    <ECRelationshipClass typeName="CustomAttributeClassHasInstanceOnProperty" modifier="Abstract" strength="referencing">
+        <Source multiplicity="(1..1)" roleLabel="has instance on" polymorphic="false">
+            <Class class="ECClassDef"/>
+        </Source>
+        <Target multiplicity="(0..*)" roleLabel="is instance of" polymorphic="false">
+            <Class class="PropertyCustomAttribute"/>
+        </Target>
+    </ECRelationshipClass>
```
ColinKerr, line 596: It's weird that these relationships are abstract but there is no other indication that they are 'virtual' or 'view' relationships.  The view classes at least have a view CA so it is more obvious

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+
+    <!-- Flat list of Custom Attributes without relationships to their container -->
+    <ECEntityClass typeName="CustomAttribute" description="" displayLabel="" modifier="Sealed">
+        <ECCustomAttributes>
+            <ClassMap xmlns="ECDbMap.02.00.00">
+                <MapStrategy>ExistingTable</MapStrategy>
+                <TableName>ec_CustomAttribute</TableName>
+            </ClassMap>
+        </ECCustomAttributes>
+        <ECProperty propertyName="ContainerId"  typeName="int"/>
```
ColinKerr, line 546: If this is the Id of the Class, property or whatever shouldn't this be a long instead of an int?

We are kind of cheating here by making this a number instead of an explicit id.  So we should probably use an extended type here to give at least some hint.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
```diff
@@ -0,0 +1,1551 @@
...
+        }
+        ASSERT_EQ(BE_SQLITE_DONE, stmt.Step(instKey));
+        m_ecdb.SaveChanges();
+    }
+   //! HARD_CODED_IDS
+   // do not indent the following string as it increases the size of string and compiler will error out.
+    BeJsDocument expected;
+    expected.Parse(R"json([
+  [
+    "0x1",
```
ColinKerr, line 873: Should this expected json be stored in its own file?

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -4070,6 +4170,91 @@ BentleyStatus ECSqlParser::ParseWindowFrameExclusion(WindowFrameClauseExp::Windo
...
+    std::unique_ptr<PropertyNameExp> propNameExp = std::make_unique<PropertyNameExp>(
+        *m_context,
+        propPath[2].GetName(),
+        *classNameExp.get(),
+        *classMap
+    );
+
+    propNameExp->SetTypeInfo(ECSqlTypeInfo(ECSqlTypeInfo::Kind::Navigation));
+    Utf8CP alias = parseNode->getParent()->getChild(1)->getTokenValue().empty() ? nullptr : parseNode->getParent()->getChild(1)->getTokenValue().c_str();
+    derivedPropertyExp = std::move(std::make_unique<DerivedPropertyExp>(std::move(propNameExp), alias));
```
rschili, line 4243: Compiler error from LLVM (Linux): mmoving a temporary object prevents copy elision

iModelCore/ECDb/ECDb/ClassViews.cpp
```diff
@@ -0,0 +1,220 @@
...
+                IssueType::ECDbIssue, ECDbIssueId::ECDb_0716,"Invalid view class '%s'. View class does not have a classmap", viewClassName);
+        return false;
+    }
+    if (viewClass.IsRelationshipClass()) {
+        auto relationshipClass = viewClass.GetRelationshipClassCP();
+        if (!IsViewClass(*relationshipClass->GetTarget().GetAbstractConstraint()) || !IsViewClass(*relationshipClass->GetSource().GetAbstractConstraint())) {
+            conn.GetImpl().Issues().ReportV(
+                IssueSeverity::Error,
+                IssueCategory::BusinessProperties,
+                IssueType::ECDbIssue, ECDbIssueId::ECDb_0722,"Invalid view class '%s'. If view is applied to the ECRelationshipClass, then source and target classes must be view classes as well.", viewClassName);
```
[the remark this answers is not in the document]
rschili: We should discuss this. I know Affan explained a bit ago why the source and target have to be views as well, but I do not remember all the details. I agree what you describe would be a powerful feature.

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: Hm yeah. Agreed. I'm not sure what to do about it. It is kind of a shortcut hack that Simonas came up with, because they can get all their information by looking for a navigation property. So giving them a view definition would be redundant.

Should we add a new custom attribute for these types of relationships? Like <DynamicallyCreatedViewBasedOnForeignKey>? I know it's a silly name, we need a better one, it's just to get the idea across...

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
khanaffan: > std::move(std::make_unique<
That is wrong. Get rid of std::move()

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
khanaffan: yes, we should use `extendedType="Id"` otherwise serializer fails to convert it into hex.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
khanaffan: Whenever possible test should be fully atomic with no external dependency.  (_This is something I & Krischan found out when we wasted a lot of time on fixing the test and realized the test should be atomic and self-contained with little as possible external dependency_). This strategy prevents the test from breaking for external reasons.  It also makes it easy to update, move, and debug tests. Some take away from our experience was
1. Data for the test should be part of the test. In our case Xml schemas and output data.
2. Assert should be in the test body. If you have an array of data you want to assert use a single assert for each case. Instead of using a loop to assert. This is to avoid figuring out which loop iteration is causing assert failure.
3. Do not assert an external re-usable function. This makes it harder to figure out where the test is failing.
4. The test should be flat with no function calls to reusable test functions with asserts or loops or external data loaded from files.
5. If the test is self-contained and when it's deleted it will have no dependency floating around that needs to be deleted as well.

Pasting those large json in the test might make it less readable but it still makes the test have zero dependencies.  Imagine when I change code in ECDb and this test fails and all you see is a test.json file getting updated in PR with no context which test it fixed or was failing. Since PR will not connect the data file to the test.

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: done

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: done

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: done

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
```diff
@@ -0,0 +1,669 @@
...
+        <ECProperty propertyName="ContainerType"  typeName="int"/>
```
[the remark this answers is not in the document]
rschili: good idea, done

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: fixed

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: done

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
rschili: closing comment not changing this json for now

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
rschili: done

rschili (approved):

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
rschili: @ColinKerr we have tests for all of the above except for this type:
- Metadata mismatch like different persistence units
What did you have in mind? The view query does not need to have a KindOfQuantity, so this will always use the one specified on the property. Is that a concern?

iModelCore/ECDb/ECDb/ECSchemas/ECDbMeta.04.00.02.ecschema.xml
rschili: Added a CA to indicate the behavior

iModelCore/ECDb/ECDb/ClassViews.cpp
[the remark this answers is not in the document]
rschili: Discussion done, we can do it in a future release.

iModelCore/ECDb/Tests/NonPublished/ClassViewsTests.cpp
rschili: According to discussion this limitation will be mentioned in the final documentation in the itwinjs-core repo

iModelCore/ECDb/ECDb/ClassViews.cpp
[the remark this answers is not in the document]
ColinKerr: I want to understand this now.  If we leave this in place then this version of ECDb will not be able to open files where a view relationship has two concrete endpoints.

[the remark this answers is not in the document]
rschili: I will put a few tests for the scenario and think about it, so we can make the current release behave in a way that is future proof.

rschili (approved):

iModelCore/ECDb/ECDb/ClassViews.cpp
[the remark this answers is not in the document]
rschili: In our call yesterday we discussed that we will limit the View CA to Entity classes for this release and introduce a new CA for relationships if we start properly supporting them in the future. Change has been made accordingly.

PR #342, opened 2023-07-03
Configurations to Consume Curl Component in iTwinNativeThirdParty

naveedkhan8067:
This upgrade allow us to consume relevant configurations of Curl component in iTwinNativeThirdParty. The reason of doing this upgrade is that now we want to consume the duplicate/similar components from a central location.

As the component coming from iTwinNativeThirdParty is **currently** consuming ares version `1.18.1` that is why the makefile there do not have `$(o)ares_rand$(oext) : $(AresLibDir)ares_rand.c ${MultiCompileDepends}` but with this upgrade the component generated there will be on ares version `1.19.1` and need this statement which is already present on central location.

### DIFF
--- a/iModelCore/libsrc/curl/BeCurl.mke
+++ b/iModelCore/libsrc/curl/BeCurl.mke
@@ -37,8 +37,14 @@ AresLibDir = $(AresDir)src/lib/
 BeCurlLibDir        = $(SourceDir)lib/
 BeCurlDepends       = $(_MakeFileSpec) $(_MakeFilePath)BeCurlConfig.h
 MultiCompileDepends = $(_MakeFileSpec) $(BeCurlDepends)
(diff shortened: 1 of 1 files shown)


nick4598 (approved):

aruniverse (approved):
deferring to @chuckkir to approve; please wait on his approval before merging

naveedkhan8067:
> deferring to @chuckkir to approve; please wait on his approval before merging

sure

aruniverse:
> > deferring to @chuckkir to approve; please wait on his approval before merging
> 
> sure but @chuckkir would be out probably until 26th and there are a couple of PR's in iTNTP that would need this change as a prerequisite.

just noticed hes out on vacation; lets go ahead and merge then

PR #361, opened 2023-07-19
Presentation: Avoid determining children for parent node

grigasp:
We just got the parent of specific child node, so we know that the parent has children - no need to determine that.


grigasp:

saskliutas (approved):

PR #364, opened 2023-07-20
Presentation: Unique values in instance filter should only be retrieved from instances that are child nodes for filtered parent node

medikornov:
Closes: https://github.com/iTwin/itwinjs-core/issues/5725
itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/presentation/unique-values-based-on-parent-node
itwinjs-core PR: https://github.com/iTwin/itwinjs-core/pull/5763


iModelCore/ECPresentation/PublicAPI/ECPresentation/Content.h
```diff
@@ -1113,6 +1113,7 @@ struct EXPORT_VTABLE_ATTRIBUTE ContentDescriptor : RefCountedBase
 
 private:
     PresentationRuleSetCPtr m_ruleset;
+    PresentationRuleSetPtr m_hierarchyLevelRuleset;
```
grigasp, line 1116: is this ruleset different from `m_ruleset`?

iModelCore/ECPresentation/Source/PresentationManagerImpl.cpp
```diff
@@ -1198,7 +1198,9 @@ ContentDescriptorCPtr RulesDrivenECPresentationManagerImpl::_GetNodesDescriptor(
             ),
         *KeySet::Create(params.GetParentNodeKey() ? NavNodeKeyList{ params.GetParentNodeKey() } : NavNodeKeyList{})
         ), params);
-    return GetContentDescriptor(descriptorParams);
+    auto descriptor = GetContentDescriptor(descriptorParams);
+    (const_cast<ContentDescriptor*>(descriptor.get()))->SetHierarchyLevelRuleset(*ruleset);
```
grigasp, line 1202: This is not good. The descriptor we get from `GetContentDescriptor` possibly comes from cache, so we definitely don't want to mutate it. Instead, we could clone it using `ContentDescriptor::Create(*sourceDescriptor)` and then modify.

grigasp (changes requested):

iModelCore/ECPresentation/Source/PresentationManagerImpl.cpp
medikornov: fixed

iModelCore/ECPresentation/PublicAPI/ECPresentation/Content.h
```diff
@@ -1285,6 +1288,11 @@ struct EXPORT_VTABLE_ATTRIBUTE ContentDescriptor : RefCountedBase
     //! Should only distinct values be returned
     bool OnlyDistinctValues() const {return HasContentFlag(ContentFlags::DistinctValues);}
 #endif
+
+    //! Set if this descriptor is created using a different from the input ruleset
+    void SetDifferentFromInputRuleset(bool value) {m_differentFromInputRuleset = value;}
+    //! Is the descriptor created using a different from the input ruleset
+    bool IsDifferentFromInputRuleset() const {return m_differentFromInputRuleset;}
```
grigasp, line 1295: maybe just set this to `true` in `SetRuleset`?

iModelCore/ECPresentation/Source/PresentationManagerImpl.cpp
```diff
@@ -1198,7 +1198,13 @@ ContentDescriptorCPtr RulesDrivenECPresentationManagerImpl::_GetNodesDescriptor(
             ),
         *KeySet::Create(params.GetParentNodeKey() ? NavNodeKeyList{ params.GetParentNodeKey() } : NavNodeKeyList{})
         ), params);
-    return GetContentDescriptor(descriptorParams);
+    auto descriptor = ContentDescriptor::Create(*GetContentDescriptor(descriptorParams));
+    if (descriptor.IsValid())
```
grigasp, line 1202: `ContentDescriptor::Create` will always return a descriptor, but `GetContentDescriptor(descriptorParams)` is not guaranteed to. This should be changed to:
```cpp
auto descriptor = GetContentDescriptor(descriptorParams);
if (descriptor.IsNull())
    return nullptr;

auto result = ContentDescriptor::Create(*descriptor);
...
```

grigasp (changes requested):

iModelCore/ECPresentation/PublicAPI/ECPresentation/Content.h
grigasp: Also, `IsDifferentFromInputRuleset` sounds like we're comparing descriptor with a ruleset. Maybe rename to `UsesModifiedRuleset`?

medikornov: fixed

iModelCore/ECPresentation/Source/PresentationManagerImpl.cpp
medikornov: fixed

grigasp (approved):

PR #369, opened 2023-07-25
Add ability to enable experimental syntax in concurrent query

RohitPtnkr1996:
I've repurposed the boolean parameter in ECSqlConfig that is being used by non-concurrent pragma ECSqlStatements to enable experimental features.
The function ECSqlRequest::MakeRequest is updated to accept an optional boolean parameter "enableExperimentalFeatures" which defaults to false.

When running concurrent ECSql queries, the function ECSqlRequest::MakeRequest() should be called with the parameter "enableExperimentalFeatures" set to true for the query that is using experimental features.
The query request which has this parameter absent or set to false, will have experimental features disabled for it and will give the appropriate error message if an experimental feature is used in it.

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/enable-experimental-feature-concurrent-queries


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -57,12 +57,28 @@ int QueryRetryHandler::_OnBusy(int count) const {
 //---------------------------------------------------------------------------------------
 // @bsimethod
 //---------------------------------------------------------------------------------------
-std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool usePrimaryConn, bool suppressLogError, ECSqlStatus& status, std::string& ecsql_error) {
+std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool usePrimaryConn, bool suppressLogError, ECSqlStatus& status, std::string& ecsql_error, const bool ignoreCache) {
```
rschili, line 60: The ignoreCache flag needs explaining. It seems to default to false, but turns true if there are no experimental features enabled for the request.
Despite what the flag says, we still try and go to the cache, pick the entry, move it to the front, but then go and prepare the statement again.
I guess "reprepareCachedStatement" would be a better name for this parameter.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -71,22 +87,23 @@ std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool
         }
         entry->GetStatement().Reset();
         entry->GetStatement().ClearBindings();
+
+        if (ignoreCache)
+            {
+            // Experimental Features are disabled, so finalize and re-prepare the statement to handle error
```
rschili, line 93: This comment is more misleading than useful. IgnoreCache is in the current code only true, if experimental features are disabled. But that doesn't mean we should conclude experimental features are disabled because the flag is true.
Can you explain here, why we need to reprepare the statement if experimental features are not enabled?

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -1171,28 +1188,47 @@ void QueryHelper::Execute(QueryAdaptorCache& adaptorCache, RunnableRequestBase&
...
         ECSqlStatus status;
         std::string err;
-        auto adaptor = adaptorCache.TryGet(sql.c_str(), request.UsePrimaryConnection(), request.GetSuppressLogErrors(), status, err);
+
+        // Cache experimental features status to be reset after concurrent query is done
+        const auto experimentalFeaturesStatusCached = adaptorCache.GetConnection().GetPrimaryDb().GetECSqlConfig().GetExperimentalFeaturesEnabled();
+        auto updateExperimentalFeatures = [&adaptorCache] (const bool experimentalFeaturesStatus) { adaptorCache.GetConnection().GetPrimaryDb().GetECSqlConfig().SetExperimentalFeaturesEnabled(experimentalFeaturesStatus); };
+
+        // Enable/Disable experimental features just for this query
+        const auto experimentalFeaturesInRequest = runnableRequest.AreExperimentalFeaturesEnabledForRequest();
```
rschili, line 1197: This is a bit nit-picky, but I believe this would be cleaner if you called the function restorePreviousExperimentalFlag() and the variable previousExperimentalFeaturesFlag.
Better name experimentalFeaturesInRequest into experimentalFeaturesEnabledForRequest (the other name may suggest they are actually being used instead of being enabled)

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
RohitPtnkr1996: The need for this re-prepare arises from the following example scenario:
1. An instance query is executed with experimental features enabled and it executes successfully.
2. The same query is called again but this time with experimental features disabled.
In  this scenario, the cached adaptor will be picked up and the already prepared statement will be executed successfully.
We expect a failure in this scenario. Hence the re-prepare, since that's where the experimental features are checked.

Fair point though, the comment and arg name is a bit misleading.

RohitPtnkr1996: Also, the function TryGet() is only called form one place. So the ignoreCache arg being passed is based on the experimental features status. I can change the arg name to be a bit more specific about experimental features and not so generic as ignoreCache.

rschili: Ah now I get it. And the opposite case does not need handling because if the request failed to prepare, it would not have been added to the cache, so we don't have to bother if they were turned on, only if they were turned off.
Please document this in the code, it's not obvious :-)

rschili (approved):

khanaffan:
@RohitPtnkr1996 Can you wait on this PR? Do not merge it yet.

PR #371, opened 2023-07-25
Fix for parsing JOIN-USING with relationship with alias

ArturasDeltuvas:
[Unable to parse alias JOIN-USING with relationship with alias #815](https://github.com/iTwin/itwinjs-backlog/issues/815)


iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -1899,6 +1899,47 @@ BentleyStatus ECSqlParser::ParseTableValuedFunction(std::unique_ptr<TableValuedF
...
+        {
+        BeAssert(false && "Wrong grammar. Expecting table_node_ref");
+        return ERROR;
+        }
+
+    std::unique_ptr<ClassNameExp> table_node = nullptr;
+    if (SUCCESS != ParseTableNodeWithOptMemberCall(table_node, *tableNode.getChild(0/*table_node_with_opt_member_call*/), ecsqlType, PolymorphicInfo::All(), false))
+        return ERROR;
+
+    std::unique_ptr<RangeClassRefExp> rangeClassRef;
```
rschili, line 1918: I think you can merge those two lines with a make_unique<> call.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -1899,6 +1899,47 @@ BentleyStatus ECSqlParser::ParseTableValuedFunction(std::unique_ptr<TableValuedF
     exp = std::make_unique<TableValuedFunctionExp>(schemaName.c_str(), std::move(memberFuncCall), PolymorphicInfo::All());
     return SUCCESS;
 }
+
+//-----------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+BentleyStatus ECSqlParser::ParseTableNodeRef(std::unique_ptr<ClassRefExp>& exp, OSQLParseNode const& tableNode, ECSqlType ecsqlType) const
```
rschili, line 1906: Since ECSqlParser is very big, I do not like this method name. It's so vague. That is a tableNodeRef?
But I see the other method is already called ParseTableNode, mmh....
If this turns out to be only used in relationships, maybe at least call it ParseRelationshipJoinTableNode, that would add a little context

rschili: Hm closing this comment. After reviewing the other method names I don't want to break the existing scheme.

ArturasDeltuvas: I wasn't sure what to call it and as ParseTableNode(...) made a prt<ClassNameExp> and I needed a ClassRefExp  to store the table node alias, I called the function ParseTableNodeRef(...)

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -10790,4 +10790,185 @@ TEST_F(ECSqlStatementTestFixture, verify_inf_and_nan_handling) {
...
+        ECSqlStatement stmt;
+        auto statementString = "SELECT 1 FROM bis.Element a JOIN bis.Element b USING bis.ElementOwnsChildElements c";
+        ASSERT_EQ(ECSqlStatus::InvalidECSql, stmt.Prepare(m_ecdb, statementString)) << "FORWARD or BACKWARD must be specified for joins where source and target cannot be identified unambiguously, e.g. joins between the same class.";
+        }
+
+    if("without alies and FORWARD")
+        {
+        ECSqlStatement stmt;
+        auto statementString = "SELECT 1 FROM bis.Element a JOIN bis.Element b USING bis.ElementOwnsChildElements FORWARD";
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, statementString));
```
rschili, line 10842: I know this is tedious, but just checking if the query prepares successfully does not mean that it works.
We should probably have a test which checks the native SQL after prepare, OR one which runs the query and checks that the results are correct.
Also, in the context of the test statements, the alias does not really do anything.
I'd like to see a statement which refers or selects something from "c". You can add properties to "ElementOwnsChildElements" and then try to select those, if that even makes sense..... the using statement is not a JOIN or FROM clause, so I'm not sure in which scope that c alias can be used.
Find out what you can do with c and add a test which picks it up for something.

rschili (approved):

PR #373, opened 2023-07-25
Revert "Let clients of the daemon work with readonly perms to cloudca…

nick4598:
…che. (#211)"

This reverts commit 422d88f3144c2d64f4364238b60a530a262465d4.

After one of our deployed backends was experiencing a performance degradation when using the daemon and it disappearing in daemonless mode, I am suspicious of this change that lets clients of the daemon work with readonly permissions to the cloudcache. Part of the reason that I am suspicious is that there is a flood of daemon logs which have the following pattern:
12->D: READ(0, []) 
D->12: READ_REPLY(0, "", [array of blocks in the cachefile])
12->D: END([])
The above pattern repeats itself often.

The performance degradation is seen when typing into the search bar on the model tree. Even though the hierarchy cache appears to be built up, the second search attempt is almost as slow as the first as if the cache is not being used. Without daemon mode on, the second search attempt is extremely fast. I don't quite understand how these two things would affect eachother but the presence of the above pattern in daemon logs when sending more requests is what makes me suspicious.

### DIFF
--- a/iModelCore/BeSQLite/SQLite/bcv_int.h
+++ b/iModelCore/BeSQLite/SQLite/bcv_int.h
@@ -19,7 +19,6 @@ typedef sqlite3_int64 i64;
 typedef sqlite3_uint64 u64;
 typedef unsigned char u8;
 typedef unsigned int u32;
-typedef unsigned short u16;
 
 #ifdef __WIN32__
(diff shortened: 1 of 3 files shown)


nick4598:

PR #378, opened 2023-07-31
Presentation: Use request-level read transactions

grigasp:
I was pretty sure we have them, but turns out we don't...

Provides ~6-7% performance boost for requests that run large numbers of queries (Models tree initial load & filtering, consolidated content descriptor).


saskliutas (approved):

nick4598:

PR #386, opened 2023-08-01
Abbreviated blobs as a Json string of their size

mattbjordan:
Instead of a shortened encoded string (e.g., `"encoding=base64;blah"`), blobs are a json string containing their size e.g., `"{\"bytes\":123}"`. Formatting it like this also allows for adding in more blob info to the json later if necessary.

Fixes https://github.com/iTwin/itwinjs-backlog/issues/702

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/mattbjordan/update-abbreviated-blobs-tests


iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
```diff
@@ -1,2332 +1,2332 @@
...
-
-
-        // BeJsDocument actual;
-        // actual.Parse(stmt.GetValueText(0));
-        // EXPECT_STRCASEEQ(expected.Stringify(StringifyFormat::Indented).c_str(), actual.Stringify(StringifyFormat::Indented).c_str());
-    }
-}
-
-END_ECDBUNITTESTS_NAMESPACE
+/*---------------------------------------------------------------------------------------------
```
rschili, line 1: Github marks this entire file as changed (12332 lines removed, same amount added).
Looks broken to me, have the line endings changed or something similar?

rschili (approved):
Please check the InstanceReaderTest.cpp file before merging. Looks like it has not changed at all but the entire file is marked as modified. Try revert.

khanaffan (approved):

Issue #391, opened 2023-08-03
Crash when creating a `Buffer` on Electron 21+


grigasp:
Starting from addon version `4.1.0` (https://github.com/iTwin/imodel-native/pull/283), we used `Napi::Buffer` to expose our large responses to javascript without needing to copy them.

As described in https://github.com/electron/electron/issues/35801 and https://github.com/microsoft/vscode/issues/177338, due to security reasons Electron 21+ stopped supporting creating buffers pointing to external data and now throws a runtime error when doing so.

PR #392, opened 2023-08-03
Presentation: Use `Buffer::NewOrCopy` instead of `Buffer::New`

grigasp:
Fixes https://github.com/iTwin/imodel-native/issues/391.

With Electron 21+ we can't use `Buffer::New` anymore. 
- Update `node-addon-api` to [`7.0.0`](https://github.com/nodejs/node-addon-api/releases/tag/v7.0.0) which adds a new API `Buffer::NewOrCopy`. 
- In our addon code, define the `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` macro to disable `Buffer::New` functions so uses of them are captured at compile time.
- Don't check for `NODE_API_NO_EXTERNAL_BUFFERS_ALLOWED` in `NewOrCopy` - we want to always prefer using external buffers and only fall back to copies when they're not supported.


aruniverse (approved):
looks like the breaking changes in node-addon-api should be ok for us;
https://github.com/nodejs/node-addon-api/compare/v6.1.0...v7.0.0#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR9

grigasp:
> Will this prevent calls to `Buffer::New` in other layers (e.g., iModelPlatform)?

I wasn't aware napi is used outside of addon code. Moved the define elsewhere - I believe now it will be defined for all napi users in our code.

pmconne (approved):

PR #397, opened 2023-08-07
fix text offset for SimplifyGraphic::ProcessText

MarcNeely:
The offset from the text itself is now taken into account along with the glyph offset.


pmconne (approved):

iModelCore/iModelPlatform/DgnCore/SimplifyGraphic.cpp
```diff
@@ -860,6 +861,7 @@ void SimplifyGraphic::_AddBody(IBRepEntityCR geom)
 +---------------+---------------+---------------+---------------+---------------+------*/
 void SimplifyGraphic::_AddTextString(TextStringCR text)
     {
+//*q*/printf("SimplifyGraphic::_AddTextString\n");
```
pmconne, line 864: Remove?

MarcNeely: oops, thought I had gotten all of those.

PR #398, opened 2023-08-07
Fix for ALL, ANY and SOME boolean expressions not working

ArturasDeltuvas:
[ALL, ANY or SOME boolean test expression does not work#799](https://github.com/iTwin/itwinjs-backlog/issues/799)


rschili (approved):
Good job going the extra mile implementing this in the preparer. Well done!

PR #401, opened 2023-08-09
Handle nested content similar to how we handle direct content

grigasp:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/846.

1. Related content may have a large number of nested related content fields, causing us to exceed max joins limit. So we have to handle related content similar to how we handle direct content - count related classes and query related content separately if the number of related classes is too large.

2. When counting related classes, we have to add one for every navigation property as they require joining the target class.


grigasp:

saskliutas (approved):

PR #411, opened 2023-08-10
Allow line break in ECSQL quoted names

khanaffan:
Fix an issue where quoted name did not allow line break. This allows `PRAGMA` value to accept multiline value.

We had special code to handle `[]` for array descriptor. It has been removed until we figure out how to handle array or collection in future.


rschili (approved):
Are we good removing that array support? I guess so since you went ahead and did it.... How likely is it that it will break somebody?

PR #421, opened 2023-08-21
Experimental features should be inherited by nested instance queries or props

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/856


iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
```diff
@@ -2329,4 +2329,53 @@ TEST_F(InstanceReaderFixture, nested_struct) {
     }
 }
 
+TEST_F(InstanceReaderFixture, NestedSelectShouldInheritExperimentalFeatures) {
+    ASSERT_EQ(BE_SQLITE_OK, OpenECDbTestDataFile("test.bim"));
+    ASSERT_FALSE(IsECSqlExperimentalFeaturesEnabled(m_ecdb));
+
+    for (const auto& [lineNumber, sqlStatement, sqlStatus] : std::vector<std::tuple<unsigned int, Utf8CP, ECSqlStatus>>
```
rschili, line 2336: All of these queries have their nested experimental features in the FROM clause. Do we have a test which uses experimental features in the WHERE clause?
Didn't actually try this query, but I imagine something like:
`SELECT * FROM Meta.ECClassDef WHERE Name IN (SELECT distinct $->Name from Meta.ECClassDef) ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES;`

RohitPtnkr1996: Done

rschili (approved):

RohitPtnkr1996:
Fixed in #446

PR #425, opened 2023-08-23
Support multi-model feature tables

pmconne:
- `Feature` now includes a model Id.
- When creating a `FeatureTable`, the caller specifies whether it is intended to hold features from one or multiple models. This controls how we pack the table.
- `FeatureTable` no longer has a model Id property.
- If the table only supports one model, attempting to add features from multiple models will produce a run-time exception.
- If the table supports multiple models, the features must be added in groups by model Id (ascending). This is to eliminate the need to otherwise remap all feature indices for each vertex after the table is fully populated.

ImageTests produced no differences (unrelated flaky TimingTest and Iceland map imagery aside).

imodel-native-internal: https://github.com/iTwin/imodel-native-internal/tree/pmc/multi-model-feature-table
[PR](https://github.com/iTwin/imodel-native-internal/pull/154)


pmconne:
Awaiting ImageTests results.

pmconne:
@DanRod1999 do you understand what's going on with [the builds](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2355478&view=logs&j=dd54af87-4668-5a7c-caf9-f383dd6d34a4&t=3390e708-64f4-5d3b-8551-70d9e26eae08&s=7f4420e9-0df3-5bdc-6ca4-f1e9b985fa65)?
Purge bbcache produces what looks like a warning (I've seen this several times recently - seems reasonable if the dir doesn't exist):
```
##[error]Container path not found: 'D:\vsts_b\213\bbsrc\bbcache'
```

Then build exits after 1 second:
```
Evaluating: and(succeeded(), eq(variables['getProfileVersion.EXECUTE_PV_TESTS'], 'True'))
Expanded: and(True, eq('False', 'True'))
Result: False
```

DanRod1999:
> @DanRod1999 do you understand what's going on with [the builds](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2355478&view=logs&j=dd54af87-4668-5a7c-caf9-f383dd6d34a4&t=3390e708-64f4-5d3b-8551-70d9e26eae08&s=7f4420e9-0df3-5bdc-6ca4-f1e9b985fa65)? Purge bbcache produces what looks like a warning (I've seen this several times recently - seems reasonable if the dir doesn't exist):

I don't know exactly what is happening here or what the true cause for this is, however I do know that this step is set to continue on error for a reason. Ive seen it fail multiple times and the builds continue and pass as expected. However, clearly this specific pipeline does not allow this. The solution is either fix the warning (I don't know how to do this but I don't think it's necessary) or set the Boolean that is failing to  and(SucceededwithErrors(), eq(variables['getProfileVersion.EXECUTE_PV_TESTS'], 'True')). You might need another 'or(succeeded()' inside that.

DanRod1999:
And( or(succeededWithErrors(), succeeded),eq(variables['getProfileVersion.EXECUTE_PV_TESTS'], 'True')). 

I think this is the correct condition that would allow the pipeline to continue even if the b cache warning appears

DanRod1999:
Wait sorry. I may have misunderstood the evaluation that you shared. I'll take a closer look

pmconne:
3rd attempt made it to Publish Profile Version Artifacts and died.
```
##[error]Artifact DgnDbArtifacts already exists for build 2355478.
```

imodel-native-internal build meanwhile has successfully built everything except iOS (still in progress).
I gather this profile version validation step is newish? If it's buggy, please consider turning it off until it's fixed to avoid blocking PRs.

DanRod1999:
I will contact who worked on this and let them know about the issues.

PR #426, opened 2023-08-23
Add support to enumerate local changes

khanaffan:
Added `getLocalChanges()` allows the enumeration of all local pending txns or optionally includes in-memory changes. 

* Users can specify a list of root classes to filter the changes polymorphically.
* The function only returns the instance key and changes type and does not return any more details.

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/affanK/get-local-changes1

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
@@ -467,6 +467,179 @@ void TxnManager::WriteChangesToFile(BeFileNameCR pathname, DdlChangesCR ddlChang
 }
 
 /**
+ * Iterate over local txn and notify changed instance keys and change type.
+ * rootClassFilter is optional but if provided iterator will only iterate over
+ * changes if it to one of root class or any of its derived class.
+ *
+ * Note: For large set of changes this function may consume a lot of none contigious memory.
+ */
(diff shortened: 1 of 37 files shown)


iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
```diff
@@ -466,6 +466,179 @@ void TxnManager::WriteChangesToFile(BeFileNameCR pathname, DdlChangesCR ddlChang
...
+            }
+        }
+    }
+
+    // Group all local txn
+    auto endTxnId = GetCurrentTxnId();
+    auto startTxnId = QueryNextTxnId(TxnId(0));
+    DdlChanges ddlChanges;
+    ChangeGroup dataChangeGroup;
+    for (auto currTxnId = startTxnId; currTxnId < endTxnId; currTxnId = QueryNextTxnId(currTxnId)) {
```
pmconne, line 508: `QueryNextTxnId` will return deleted (undone) txns...you probably want to filter those out.

pmconne: Nope, I misread. You use `QueryNextTxnId(0)` to get the *first* txn and iterate until >= curr txn. Deleted txns will always be > curr txn.

khanaffan: Correct, that's what I found out.

pmconne:
Will you submit a corresponding TypeScript PR to provide a beta API?

khanaffan:
> Will you submit a corresponding TypeScript PR to provide a beta API?

done

pmconne (approved):

PR #427, opened 2023-08-24
Adjust tests to avoid failure due to FP rounding

tcobbs-bentley:
Changes in code generation in Xcode 14.3 triggered unit test failures due to differences in floating point rounding. This adjusts the factor used to calculate the maximum error for two specific tests to avoid having them fail.


dassaf4 (approved):

tcobbs-bentley:
@ColinKerr @c-macdonald @calebmshafer @rschili This PR must be merged before @chuckkir can upgrade Xcode on the CI Macs, and that Xcode upgrade is necessary in order to be able to do sanity check tests on mobile. The only thing it is missing is approval (I just re-merged master into this PR). If there is a problem with the PR, please let me know. Otherwise, please approve.

calebmshafer (approved):
Approving but do you know what caused the change in rounding?

tcobbs-bentley:
The clang that comes with Xcode 14.3 has different code generation. I found [this](https://www.gaige.net/fp-concat-accuracy.html) article.

calebmshafer:
@tcobbs-bentley thanks for the article link!

PR #431, opened 2023-08-28
make writeLockExpires a getter

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/nick/exposebcvkvtime

I noticed that write lock expiry time no longer gets updated when you call acquireWriteLock in TS with the same user. I didn't want to test against the containerInternal object and preferred to have validation from the getter that I'm adding in which reads the actual bcv_kv table.

### DIFF
--- a/iModelJsNodeAddon/JsCloudSqlite.cpp
+++ b/iModelJsNodeAddon/JsCloudSqlite.cpp
@@ -438,7 +438,8 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
     Napi::Value GetWriteLockExpiryTime(NapiInfoCR) {
         RequireConnected();
         BeJsDocument lockedBy;
-        ReadWriteLock(lockedBy); // What if write lock is empty? 
+        ReadWriteLock(lockedBy); 
+        // Returns empty string if writeLock is empty.
         return Napi::String::New(Env(), lockedBy[JSON_NAME(expires)].asString());
     }
 
--- a/iModelJsNodeAddon/JsCloudSqlite.cpp
+++ b/iModelJsNodeAddon/JsCloudSqlite.cpp
@@ -435,6 +435,13 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
         });
     }
 
+    Napi::Value GetWriteLockExpiryTime(NapiInfoCR) {
+        RequireConnected();
+        BeJsDocument lockedBy;
+        ReadWriteLock(lockedBy); // What if write lock is empty? 
+        return Napi::String::New(Env(), lockedBy[JSON_NAME(expires)].asString());
+    }
+
     void ReadWriteLock(BeJsDocument& out) {
         Statement stmt;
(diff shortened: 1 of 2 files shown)


khanaffan (approved):

iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -435,6 +435,14 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
...
+        RequireConnected();
+        BeJsDocument lockedBy;
+        ReadWriteLock(lockedBy); 
+        // Returns empty string if writeLock is empty.
+        return Napi::String::New(Env(), lockedBy[JSON_NAME(expires)].asString());
+    }
+
     void ReadWriteLock(BeJsDocument& out) {
         Statement stmt;
         auto rc = stmt.Prepare(m_containerDb, "SELECT value FROM bcv_kv WHERE name=?");
```
khanaffan, line 448: Is there a reason you cannot do this from the typescript side? It will save from adding new method to addon. If its just a select query you can do it from typescript.

nick4598: I don't believe I can do this from typescript because  bcv_kv is a virtual table, and we don't let virtual tables be queried freely. The connection in question must call sqlite3_bcvfs_register_vtab() ( from the c++ side?). The only connection which currently registers these vtables is only available in the addon.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -1004,6 +1004,10 @@ export declare namespace IModelJsNative {
     public get alias(): string;
     /** The logId. */
     public get logId(): string;
+    /** The time that the write lock expires. Of the form 'YYYY-MM-DDTHH:MM:SS.000Z' in UTC.
+     *  Returns empty string if no write lock expiry time.
```
kabentley, line 1008:
```suggestion
     *  Returns empty string if write lock is not held.

```

nick4598: Technically if the write lock was never released (maybe due to some failed process), the write lock expiry time would not be an empty string but the write lock would also not be held (since expiry time is at some point in the past). Should this be reflected in the comment?

kabentley: The important point for the comment is that the caller must be aware that the string will be empty if the lock is not held. If the string is valid, it's a date string that tells you that the lock was successfully acquired and hasn't yet been released. Whether or not the lock has expired (or was "stolen" by an admin) isn't relevant for this call, I don't think. BTW, even if the lock expires things can still work fine unless/until someone else attempts to acquire the lock, notices it is expired, and  acquires it.

iModelJsNodeAddon/JsCloudSqlite.cpp
nick4598: @khanaffan okay to resolve this comment?

PR #433, opened 2023-09-01
CloudSQLite update: add bcv_stats table and nTrans to bcv_database, also expose to JS

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/nick/cachefilemetrics

This will be used in the daemon prometheus client.


iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -384,6 +384,53 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
...
+            if (result != BE_SQLITE_ROW) {
+                value[jsNames[i]] = -1; 
+            } else {
+                value[jsNames[i]] = stmt.GetValueInt(0); 
+            }
+            stmt.Reset();
+            stmt.ClearBindings();
+        }
+        stmt.Finalize();
+        // TODO: Would a where clause speed up the performance of this query?
```
nick4598, line 414: @khanaffan do you have an opinion on this TODO? 
The bcv_database virtual table will have one row for every database in an attached container, which seems like it could grow to a decent size, since there is one database for each checkpoint. I could add a where clause like 'where nclient >= 1' to only grab databases which have some open connection to them, but not sure if its worth doing.

khanaffan: I would not worry about it. Unless you get to a million files per container which seem unrealistic.

nick4598:
@DanRod1999 Do you know why it seems to be skipping the itwinjs-core stage? https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2375947&view=logs&j=038bd95d-6665-5615-4ef7-fcc16fc7c337

khanaffan (approved):

PR #434, opened 2023-09-05
Modify convertCustomAttributes to exclude referenced schemas

akshay05n:
**Summary:**
This pull request introduces changes to update the behaviour of the `convertCustomAttributes` API. Specifically, it modifies the API response to exclude referenced schemas that are not present in the input, aligning it with our new requirements.

**Details:**
Previously, the `convertCustomAttributes` API in the `SchemaUtility` class returned both input schemas and their references in the response in dependency order. With this update, the API will exclude referenced schemas from responses that are not in the input schema.


iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -990,14 +992,19 @@ BentleyStatus JsInterop::ConvertSchemas(bvector<Utf8String> const& inputStrings,
...
+                    SchemaWriteStatus writeStatus = schema->WriteToXmlString(outputStrings[i]);
+                    if (SchemaWriteStatus::Success != writeStatus)
+                        {
+                        outputStrings.clear();
+                        return BentleyStatus::ERROR;
+                        }
+                    break;
+                    }
                 }
-            outputStrings.push_back(schemaXml);
```
c-macdonald, line 1000: Don't you still need to populate the outputStrings vector somewhere?

akshay05n: Indeed, I'm resizing the `outputStrings` before iterating through the schemas vector. Following this, `schema->WriteToXmlString(outputStrings[i])` method is used to populate the `outputStrings` vector.

c-macdonald (approved):

rschili (approved):

khanaffan (approved):

PR #443, opened 2023-09-05
Ignore non-EC changes and avoid preparing classid for table with no classid column

khanaffan:
There were two issues reported and fixed in this PR.
1. Api throws an exception when non-EC data changes are detected.
2. Change to table with no `ECClassId` column threw an exception.

Add unit test to cover the above two cases.


pmconne (approved):

pmconne:
Please verify that the itwinjs-core build runs before merging - seems like it hasn't been doing so reliably.

nick4598:
Theres an open PR to fix the issue Paul mentions(itwinjs-core builds not running):  https://github.com/iTwin/imodel-native-internal/pull/159

PR #447, opened 2023-09-07
Add create and query count annotation text style APIs

bentleyttao:


kabentley:
I don't see why we should make this part of the native code at all. What is this for?

iModelJsNodeAddon/JsInteropDgnDb.cpp
```diff
@@ -1386,3 +1386,56 @@ void JsInterop::WriteAffectedElementDependencyGraphToFile(DgnDbR db, Utf8StringC
...
+        auto fontId = textStyle[json_fontId()].asUInt64();
+        auto height = textStyle[json_height()].asDouble();
+        auto lineSpacingFactor = textStyle[json_lineSpacingFactor()].asDouble();
+        auto isBold = textStyle[json_isBold()].asBool();
+        auto isItalic = textStyle[json_isItalic()].asBool();
+        auto isUnderlined = textStyle[json_isUnderlined()].asBool();
+        auto widthFactor = textStyle[json_widthFactor()].asDouble();
+        auto description = textStyle[json_description()].asCString();
+
+        auto ats = AnnotationTextStyle::Create(db.GetDictionaryModel());
```
jffmarker, line 1405: @abeesh, do you think we'll need the ability to insert definition elements like text styles into specific models (vs. the global dictionary)?

PR #450, opened 2023-09-12
Add static libs to upack for downstream consumers.

chuckkir:


aruniverse (approved):

iModelCore/iModelCore.PartFile.xml
```diff
@@ -64,6 +64,7 @@
         <ProductDirectory Name="VendorApi"              LibType="Static"    RelativeTo="Root" Path="include"/>
         <ProductDirectory Name="VendorNotices"                              RelativeTo="Root" Path="notices"/>
         <ProductDirectory Name="VendorNotices"          LibType="Static"    RelativeTo="Root" Path="notices"/>
+        <ProductDirectory Name="BeStaticLibDir"         LibType="Static"    RelativeTo="Root" Path="lib/static"/>
```
mlemuzic-bentley, line 67: IIRC libs are all static on linux, so not sure the "static" suffix is needed in the path "lib/static" for these three libs.

chuckkir: We create the same directory structure for Windows and Linux which allows for consistent consumption.

PR #451, opened 2023-09-13
Remove experimental feature option from instance access ecsql

RohitPtnkr1996:
Removed the experimental feature option check from instance access and instance property access.
Updated tests to remove the option.
Modified the options inheritance test to use USE_JS_PROP_NAMES option instead of Instance access to test inheritance in nested selects.

itwinjs-core: https://github.com/iTwin/itwinjs-core/tree/rohitptnkr/update-tests-instance-query-getAspects


iModelCore/ECDb/CHANGES.md
```diff
@@ -7,6 +7,19 @@ This document including important changes to syntax or file format.
 | Profile | `4.0.0.3` |
 | ECSQL   | `1.2.7.0` |
 
+## `9/13/2023`: Runtime instance and property accessor no longer experimental
+
+1. ECSql version updated `1.2.7.0` -> `1.2.8.0`.
```
khanaffan, line 12: You did not actually change the ECSQL version its define in ECDb.h

iModelCore/ECDb/CHANGES.md
```diff
@@ -7,6 +7,19 @@ This document including important changes to syntax or file format.
 | Profile | `4.0.0.3` |
 | ECSQL   | `1.2.7.0` |
```
khanaffan, line 8: also update this.

iModelCore/ECDb/CHANGES.md
khanaffan: There is a unit test that verifies ECSql version that needs to be updated.

khanaffan (changes requested):

khanaffan (approved):

PR #452, opened 2023-09-14
Fixed override property check when deleting a property

RohitPtnkr1996:
The check for overridden property was moved to no longer have  highest precedence when we added support to allow major schema changes for dynamic schemas.

Fixed the condition to ensure overridden properties can be deleted irrespective of the schema change (major or otherwise).
Added unit tests to cover the scenario for dynamic and non-dynamic schemas.


khanaffan (approved):

iModelCore/ECDb/ECDb/SchemaWriter.cpp
```diff
@@ -3436,7 +3436,7 @@ BentleyStatus SchemaWriter::DeleteProperty(Context& ctx, PropertyChange& propert
 
     // fail if major version changes are not allowed or major version has not been incremented, but continue if newBaseProperty is available, meaning we are merely removing an overridden property
     // Allow Major schema upgrade for dynamic schemas if AllowMajorSchemaUpgradeForDynamicSchemas import option is set irrespective of the DisallowMajorSchemaUpgrade import option
-    if (!ctx.IsMajorSchemaVersionChange(deletedProperty.GetClass().GetSchema().GetId()) && !isOverriddenProperty || (!ctx.AreMajorSchemaVersionChangesAllowed() && (!ctx.IsMajorSchemaVersionChangeAllowedForDynamicSchemas() || !isDynamicSchema)))
+    if ((!ctx.IsMajorSchemaVersionChange(deletedProperty.GetClass().GetSchema().GetId()) || (!ctx.AreMajorSchemaVersionChangesAllowed() && (!ctx.IsMajorSchemaVersionChangeAllowedForDynamicSchemas() || !isDynamicSchema))) && !isOverriddenProperty)
```
rschili, line 3439: Hey, I will approve, but I'd really like to see this if condition simplified. Can you try and break it into better readable chunks?
Reduce the number of necessary negations and brackets so everybody can easily see what is happening. Could be done using some temporary variables.

rschili (approved):

iModelCore/ECDb/ECDb/SchemaWriter.cpp
rschili: I was also thinking... since these rules and exceptions to the rules got so complex... It may be a good idea to include the reason for the error in the log message, since there can now be so many different reasons...

pmconne: I second this.

rschili: > Are these rules documented somewhere _outside of_ the implementation?

@khanaffan just recently pointed me to some documentation that he wrote a while ago about allowed schema modifications. We agreed it should be put and maintained somewhere more prominent. I just can't remember the link now, which is just confirming how not-discoverable it is :( Also it proably does not cover the dynamic schema checks added in this condition.

RohitPtnkr1996: Done, I've expanded the if conditions and added more comments to make the checks clearer.
Also added separate error messages for all the different possible scenarios.

rschili: Thanks, I'm satisfied, marking as resolved. I'll bring up the documentation thing with Colin separately.

rschili (approved):

PR #456, opened 2023-09-20
Fix CA loading when schema context has multiple versions of same schema

rschili:
Previously, the CA deserialization would go through the schema read context and may occasionally obtain a wrong (compatible) schema version which would then be used to load the CA and be added as a reference to the schema in which the CA is in.

The deserializer now prioritizes the container schema and its references when locating schemas used for custom attributes. Also added a defensive check to make sure schemas never reference another schema of the same name.

### DIFF
--- a/iModelCore/ecobjects/PublicApi/ECObjects/ECContext.h
+++ b/iModelCore/ecobjects/PublicApi/ECObjects/ECContext.h
@@ -223,14 +223,15 @@ struct ECInstanceReadContext : RefCountedBase
 private:
     IStandaloneEnablerLocaterP      m_standaloneEnablerLocater;
     ECSchemaCR                      m_fallBackSchema;
+    ECSchemaCP                      m_containerSchema;
     IPrimitiveTypeResolver const*   m_typeResolver;
     IUnitResolver const*            m_unitResolver;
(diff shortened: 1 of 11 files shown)


ColinKerr:
I think the same behavior can be achieved by creating a schema context that has just the schema and its references and use that instead of the larger schema context.  We may need to avoid creating a new scheme context for each CA that will be deserialized for performance reasons

rschili:
> I think the same behavior can be achieved by creating a schema context that has just the schema and its references and use that instead of the larger schema context. We may need to avoid creating a new scheme context for each CA that will be deserialized for performance reasons

After thinking about this some more, I really feel ECSchemaReadContext is a bad fit for this because it is too powerful.
We already have an abstraction here which is ECInstanceReadContext, and the existing implementation uses an ECSchemaReadContext. Alongside this existing one I'd like to put an ECInstanceReadContext which holds only the containerSchema.

I'll implement it like this and show it to you again. It is cheaper than constructing a full SchemaReadContext and going through all that LocateSchema/SchemaCache mechanisms when we actually know in advance that it must either be our current or one of its referenced schemas. There is no need to run complicated matching logic.

ColinKerr (approved):

PR #460, opened 2023-09-21
fix exports

MarcNeely:
These methods need to be exported for use by another project.

### DIFF
--- a/iModelCore/iModelPlatform/PublicAPI/DgnPlatform/Render.h
+++ b/iModelCore/iModelPlatform/PublicAPI/DgnPlatform/Render.h
@@ -621,8 +621,8 @@ struct TextureMapping
     struct Trans2x3
     {
         double m_val[2][3];
-        Trans2x3() {}
-        Trans2x3(double t00, double t01, double t02, double t10, double t11, double t12)
+        DGNPLATFORM_EXPORT Trans2x3() {}
+        DGNPLATFORM_EXPORT Trans2x3(double t00, double t01, double t02, double t10, double t11, double t12)
             {
             m_val[0][0] = t00;
             m_val[0][1] = t01;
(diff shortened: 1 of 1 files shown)


pmconne:
Thanks @MarcNeely.
@danieliborra @markschlosseratbentley if that other project is going to continue to reside in its current repository, we should update the build scripts to build it.

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/Render.h
```diff
@@ -621,8 +621,8 @@ struct TextureMapping
     struct Trans2x3
     {
         double m_val[2][3];
-        Trans2x3() {}
-        Trans2x3(double t00, double t01, double t02, double t10, double t11, double t12)
+        DGNPLATFORM_EXPORT Trans2x3() {}
```
pmconne, line 624: Did you find these necessary? It won't hurt anything, but functions defined inline shouldn't need an export attribute.

pmconne (approved):

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/Render.h
MarcNeely: Yes, you're right.  I misread the error & thought the constructor was what it was complaining about, but it was the FromJason.  I'll remove it from the constructors & test it out to make sure.

MarcNeely: OK, Done.

danieliborra:
> Thanks @MarcNeely. @danieliborra @markschlosseratbentley if that other project is going to continue to reside in its current repository, we should update the build scripts to build it.

You mean having a pipeline in imodel-native that compiles imodel-native AND imodel-native-internal, right?

PR #461, opened 2023-09-21
Fix typo in SolveBezierCubic declaration

tcobbs-bentley:
It lists the `uvals` array as being of size 2 when the C++ file lists (and requires) 3. The incorrect code causes a compile error in Xcode 15.


dassaf4 (approved):
You beat us to it!  Thanks @tcobbs-bentley.

PR #464, opened 2023-09-22
Add IssueId to Badgers logs in ECObjects and ECDb

akshay05n:
The IssueId has been integrated into the Badgers logs, a change that originated from our internal development processes.
This enhancement affects both ECObjects and ECDb, ensuring consistent logging practices across our ecosystem.


rschili (approved):

ColinKerr (approved):

nick4598:
How are these issueIds used? Is it to uniquely identify every possible source of an error message which could show up in a badgers report? I.e. issue id 100 shows up in badgers report, that id can be given to the team and they instantly know which line of code produced that issue in the report?

rschili:
> How are these issueIds used? Is it to uniquely identify every possible source of an error message which could show up in a badgers report? I.e. issue id 100 shows up in badgers report, that id can be given to the team and they instantly know which line of code produced that issue in the report?

This is just a merge from the imodel02 code stream. The primary goal in as far as I understand was so we can group log messages in reports by ID so we can see and count the number of messages unique by id.

pmconne:
> This is just a merge from the imodel02 code stream. The primary goal in as far as I understand was so we can group log messages in reports by ID so we can see and count the number of messages unique by id.

But the implementation of _OnIssueReported in the addon completely ignores the new issue Id argument.
Where do these issues actually get logged?

ColinKerr:
> > This is just a merge from the imodel02 code stream. The primary goal in as far as I understand was so we can group log messages in reports by ID so we can see and count the number of messages unique by id.
> 
> But the implementation of _OnIssueReported in the addon completely ignores the new IssueId argument (and the old IssueType argument). Where do these issues actually get logged?

At the moment we are just merging this in from the connector codebase to keep things in sync.  In the future we will collaborate with the connector team so TypeScript connectors and other editors can use the structured report message.

pmconne (approved):

PR #465, opened 2023-09-22
 Pragma disqualify_type_filter only take effect if there was more then one class name in query

khanaffan:


ColinKerr (approved):

khanaffan:
@Mergifyio backport release/4.x

Issue #458, opened 2023-09-21
Clip of linear sweep of parity region produces no facets


dassaf4:
Bug #[1279805](https://dev.azure.com/bentleycs/iModelTechnologies/_workitems/edit/1279805)

- [x] fix in imodel-native: https://github.com/iTwin/imodel-native/pull/466
- [x] backport to 4.1: https://github.com/iTwin/imodel-native/pull/471 (thanks, Arun)
- [x] port to imodel02: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel02/pullrequest/370824
- [x] port to PPBase: https://dev.azure.com/bentleycs/PowerPlatform/_git/PPBase/pullrequest/370898

Clip the geometry below with `ClipPlaneSet::ClipPlaneSetSectionPolyface`

> {"linearSweep":{"contour":{"parityRegion":[{"loop":[{"lineString":[[-3.2977383001558134,6.235758360584979,-0.10000000000000053],[-6.46664378235049,3.856966271116766,-0.10000000000000053],[-6.46664378235049,3.856966271116766,8.418576],[-3.2977383001558134,6.235758360584979,8.418576],[-3.2977383001558134,6.235758360584979,-0.10000000000000053]]}]},{"loop":[{"arc":{"center":[-5.06501251137976,4.909124310689251,1.0667999999999997],"vectorX":[0.8531668605908739,0.6404440240875956,0],"vectorY":[0,0,1.0668000000000002],"sweepStartEnd":[179.99999999999983,539.9999999999998]}}]}]},"capped":true,"vector":[0.5489520206465107,-0.7312858805064635,0]}}

Solid in question is on the left. The clipper appears to have a single relevant plane, horizontal at top, but actually the problematic plane is at bottom, through the circular cutout. User expects facets on the clip planes, like the solid on the right.

![Image](https://github.com/iTwin/imodel-native/assets/68340676/8b66991b-66f5-4898-9e12-2e4d163da1d9)

PR #466, opened 2023-09-25
disconnect in mesh ruins point range

dassaf4:
Addresses #458.

Disconnect points are used to separate loops when triangulating a ParityRegion in vu. When meshing a capped DgnExtrusion created from a ParityRegion, this causes a disconnect to be installed in the builder's point map, where it is unreferenced by face loops and thus mostly harmless... until someone downstream computes a range from the Polyface point array, and it blows up, such as what happens when you hit one of these meshes with a ClipPlane.

The least risky solution with (no impact to polyface index arrays) is to reset the unreferenced disconnect to something in range (e.g., a non-disconnect vertex).


aruniverse:

iModelCore/GeomLibs/geom/test/PolyfaceTest/t_ClipPlaneSet.cpp
```diff
@@ -2666,3 +2666,43 @@ TEST(ClipPlaneSet,CloneBetweenDirectedFractions)
 
     Check::ClearGeometry("ClipPlaneSet.CloneBetweenDirectedFractions");
     }
+
+TEST(ClipPlaneSet, ClipParityRegionSweep)   // verifies the fix for Bug #1279805
```
saeeedtorabi, line 2670: Not sure if merging this comment makes sense. All code changes are to fix some bugs!

iModelCore/GeomLibs/geom/src/polyface/IPolyfaceConstruction_Add.cpp
```diff
@@ -2553,6 +2573,8 @@ bool IPolyfaceConstruction::AddRuledBetweenCorrespondingCurves
                 }
             if (capped)
                 context.EmitCaps (reversedPrimary);
+            // Bug #1279805: disconnects used to separate loops get copied into mesh xyz and prevent future clipping
```
saeeedtorabi, line 2576: I dont think we want to merge bug numbers in our repos!

iModelCore/GeomLibs/geom/test/PolyfaceTest/t_ClipPlaneSet.cpp
dassaf4: This is an old habit I found immensely useful in pre-GitHub days, but you're right. With VSCode's GitHub integration giving context for every line of code, it's mostly not needed anymore.

iModelCore/GeomLibs/geom/src/polyface/IPolyfaceConstruction_Add.cpp
dassaf4: I'm not aware of any proscription on this, but I agree that it doesn't add value these days. Old habits die hard!

saeeedtorabi (approved):

PR #470, opened 2023-09-27
Schema should be serialized as ECXml 3.2 if ECXml 3.1 fails due to incorrect KoQ

RohitPtnkr1996:
We have instances where there are ECXml 3.1 schemas that have KoQs using ECXml 3.2 units as their persistence units.

When these schemas are exported (serialized to a file/xml string), we get a failure as the 3.1 logic only looks at the legacy unit mappings and not the referenced ECXml 3.2 Units and Formats schemas.

With this change, if serializing an ECXml 3.1 schema to a XML string/file fails due to an incorrect KoQ, we will retry serializing the schema as ECXml 3.2.


pmconne (commented):
Is there something in export-schema-test.bim that cannot be created by the test at run-time, making it necessary to add a large binary file to the repo?

ColinKerr (approved):

ColinKerr:
> Is there something in export-schema-test.bim that cannot be created by the test at run-time, making it necessary to add a large binary file to the repo?

Maybe it's possible to create programmatically by constructing a new schema and setting the ECVersion in the constructor to 3.1.  That will set the original xml version, you can then update to 3.2 using SetECVersion and add the unit.

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3557,9 +3557,22 @@ SchemaWriteStatus ECSchema::WriteToXmlString(WStringR ecSchemaXml, ECVersion ecX
 
     BeXmlWriterPtr xmlWriter = BeXmlWriter::Create();
 
-    SchemaWriteStatus status;
     SchemaXmlWriter schemaWriter(*xmlWriter.get(), *this, ecXmlVersion);
-    if (SchemaWriteStatus::Success != (status = schemaWriter.Serialize()))
+
```
rschili, line 3561: It doesn't seem ideal to have this code three times with only minor differences. SchemaWriter::Serialize() may be a better place to deal with this. Also, at this spot here we can only assume the failure was due to the KoQ thing, we do not really know. It may be a different future error which we don't know yet, but still we'd be logging a KoQ problem.

I haven't checked if there is a way to reset BeXmlWriter, but if it's possible, we could handle this inside SchemaWriter::Serialize() at WriteSchemaChild<KindOfQuantity>() so if we get back an error code we can be absolutely sure it is the KoQ problem and nothing else.

RohitPtnkr1996: I had checked this and currently BeXmlWriter doesn't have a way to rest the writer.
Also SchemaWriter has no way to reset the EcXml Version of an object with a setter.
So, I decided to put the fix it at the last common point.
I can maybe add a reset function to BeXmlWriter and a setter to SchemaWriter.

I see your point about future non-koq errors returning FailedToSaveXml and confusing the logs.
I'll try moving the code in Serialise() to resolve this.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -649,6 +649,7 @@ export declare namespace IModelJsNative {
     public enableWalMode(yesNo?: boolean): void;
     public performCheckpoint(mode?: WalCheckpointMode): void;
     public setAutoCheckpointThreshold(frames: number): void;
+    public setSchemaECVersion(schemaName: string, ecVersion: ECVersion): boolean
```
pmconne, line 652: Is this method strictly for your test? If not, in what scenarios would core-backend call it?

iModelCore/ecobjects/src/ECSchema.cpp
RohitPtnkr1996: Had a word with Colin, he prefers to keep the check here itself.
Since the error message already contains the reason for the failure, I've made the new log message more generic by just mentioning that the schema will be re-serialized as ECXml 3.2.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
RohitPtnkr1996: Sorry !
Premature commit, more changes incoming.

iModelJsNodeAddon/api_package/ts/package.json
```diff
@@ -21,12 +21,14 @@
     "@itwin/core-common": "rc",
     "@itwin/core-geometry": "rc",
     "@itwin/eslint-plugin": "rc",
+    "@types/better-sqlite3": "^7.6.5",
```
pmconne, line 24: Is `IModelJsNative.SqliteDb` missing something you need, to instead motivate adding this new dev dependency?

pmconne (approved):

rschili (approved):

PR #472, opened 2023-09-27
Merge new version of crashpad to address Mend errors

chuckkir:


pmconne (approved):
Holy cannoli that's a lot of files.

chuckkir:
> Holy cannoli that's a lot of files.

There are a lot of files in crashpad I guess. I think they changed their copyright message slightly which affected a lot for files and quite a few of the files were deleted: from 3198 files to 1329 files. Most of the removed files were tests / test data and additional apps. I used a different tactic than the previous crashpad import so it picked up fewer files. 

The large number of changesets is a direct result of forcing all commits to be Squash. Since these are vendor pulls, we put them on a separate clean branch and then merge them in. This way Bentley changes don't get lost. However, merging in git is based on changesets, so when things get squashed it doesn't relate to the changesets that were already merged, and thus thinks every changeset is part of the merge. It's not a great workflow.

nick4598:

PR #473, opened 2023-09-27
Fix error were db fail to close due to open blobio handle held by FontManager

khanaffan:
The issue was when Transformer called `CloseDb()` after transformation we failed to close db. It turn out that particular imodel had uncompress font which use BlobIo (unlike compressed font which use queryProperty to read font face). 

`CloseDb()` caused `~DbFile()` to be called before `~DgnDb()`. `FontManager` was destroyed in `~DgnDb()` instead of `CloseDb()`.

The fix is to delete FontManager in `DgnDb::Destroy()` so it releases BlobIO handles (if any) that could prevent db from getting closed.


pmconne (approved):

khanaffan:
@Mergifyio backport  release/4.x release/3.x

PR #475, opened 2023-09-28
Fix to line style max offset calculation to properly account for modifiers.

bbastings:
Partial fix for [issue](https://github.com/iTwin/itwinjs-core/issues/6010#event-10498971483l) also need change to batch connector.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6039

Previously padding only considered width of line style definition without considering modifiers which can both increase or decrease the overall width. There was a special case for continuous with width, but this did not account for point symbol or stroke pattern width.

This also fixes keeping the line style name map up to date when new styles are inserted. This previously was done in the C++ method and has now been move to the _OnInsert method of the line style definition element.


pmconne:
Is this fix needed in the imodel02 repository? Presumably that is how the iModel in the issue was created.

pmconne (approved):

bbastings:
> Is this fix needed in the imodel02 repository? Presumably that is how the iModel in the issue was created.

Yes, I'm doing that PR now. It doesn't need the _OnInsert changes, just the max offset calc.

PR #478, opened 2023-09-28
Fix exception handling problem on Apple builds

tcobbs-bentley:
Since Apple builds don't include runtime type information (RTTI), exception handlers meant to catch a specific type and its subclasses don't work right, so the subclasses have to be caught explicitly, along with a ... fallback.

### DIFF
--- a/iModelJsNodeAddon/JsInteropDgnDb.cpp
+++ b/iModelJsNodeAddon/JsInteropDgnDb.cpp
@@ -392,6 +392,17 @@ Napi::String JsInterop::InsertElement(DgnDbR dgndb, Napi::Object obj, Napi::Valu
         return Napi::String::New(Env(), newEl->GetElementId().ToHexStr());
     } catch (std::logic_error const& err) {
         BeNapi::ThrowJsException(Env(), err.what(), (int)DgnDbStatus::BadArg);
+#if defined (__APPLE__)
+    // Apple builds don't include runtime type information (RTTI), which means that exception
+    // handling doesn't support subclasses. So even though std::invalid_argument is a subclass of
+    // std::logic_error, the above catch does not work.
+    } catch (std::invalid_argument const& err) {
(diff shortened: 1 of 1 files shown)


pmconne:
> Apple builds don't include runtime type information (RTTI)

Is that intrinsic to the compiler, or the way we configure the compiler? Presumably it only applies to exceptions, not all types?

tcobbs-bentley:
> > Apple builds don't include runtime type information (RTTI)
> 
> Is that intrinsic to the compiler, or the way we configure the compiler? Presumably it only applies to exceptions, not all types?

There is an option (`-frtti`) that enables runtime type information, but unfortunately that option isn't enabled for our nuget build of Google V8, so enabling it for imodel-native results in link errors between Node and V8. I am investigating generating a V8 nuget that is compiled with the option enabled, but I'm not sure if it will work, and I have no idea how long it will take me to create.

pmconne:
> I am investigating generating a V8 nuget that is compiled with the option enabled,

That seems like a better solution. Is this PR solving an urgent problem or can it wait until you have completed your investigation?

tcobbs-bentley:
> > I am investigating generating a V8 nuget that is compiled with the option enabled,
> 
> That seems like a better solution. Is this PR solving an urgent problem or can it wait until you have completed your investigation?

I'm trying to get unit tests working on iOS again, and without any fix they simply crash. With this fix they run (although there are 13 tests that don't pass that I have to look at). Unit tests on iOS have been broken for a long time, but the longer they don't run, the longer we have to live with not knowing if things work, so it's hard to gauge the level of urgency.

I've personally never done any buildology for Google V8, so while I will try to get RTTI into it, I'm unsure about my chances of success.

tcobbs-bentley:
In my opinion, doing this now and reverting it if I can get RTTI into V8 is preferred. However, I can see the argument for not approving this PR.

pmconne (approved):
Temp workaround until we determine whether we can enable rtti or have a good reason not to.

PR #481, opened 2023-09-29
Add `base36` ECSQL function

grigasp:
We need ability to convert a decimal number to base36 to create BIS element suffixes using ECSQL. There's no way to achieve that through pure SQLite as far as I can tell, so I added it as a custom function.

Also, updated ECPresentation code to use less of its own custom functions and instead rely on bitwise operators and the new `base36` ECSQL function.


saskliutas (approved):

khanaffan (approved):

grigasp:
@khanaffan, the CI build is failing because our worker connections don't see the new `base36` function. Looking at [this line](https://github.com/iTwin/imodel-native/blob/fec78849e9f848c506585d7bc468c02c5f2fce55/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp#L215), I believe it won't be available on your worker connections either. 

Before I make any changes, could you please help me with the following:
1. What's the reason of not registering **all** primary db functions on worker connections?
2. Could that be done in `BeSQLite::Db::OpenSecondaryConnection` so it's callers don't need to worry about re-registering those functions?

khanaffan:
> @khanaffan, the CI build is failing because our worker connections don't see the new `base36` function. Looking at [this line](https://github.com/iTwin/imodel-native/blob/fec78849e9f848c506585d7bc468c02c5f2fce55/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp#L215), I believe it won't be available on your worker connections either.
> 
> Before I make any changes, could you please help me with the following:
> 
> 1. What's the reason of not registering **all** primary db functions on worker connections?
> 2. Could that be done in `BeSQLite::Db::OpenSecondaryConnection` so it's callers don't need to worry about re-registering those functions?

You can just use `ECDb` as a worker connection instead of `Db`. I did that in ConcurrentQuery.  The copying function caused issues for me in the past where it held the primary connection as a member variable causing concurrent queries to wait for each other. So, I moved up from Db to ECDb.

You can also move your function up in BeSQLite as well. SQLite [ext/misc ](https://sqlite.org/src/dir?ci=trunk&name=ext/misc) has "base64.c" and "base85.c". There are optional extensions and we do not build with it but it shows base36 is a primitive function and does not require ecdb or higher profile.

grigasp:
> You can just use `ECDb` as a worker connection instead of `Db`. I did that in ConcurrentQuery. The copying function caused issues for me in the past where it held the primary connection as a member variable causing concurrent queries to wait for each other. So, I moved up from Db to ECDb.
> 
> You can also move your function up in BeSQLite as well. SQLite [ext/misc ](https://sqlite.org/src/dir?ci=trunk&name=ext/misc) has "base64.c" and "base85.c". There are optional extensions and we do not build with it but it shows base36 is a primitive function and does not require ecdb or higher profile.

I chose the latter approach - reverted the change that registered the function in `ECDb` and moved it to `BeSQLite::Db`. Please have a look.

PR #487, opened 2023-10-05
Modify Default Branch Setup

DanRod1999:


DanRod1999:
I forgot that since this change is inside the azure-pipeline.yml file this PR wont work until the internal PR is merged. This means we will need to wait an hour for this to build after internal merges with the appropriate changes.

This will break imodel-native PR val for that hour until this PR is merged, or I can add some extra code to my internal PR and that should keep it running. This is fine, but kinda defeats the purpose of some of my changes, and also would technically be unnecessary once this PR merges.

I also don't mind waiting to merge my PRs late tonight when there's little traffic happening. This would be my preferred solution so I don't need to add redundant code(or remove this redundant code later). Any suggestions @calebmshafer @chuckkir

chuckkir:
For stuff like this I just wait until the 8 PM - midnight slack period (CA leaves but it's a while before Asia starts) or a weekend. I don't think it's urgent.

DanRod1999:
> For stuff like this I just wait until the 8 PM - midnight slack period (CA leaves but it's a while before Asia starts) or a weekend. I don't think it's urgent.

Sounds good, thanks. Then this PR is good for review, but I won't touch them until this evening.

chuckkir (approved):

DanRod1999:

PR #488, opened 2023-10-05
Bug Fix: Resolve StandardValues custom attribute conversion issue in nested schema reference

akshay05n:
**Bug Details:**
We have a custom attribute called `StandardValues` on a property. This property is part of `SchemaC::ClassC`, which inherits from `SchemaB::ClassB`, and `SchemaB::ClassB` further inherits from `SchemaA::ClassA`. The converter fails to set the type for `ClassC` in `SchemaC` because the enumeration is defined in `SchemaA` and `SchemaC` only references `SchemaB`.

**Changes Made:**
To resolve this issue, we will add a direct schema reference (if it doesn't already exist) whenever we directly reference a schema item from another schema.


iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3446,12 +3446,29 @@ POP_DISABLE_DEPRECATION_WARNINGS
...
+    return false;
+    }
+
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
 bool ECSchema::IsSchemaReferenced(ECSchemaCR thisSchema, ECSchemaCR potentiallyReferencedSchema)
     {
-    return ECSchema::IsSchemaReferenced(thisSchema, potentiallyReferencedSchema, ECN::SchemaMatchType::Exact);
+    return IndirectlyReferences(thisSchema, potentiallyReferencedSchema, ECN::SchemaMatchType::Exact);
```
ColinKerr, line 3471: We should not change the behavior of IsSchemaReferenced.  It is really supposed to look in the direct references only.  If we were going to change the behavior of this overload of IsSchemaReferenced we would also have to change the other one so it was consistent.  Changing just one is very unexpected.  We should not change either.

The method 'DirectlyReferences' is confusingly named as it tests memory reference instead of name.  Luckily it is an internal method.

I kind of suspect that the root class passed into the convert method is not correct.  Does it only go one level deep?  You can always find the root class by walking the BaseProperty of each property until it's null ... that's the root class that defined the property first.

In general I think the algorithm is supposed to work at the base property then change derived properties as it goes.  In this case I guess it doesn't start at the base because the CA is defined on the leaf

When writing the test you should also validate that the intermediate property has been converted to the correct type.

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3446,12 +3446,29 @@ POP_DISABLE_DEPRECATION_WARNINGS
     return status;
     }
 
+/*---------------------------------------------------------------------------------**//**
+* Returns true if thisSchema indirectly references potentiallyReferencedSchema
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+static bool IndirectlyReferences(ECSchemaCR thisSchema, ECSchemaCR potentiallyReferencedSchema, ECN::SchemaMatchType matchType)
```
rschili, line 3453: This name is ambiguous. One might (even if it makes no sense) assume that it checks if a schema is indirectly referenced (and return false if it's directly referenced)
It's probably a good idea to refactor/rename the "DirectlyReferences" method and make it take a boolean parameter instead which sets the behavior (direct only vs all)

iModelCore/ecobjects/src/ECSchema.cpp
akshay05n: > I kind of suspect that the root class passed into the convert method is not correct. Does it only go one level deep?

The `ConvertToEnum()` method receives the root class `baseTest:TestClass` and traverses the class hierarchy recursively. However, when it attempts to set the type, the `PrimitiveECProperty::SetType()` method only checks if the `test:TestClass` class references `baseTest:TestClass`. This returns an `ECObjectsStatus::SchemaNotFound` status because we only check if `baseTest:TestClass` is directly referenced by `test:TestClass`.

iModelCore/ecobjects/test/NonPublished/CustomAttributeConversionsTest.cpp
```diff
@@ -2133,6 +2133,75 @@ TEST_F(StandardValueToEnumConversionTest, TestRootClassInSeparateSchema)
...
+    ASSERT_TRUE(middleSchema.IsValid());
+    ASSERT_EQ(SchemaReadStatus::Success, ECSchema::ReadFromXmlString(schema, schemaXml, *context));
+    ASSERT_TRUE(schema.IsValid());
+ 
+    EXPECT_TRUE(ECSchemaConverter::Convert(*schema.get(), *context.get())) << "Schema conversion failed";
+ 
+    ASSERT_EQ(0, schema->GetEnumerationCount()) << "The number of enumerations created is not as expected.";
+    ASSERT_EQ(1, baseSchema->GetEnumerationCount()) << "The number of enumerations created is not as expected.";
+
+    CheckTypeName("ts:TestClass_testProp", *schema, "testProp", {"TestClass"});
```
rschili, line 2201: We should CheckTypeName on the middleSchema as well, to make sure it's properly wired up. In this case both, the schema and middleSchema use the same alias for "base", you could write another test where they use different aliases and make sure the GetTypeName() method returns the correct one for each level.

We could also add another test where the middle schema does not have its own declaration of testProp, and check if it still behaves the same.

akshay05n: Added additional tests

iModelCore/ecobjects/src/ECProperty.cpp
```diff
@@ -1399,7 +1399,7 @@ ECObjectsStatus PrimitiveECProperty::SetType (ECEnumerationCR enumerationType)
     {
     if (&(enumerationType.GetSchema()) != &(this->GetClass().GetSchema()))
         {
-        if (!ECSchema::IsSchemaReferenced(this->GetClass().GetSchema(), enumerationType.GetSchema()))
+        if (!ECSchema::IsSchemaReferenced(this->GetClass().GetSchema(), enumerationType.GetSchema(), ECN::SchemaMatchType::Exact, true))
```
ColinKerr, line 1402: A schema must directly reference a schema to use something from it.  The schema that owns the enumeration must be added as a reference before the type can be set.

iModelCore/ecobjects/src/ECSchemaConverter.cpp
```diff
@@ -873,6 +873,10 @@ ECObjectsStatus StandardValuesConverter::ConvertToEnum(ECClassP rootClass, ECCla
         prop->RemoveCustomAttribute(BECA_SCHEMANAME, STANDARDVALUES_CUSTOMATTRIBUTE);
         }
 
+    // Adds a direct reference to the grandparent schema, if missing
+    if (ECObjectsStatus::Success == status)
```
ColinKerr, line 877: This seems misplaced.  There could be an arbitrary depth of hierarchy and multiple branches in the hierarchy.  The algorithm needs to handle adding the reference integrated to setting the type.

iModelCore/ecobjects/test/NonPublished/CustomAttributeConversionsTest.cpp
```diff
@@ -2133,6 +2133,217 @@ TEST_F(StandardValueToEnumConversionTest, TestRootClassInSeparateSchema)
...
+ 
+    EXPECT_TRUE(ECSchemaConverter::Convert(*schema.get(), *context.get())) << "Schema conversion failed";
+ 
+    ASSERT_EQ(1, schema->GetEnumerationCount()) << "The number of enumerations created is not as expected.";
+    ASSERT_EQ(0, middleSchema->GetEnumerationCount()) << "The number of enumerations created is not as expected.";
+    ASSERT_EQ(0, baseSchema->GetEnumerationCount()) << "The number of enumerations created is not as expected.";
+
+    EXPECT_EQ(nullptr, middleSchema->GetClassP("TestClass")->GetPropertyP("testProp"));
+    CheckTypeName("TestClass_testProp", *schema, "testProp", {"TestClass"});
+    CheckTypeName("int", *baseSchema, "testProp", {"TestClass"});
```
ColinKerr, line 2344: This is wrong and I think highlights that the solution proposed is fundamentally wrong.  All properties in the hierarchy should be converted to use the enumeration and the enumeration must be defined in the base schema.  

If a property override exists that uses this enum then the schema must have a direct reference to the schema defining the enum

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -3987,7 +3987,13 @@ friend struct SchemaJsonWriter;
     //! @param[in]    thisSchema                   The base schema to check the references of
     //! @param[in]    potentiallyReferencedSchema  The schema to search for
     //! @return True if thatSchema is referenced by thisSchema, false otherwise
-    ECOBJECTS_EXPORT static bool IsSchemaReferenced (ECSchemaCR thisSchema, ECSchemaCR potentiallyReferencedSchema, ECN::SchemaMatchType matchType);
+    ECOBJECTS_EXPORT static bool IsSchemaReferenced (ECSchemaCR thisSchema, ECSchemaCR potentiallyReferencedSchema, ECN::SchemaMatchType matchType, bool checkIndirectReferences = false);
```
ColinKerr, line 3990: Do not complicate the IsScheamReferenced method with a check to indirect references.  I'm not sure why we need this check to solve the current problem.

akshay05n: Removed `checkIndirectReferences` optional parameter

iModelCore/ecobjects/src/ECSchemaConverter.cpp
akshay05n: Moved code to add schema reference in `PrimitiveECProperty::SetType()` function

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3483,7 +3478,7 @@ static bool DirectlyReferences(ECSchemaCP thisSchema, ECSchemaCP possiblyReferen
 //---------------------------------------------------------------------------------------
 static bool DependsOn(ECSchemaCP thisSchema, ECSchemaCP possibleDependency)
     {
-    if (DirectlyReferences(thisSchema, possibleDependency))
```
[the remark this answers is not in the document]
rschili: I'm fine with that. I also don’t see a problem removing it and comparing by key. I can imagine a small chance of this breaking consumers which do wonky things, but that's probably a good thing in this case.

iModelCore/ecobjects/test/NonPublished/CustomAttributeConversionsTest.cpp
akshay05n: I initially planned to remove only the property from the `mid:TestClass` class, but I accidentally removed the base class as well. I have updated the test case, and we can see that the enum is only created in the base schema.

iModelCore/ecobjects/src/ECProperty.cpp
```diff
@@ -1397,10 +1397,15 @@ ECObjectsStatus PrimitiveECProperty::SetType (PrimitiveType primitiveType)
 +---------------+---------------+---------------+---------------+---------------+------*/
 ECObjectsStatus PrimitiveECProperty::SetType (ECEnumerationCR enumerationType)
     {
-    if (&(enumerationType.GetSchema()) != &(this->GetClass().GetSchema()))
+    ECSchemaCR enumSchema = enumerationType.GetSchema();
+    ECSchemaCR currentSchema = this->GetClass().GetSchema();
+    if (&enumSchema != &currentSchema)
         {
-        if (!ECSchema::IsSchemaReferenced(this->GetClass().GetSchema(), enumerationType.GetSchema(), ECN::SchemaMatchType::Exact, true))
+        if (!ECSchema::IsSchemaReferenced(currentSchema, enumSchema) && ECObjectsStatus::Success != const_cast<ECSchemaR>(currentSchema).AddReferencedSchema(const_cast<ECSchemaR>(enumSchema)))
```
ColinKerr, line 1404: The Set method should not add the schema reference, it should only check and return an error if the reference is not found.

This is the consistent behavior for all SetType overloads (Enumeration, struct and struct array), add base class, adding a custom attribute (except for deserialization to handle schemas created w/o property references to shipped bugs), setting a properties category, KOQ, nav prop relationship, etc.

It's a bit of a pain but it is worse if we are inconsistent.  So the calling code will have to check and add the reference if it is necessary.

Also class also has a GetSchemaR method to get the non const reference ... it just does the const_cast but it looks cleaner on the calling side.

akshay05n: I have updated the code based on your feedback.

ColinKerr (approved):

PR #501, opened 2023-10-11
Presentation: Correctly determine select classes based on available fields

saskliutas:
Fixed select classes lookup based on fields. Some condition were backwards and was falling in some cases.
* Lookup based on properties field was working because properties stored in field had class used to access that property (select class) instead of actual class where that property is defined. Because this condition is associative, I turned around it to match other cases.
* Lookup based on related content field was not working correctly when same field is created for multiple select classes that share base source class. Turned around condition to check if select class is equal or derived from related field source class.


grigasp (approved):

grigasp:

PR #505, opened 2023-10-11
add invalid category err

DanRod1999:
someone recently ran into an invalid category error, but was given the hex error code in return which was hard to decipher.


DanRod1999 (commented):
if spacing should be change let me know. If not this can be resolved

pmconne (approved):

khanaffan (approved):

PR #510, opened 2023-10-11
remove ci trigger for main and release branches [buildNum:2024.02.21.15]

nick4598:
Motivation behind this: 

Whenever we release a new addon we first start by running bump-version.yml, which bumps the package_version.txt by a commit to main / release branch from imodeljs-admin. This causes two pipeline runs to trigger, one as a result of the trigger which I am removing in this PR and one triggered by the bump-verson.yml pipeline. 

The pipeline run as a result of the trigger is useless and wastes build time.

We could potentially special case the scenario where we're releasing a new addon and only skip the trigger for that scenario(when imodeljs-admin pushes to the branch, Scenario A), but I don't see a benefit in the other scenario that this pipeline would trigger and thats when a PR merges to main/release branch (Scenario B). I could see value in running CI/ the trigger in scenario B if we didn't already have the github policy to require that our PRs were up to date with main. @aruniverse could you confirm this policy is in place for the imodel-native repo?

The value that it would bring without the github policy is that it would confirm that whatever is in main + the PR that merged runs nicely together(no errors). With the github policy PRs are always running with whatever is already in main so it adds no value. 

Please correct me if I am wrong regarding scenarios A / B or if there is some other valuable reason to have these triggers.


rschili (approved):

aruniverse:
Is this pr still needed?

nick4598:
I would assume we're still wasting build time on a merge to main with this. But, looking at the imodel-native-internal pipeline, it deosn't seem to have an issue.

image

I took a look at [imodel-native](https://bentleycs.visualstudio.com/iModelTechnologies/_build?definitionId=8443)

And it does seem pointless to have some of these pipelines run 'Individual CI for main'
I also wonder if even imodeljs-admin's commit causing a pipeline run is not that useful. It doesn't seem to be a release build, do we need to run non release builds off of main when with Dan's changes we're guaranteed to run a release build for every change into main?

PR #519, opened 2023-10-16
Presentation: Get instance ids from instance symbol context

grigasp:
When creating a calculated property with a custom renderer, we may want to know `ECInstanceId` of the instance the property was created for. Obvious way to achieve that is to assign `this.ECInstanceId` as the raw calculated property value, but evaluation of this ECExpression is failing because `ECInstanceId` is not an `ECProperty`. This PR adds special handling for it so we can retrieve the ID in hex format.


saskliutas (approved):

iModelCore/ecobjects/src/ExpressionContext.cpp
```diff
@@ -509,6 +503,17 @@ static ExpressionStatus GetInstanceValue (EvaluationResultR evalResult, uint32_t
         evalResult.Clear();
         if (NULL == currentProperty)
             {
+            if (accessString.Equals("ECInstanceId"))
+                {
+                evalResult = ECValue(BeInt64Id::FromString(instance.GetInstanceId().c_str()).ToHexStr().c_str());
```
rschili, line 508: I believe this needs some error handling. BeInt64Id::FromString() takes an optional BentleyStatus pointer as second parameter, and that returns ERROR for invalid IDs. Like, when it's an empty string, or a negative number or decimal number, or alphanumeric letters.
We should cover this scenario with tests. What is supposed to happen with non-numeric Instance IDs? Also, maybe we'll need a non-hex version of this in the future.

grigasp: > I believe this needs some error handling. BeInt64Id::FromString() takes an optional BentleyStatus pointer as second parameter, and that returns ERROR for invalid IDs. Like, when it's an empty string, or a negative number or decimal number, or alphanumeric letters. We should cover this scenario with tests. What is supposed to happen with non-numeric Instance IDs?

Do we have any real scenarios in iTwin.js when an `ECInstance` has empty, negative, decimal, etc. `ECInstanceId`?

> Also, maybe we'll need a non-hex version of this in the future.

We could change result of these expressions to `int64` and then introduce a function symbol `Id64ToHex` or similar to convert it to hex. How does that sound?

rschili: I'll discuss with the rest of the team in our call and get back to you. We try to keep the old imodel02 and imodel-native code in sync, so even if we don't have non-number Ids in iTwin, this code will probably affect others. Since the API uses strings, it seems very possible we will run into the case.

rschili: We discussed and it seems safe to assume only numbers here, I'm closing the discussion.

rschili (approved):

iModelCore/ecobjects/src/ExpressionContext.cpp
ColinKerr: I think it's fine to do this.  ids are always ints in the context of an iModel.  Even on imodel02 ids are always ints.  When elements are brought over from dgn side they may have their v8 ids but calculated properties are stripped and not run so this won't possibly impact that.

grigasp:

PR #529, opened 2023-10-25
Added new flag to allow class and property deletions with a minor schema increment

RohitPtnkr1996:
Added a new schema import flag "AlwaysAllowDeletions" which is disabled by default.
User needs to call SetAlwaysAllowDeletions(bool) to enable the flag.

This flag will allow class and property deletion irrespective of whether the major version is incremented.

As of right now, this flag is only checked/used by the DgnDb::ImportV8LegacySchemas().
The normal DgnDb::ImportSchemas() remains unaffected.


iModelCore/ECDb/ECDb/SchemaWriter.cpp
```diff
@@ -5236,7 +5247,7 @@ BentleyStatus SchemaWriter::UpdateSchema(Context& ctx, SchemaChange& schemaChang
             return ERROR;
             }
 
-        if (!ctx.AreMajorSchemaVersionChangesAllowed() && !(newSchema.IsDynamicSchema() && ctx.IsMajorSchemaVersionChangeAllowedForDynamicSchemas()))
+        if (!ctx.IsAlwaysAllowDeletionsEnabled() && !ctx.AreMajorSchemaVersionChangesAllowed() && !(newSchema.IsDynamicSchema() && ctx.IsMajorSchemaVersionChangeAllowedForDynamicSchemas()))
```
rschili, line 5250: This one here seems not logical to me. The flag does not state that is allows incrementing the read version. There is a separate flag for that already (AreMajowSchemaVersionChangesAllowed) so that side effect seems unnecessary.

rschili (approved):

PR #532, opened 2023-10-26
Add support to delete KoQs and Enums with a major schema version change

RohitPtnkr1996:


iModelCore/ECDb/Tests/NonPublished/SchemaUpgradeTests.cpp
```diff
@@ -10622,6 +10622,82 @@ TEST_F(SchemaUpgradeTestFixture, DeleteKindOfQuantityFromECSchema)
     AssertSchemaUpdate(editedSchemaXml, filePath, {false, false}, "Deleting KindOfQuantity from an ECSchema");
     }
 
+TEST_F(SchemaUpgradeTestFixture, DeleteKoQWithMajorSchemaChangeShouldPass)
+    {
+    SchemaItem schemaItem(R"xml(
+        <?xml version='1.0' encoding='utf-8'?>
+        <ECSchema schemaName='TestSchema' alias='ts' version='1.0.0' xmlns='http://www.bentley.com/schemas/Bentley.ECXML.3.1'>
```
ColinKerr, line 10629: Since this is a new test use an EC 3.2 schema

iModelCore/ECDb/Tests/NonPublished/SchemaUpgradeTests.cpp
```diff
@@ -10945,20 +11021,21 @@ TEST_F(SchemaUpgradeTestFixture, KoQDeleteWithDoNotFailFlag)
             <KindOfQuantity typeName="KoQ1" displayLabel="KoQ1" relativeError=".5" persistenceUnit="u:CM" />
         </ECSchema>)schema")));
 
+    auto schema = m_ecdb.Schemas().GetSchema("TestSchema");
```
ColinKerr, line 11024: Should we be changing this test?  It's covering the DoNotFail flag so it should be a case where it should fail.  I know we want to get rid of this flag but we have not yet so we should leave the tests as is.  If this is testing something unique with your modifications then you should create a new test for that case as I expect this test will be removed when we remove that flag

iModelCore/ECDb/Tests/NonPublished/SchemaUpgradeTests.cpp
RohitPtnkr1996: Done !

iModelCore/ECDb/Tests/NonPublished/SchemaUpgradeTests.cpp
RohitPtnkr1996: Yup, makes sense.
I've updated the code and the tests.

ColinKerr (approved):

pmconne (approved):

PR #534, opened 2023-10-27
Fix atexit cleanup function exception

tcobbs-bentley:
_See comment in changed code_.


iModelCore/iModelPlatform/DgnCore/FontManager.cpp
```diff
@@ -393,8 +393,24 @@ bool FontManager::AddWorkspaceDb(Utf8CP inName, CloudContainerP container) {
...
+    // leads to a runtime exception. Doing it this way means that the mutex is "leaked", but since
+    // it is a singleton object, that really doesn't hurt anything.
+    // Note: under normal circumstances, mobile apps never cleanly exit, so this problem is masked.
+    // It's only when something else causes the mobile app to exit (including some other
+    // uncaught exception) that the atexit handler triggers an exception from this mutex.
+    static BeMutex* s_fontsMutex = NULL;
+    if (s_fontsMutex == NULL)
+    {
+        s_fontsMutex = new BeMutex();
+    }
```
pmconne, line 408:
```suggestion
    static BeMutex* s_fontsMutex = new BeMutex();
```

C++11 guarantees that the initialization of a static  local variable executes only once, in a thread-safe manner. It makes no guarantees about thread-safety after that initialization occurs. So the way you've written it permits two (or more) threads to see `s_fontsMutex` as `NULL` and independently assign a `new BeMutex` to it.

tcobbs-bentley: Thanks.

pmconne (approved):

PR #540, opened 2023-11-02
Synced log macro in Units with imodel02

RohitPtnkr1996:
The function code now has access to the LOG macro declared in UnitsNativeLogging.h.
This avoids the need for repeated calls to NativeLogging::CategoryLogger("UnitsNative").

Main motivation for this was to sync imodel-native with imodel02 which contains repeated calls to NativeLogging::LoggingManager::GetLogger().
Identical changes have also been made to imodel02 to make future merges easier between the two repos.


iModelCore/Units/PublicAPI/Units/UnitRegistry.h
```diff
@@ -26,173 +26,10 @@ friend struct Unit;
 
     void AddSystem(UnitSystemR unitSystem); // Used for creating the static definitions.
 
-    PhenomenonP AddPhenomenonInternal(Utf8CP phenomenaName, Utf8CP definition)
```
ColinKerr, line 29: I'd rather leave the implementation in the header ... I'm trying to remember why we did it but I remember we did it for some reason.

ColinKerr (approved):

PR #545, opened 2023-11-07
Point2/3d should save INF/NAN values as NULL

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/965


iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -11036,42 +11036,97 @@ TEST_F(ECSqlStatementTestFixture, verify_inf_and_nan_handling) {
...
+                    break;
+                    }
+                case 2:
+                    {
+                    // Test Point3D properties
+                    double coordX, coordY, coordZ;
+                    stmt.GetValuePoint3d(columnIndex).GetComponents(coordX, coordY, coordZ);
+                    // Check the +inf and -inf column values
+                    EXPECT_EQ(std::isinf(coordX), false) << "Value of co-ordinate X column " << columnIndex + 1 << " should not be infinity";
+                    EXPECT_EQ(std::isinf(coordY), false) << "Value of co-ordinate Y column " << columnIndex + 1 << " should not be infinity";
```
khanaffan, line 11104: You test do not include `Point2d(Inf,1)` & `Point2d(1, Inf)` and then check `IsValueNull()`. Generally Point2d is primitive value `X` and `Y` is either both `NULL` or both not `NULL`. But if you have one component is NULL and not other we return `false` from `IsValueNull()`.

I think the correct behavior is to insert/update to NULL for all value if any value is NULL/INF/NAN. And similarly return IsValueNull() true if any component is Null/Inf/Nan. But we do not have to do it right now.

RohitPtnkr1996: Okay, I'll add a test and check it.

khanaffan (approved):

PR #568, opened 2023-12-07
Fix JS logging leading to a crash on mobile

tcobbs-bentley:
On mobile devices where backend logging is configured to forward to the frontend, doing a JS log too early in the startup process results in an unhandled exception due to the fact that the frontend and backend are not yet connected. Catch that exception when it happens and store the failed log message in the list of deferred log messages. Also process the list of deferred log messages any time IsSeverityEnabled is called so that you don't have to wait for an enabled severity message to be logged before the deferred ones get processed.

### DIFF
--- a/iModelJsNodeAddon/JsLogger.cpp
+++ b/iModelJsNodeAddon/JsLogger.cpp
@@ -143,10 +143,6 @@ void JsLogger::LogMessage(Utf8CP category, SEVERITY sev, Utf8CP msg) {
 
 bool JsLogger::IsSeverityEnabled(Utf8CP category, SEVERITY sev) {
     BeMutexHolder lock(m_deferredLogMutex);
-    if (canUseJavaScript()) {
-        // Process the deferred log messages every chance we get.
-        processDeferred();
-    }
     return isEnabled(category, sev);
 }
 
--- a/iModelJsNodeAddon/JsLogger.cpp
+++ b/iModelJsNodeAddon/JsLogger.cpp
@@ -102,9 +102,12 @@ void JsLogger::processDeferred() {
         try {
             logToJs(message.m_category.c_str(), message.m_severity, message.m_message.c_str());
         } catch (...) {
-            // Logging to JS did not work (probably due to backend logging being redirected to the
-            // frontend and the backend and frontend not currently being connected). Try again
+            // Logging to JS did not work, so leave m_deferredLogging alone so we can try again
             // later.
+            // logToJs triggers the backend JavaScript logging functionality. That functionality
+            // allows for custom log functions. Those custom log functions may throw an exception,
+            // and if they do so, that exception gets propogated to here. Catch that here so that
+            // we don't throw a C++ exception from this function.
             return;
         }
     }
@@ -128,9 +131,10 @@ void JsLogger::LogMessage(Utf8CP category, SEVERITY sev, Utf8CP msg) {
             return;
         } catch (...) {
             // Push to deferred below, because JS log failed.
-            // On mobile, it seems to fail if called when backend logging is redirected to the
-            // frontend, and the backend and frontend are not connected at the time of the log
-            // message.
+            // logToJs triggers the backend JavaScript logging functionality. That functionality
+            // allows for custom log functions. Those custom log functions may throw an exception,
+            // and if they do so, that exception gets propogated to here. Catch that here so that
+            // we don't throw a C++ exception from this function.
         }
     }
     // save this message in memory so it can be logged later on the JavaScript thread
--- a/iModelJsNodeAddon/JsLogger.cpp
+++ b/iModelJsNodeAddon/JsLogger.cpp
@@ -98,8 +98,16 @@ bool JsLogger::isEnabled(Utf8CP catIn, SEVERITY sev) {
 void JsLogger::processDeferred() {
     if (m_deferredLogging.empty())
         return;
-    for (auto const& message : m_deferredLogging)
-        logToJs(message.m_category.c_str(), message.m_severity, message.m_message.c_str());
+    for (auto const& message : m_deferredLogging) {
+        try {
+            logToJs(message.m_category.c_str(), message.m_severity, message.m_message.c_str());
+        } catch (...) {
+            // Logging to JS did not work (probably due to backend logging being redirected to the
+            // frontend and the backend and frontend not currently being connected). Try again
+            // later.
+            return;
+        }
+    }
     m_deferredLogging.clear();
 }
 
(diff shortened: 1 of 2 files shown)


tcobbs-bentley:
> > results in an unhandled exception
> 
> Please provide more details about this exception.
> 
> Can you please educate us on how we end up in a situation where we are attempting to log while the "frontend and backend are not yet connected" and if maybe that situation can be avoided in the first place. e.g., does the frontend notify the backend that is has become "connected" at some point (or vice versa)?

The exact exception text is:

> Cannot read properties of undefined (reading 'send')

Since the mobile sample's backend redirects logging to the frontend (using `IpcHost.send`), and we don't see this exception in other places, I came to the conclusion that the 'send' mentioned in the exception was from IpcHost while the logging is trying to forward the log message to the frontend. Due to the nature of the bug, there's no real way to debug, since the actual exception happens in the backend JavaScript, and this happens long before I can attach the Chrome debugger to that.

Note also that while this happens in the mobile sample, the details of what actually happens don't matter. Here,the actual code is sample code. But Logger.initialize has four custom function parameters for logging, which means that user-supplied custom code can be run as a result of the JsLogger.cpp code. We can't assume that code won't throw an exception (which we did before my change here). I'll update the comments to reflect this once the initial CI build completes.

tcobbs-bentley:
I committed the wrong file. I will roll back those changes and commit the right file.

tcobbs-bentley:
OK, fixed. I reverted the accidental changes, and updated the comments.

iModelJsNodeAddon/JsLogger.cpp
```diff
@@ -115,15 +126,27 @@ void JsLogger::LogMessage(Utf8CP category, SEVERITY sev, Utf8CP msg) {
...
+            // we don't throw a C++ exception from this function.
+        }
     }
+    // save this message in memory so it can be logged later on the JavaScript thread
+    m_deferredLogging.push_back(LoggedMessage(category, sev, msg));
 }
 
 bool JsLogger::IsSeverityEnabled(Utf8CP category, SEVERITY sev) {
     BeMutexHolder lock(m_deferredLogMutex);
+    if (canUseJavaScript()) {
```
kabentley, line 146: please remove this change. Why did you add it?

tcobbs-bentley: I added it because deferred error messages were basically never getting sent, since the only way for them to be sent was for another _error_ to be logged. With this change, they get sent the next time any severity is logged (even ones that are disabled). I can remove the change if you really want, but I feel that it is an improvement over what was there before.

kabentley: Yes, _please_ remove this. `IsSeverityEnabled` is called way more often than actual logging (in fact, it's often called multiple times for the same log site) and should not have any side effects. Differed messages are deferred until another log request (any log requests, not errors) happens on the main thread. That is not an urgent situation.

kabentley: Yes, log messages are deferred until another **enabled** message is sent on the main thread - that doesn't have to be an error. It is true that it's indeterminate when that may happen. That's why the `flushLog` method exists.

kabentley:
These changes are fine by me. 

But I still don't understand what we're doing for logging in mobile. Why are backend messages sent to the frontend? Do we somehow send all log message over the wire (to where/when)? Do we ask permission to do that?

iModelJsNodeAddon/JsLogger.cpp
kabentley: no, it can only be called from JavaScript, so `canUseJavaScript` is not necessary.

kabentley:
> This is just something the mobile samples do. And while it could be argued that they are broken for allowing exceptions to escape their logging function, the fact is that other end users may have backend log functions that throw exceptions. Also, on mobile, the backend and frontend are always running on the same device, so while there is web socket traffic, it is guaranteed to be local.

I'm not asking about exceptions. I don't understand why we send backend log messages to the frontend at all.

tcobbs-bentley:
> I'm not asking about exceptions. I don't understand why we send backend log messages to the frontend at all.

For the mobile samples, it means that all logging is handled the same way. Without this, backend log messages don't show up at all.

kabentley:
> For the mobile samples, it means that all logging is handled the same way. Without this, backend log messages don't show up at all.

What do you mean "show up". Where do they show up? What do we do in production?

tcobbs-bentley:
> What do you mean "show up". Where do they show up? What do we do in production?

For the mobile samples, they simply get printed in the debug console. There is no production for the mobile samples.

For the existing production app using iTwin Mobile SDK (SYNCHRO Field), the frontend sends messages to their native code so that they end up using the logging mechanism used by the rest of their app. I think they forward the backend log messages to the frontend, but I'm not 100% sure of that.

pmconne (commented):
If an exception occurs when trying to log a message, you add the message we attempted to log onto the deferred message queue, hoping to be able to successfully log it later.
Your changes assume the failure to log is temporary and will be resolved shortly, which is hopefully true in the mobile case.
If you're trying to address the general problem of "what happens if a logger function throws an exception", you would want to 1. add an additional deferred message to the log describing the exception that was thrown by the logger and 2. avoid enqueuing an unbounded number of deferred messages.

I don't think it's worth trying to solve the general problem. Do you?

tcobbs-bentley:
> If an exception occurs when trying to log a message, you add the message we attempted to log onto the deferred message queue, hoping to be able to successfully log it later. Your changes assume the failure to log is temporary and will be resolved shortly, which is hopefully true in the mobile case.
> 
> If you're trying to address the general problem of "what happens if a logger function throws an exception", you would want to 1. add an additional deferred message to the log describing the exception that was thrown by the logger and 2. avoid enqueuing an unbounded number of deferred messages.
> 
> I don't think it's worth trying to solve the general problem. Do you?

I don't really understand what you are asking here. Are you saying that this whole PR should just be closed? It's trivial for me to update the mobile sample not to throw an exception from the log function, but given that doing so led to a problem, I felt it was more appropriate to fix the root cause, which is that the native code freaks out if the JS log functions throw an exception.

I can add a catch for Napi::Error that adds the JS exception to the list of logged errors, and I can have it just give up if the deferred list gets too long, but I don't feel that either of those things is as useful as the simple changes that I already have in this PR.

pmconne:
> It's trivial for me to update the mobile sample not to throw an exception from the log function, but given that doing so led to a problem, I felt it was more appropriate to fix the root cause, which is that the native code freaks out if the JS log functions throw an exception.

Your changes are based on the assumption that the logger-produced exceptions are temporary and ignore-able. They do not address (instead only mask) the problem of, e.g., a logger function that throws every time it is called. I don't care about solving that problem, so I'm fine with your changes. But if you do care about that, then your changes don't suffice.

tcobbs-bentley:
> Your changes are based on the assumption that the logger-produced exceptions are temporary and ignore-able. They do not address (instead only mask) the problem of, e.g., a logger function that throws every time it is called. I don't care about solving that problem, so I'm fine with your changes. But if you do care about that, then your changes don't suffice.

I don't care about a logger function that throws every time it is called.

pmconne (approved):

tcobbs-bentley:
@Mergifyio backport release/4.x

aruniverse:
@Mergifyio backport release/4.x

PR #569, opened 2023-12-08
Add support for MemberPriority in bis:ElementRefersToElement and extend the unique index

RohitPtnkr1996:
Refers to https://github.com/iTwin/itwinjs-backlog/issues/840.

We want to add the MemberPriority property to bis:ElementRefersToElements and extend the unique index to include this new property.
This will allow us to have duplicate instances of the subclasses of bis:ElementRefersToElements with member priority values being different.
This PR contains the support to handle the soon incoming BisCore changes along with a ECDb profile version increment to 4.0.0.5.

The change to BisCore is will be rolled out soon. This new version of BisCore can only be imported when runtime ECDb version is at least 4.0.0.5


iModelCore/ECDb/CHANGES.md
```diff
@@ -4,9 +4,16 @@ This document including important changes to syntax or file format.
...
 | Module  | Version   |
 | ------- | --------- |
-| Profile | `4.0.0.4` |
+| Profile | `4.0.0.5` |
 | ECSQL   | `1.2.9.2` |
 
+## `12/08/2023`: Modified unique index in bis:ElementRefersToElements to include new MemberPriority property
+
+* ECDb profile version updated to `4.0.0.4` -> `4.0.0.5`.
+* BisCore domain schema has been upgraded to 1.0.17 which adds MemberPriority property in the class bis:ElementRefersToElement.
```
ColinKerr, line 13: This profile update should not be tied to a new version of BisCore or any classes or properties from BisCore.

The profile update must add new mapping CAs that we can add to BisCore so the new db mapping is implemented.

ECDb cannot have any knowledge of BisCore.

iModelCore/ECDb/ECDb/RelationshipClassMap.cpp
```diff
@@ -1001,6 +1001,36 @@ ClassMappingStatus RelationshipClassLinkTableMap::_Map(ClassMappingContext& ctx)
...
+
+            if (BE_SQLITE_OK != ctx.GetImportCtx().GetECDb().ExecuteDdl("DROP INDEX IF EXISTS [uix_bis_ElementRefersToElements_sourcetargetclassid]"))
+                return ClassMappingStatus::Error;
+            }
+
+        // Add the new index introduced in Profile version 4.0.0.5 if not already present
+        if (!GetPrimaryTable().ContainsIndex("uix_bis_ElementRefersToElements_sourcetargetclassid_memberpriority"))
+            AddIndex(ctx.GetImportCtx(), RelationshipIndexSpec::SourceAndTargetAndClassId, true);
+        }
+    return ClassMappingStatus::Success;
```
ColinKerr, line 1031: This logic cannot exist in ECDb.  ECDb must look at the mapping CAs applied to BisCore.ElementRefersToElements to determine that it should change the index

iModelCore/ECDb/ECDb/RelationshipClassMap.cpp
```diff
@@ -1255,8 +1285,20 @@ void RelationshipClassLinkTableMap::AddIndex(SchemaImportContext& schemaImportCo
...
+                
+                // ECDb Profile version 4.0.0.5 onwards, update the unique index `uix_bis_ElementRefersToElements_sourcetargetclassid` to `uix_bis_ElementRefersToElements_sourcetargetclassid_memberpriority`
+                if (name.Equals("uix_bis_ElementRefersToElements_sourcetargetclassid"))
+                    {
+                    if (auto memberPriorityColumn = GetPrimaryTable().FindColumn("MemberPriority"); memberPriorityColumn)
+                        {
+                        columns.push_back(memberPriorityColumn);
+                        name.append("_memberpriority");
+                        }
+                    }
```
ColinKerr, line 1299: Same here the index should be driven by the mapping CAs applied to the schema.  Doing this not only decouples the change from BisCore but it allows you to write a synthetic test in ECDb that doesn't use BisCore

iModelCore/ECDb/ECDb/RelationshipClassMap.cpp
khanaffan: I second that, having a conditional code on the content of incoming data is a red flag for the maintainer. You need to remove this.

khanaffan:
A general assumption in PR is that `BisCore` does not remap when a referenced schema is imported. If you import even unrelated schema all indexes will be verified and if they do not match what the current code thing should be they are dropped and recreated. 

If you change default indexes in code then as I mentioned before older code will recreate them differently and new cost will recreate them differently. Given new index enforces uniqueness what happens when there is existing duplicate data and imodel is upgraded? The schema import will fail, right? (you do not have a unit test for this case)

One other thing you missing is having a test on the right layer which is imodel platform with actual biscore schema and verifying the newly created imodel and an upgraded imodel.

To me, it looks it's a write version change. We need to discuss this.

PR #570, opened 2023-12-08
don't segfault, return nullptr if `fkView` was null during change iteration

MichaelBelousov:
short term fix for https://github.com/iTwin/itwinjs-backlog/issues/997,

> The crash happened because the change summary attempted to locate classmap for the foreign key relationship but the RelECClassId represents a link table relationship and it returned null.
>
> It should be noted all these invalid relationship seem to be deleted or fixed as they do not appear in tip level imodel.

[full explanation of what happened](https://github.com/iTwin/itwinjs-backlog/issues/997#issuecomment-1845410536)


MichaelBelousov:
closing affan is taking this over

PR #571, opened 2023-12-08
Fix crash in ChangeSummary where RelECClassId was valid but not EndTableRelationship

khanaffan:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/997

This fix defaults to the base relationship for a given navigation property if `RelECClassId` in change is either an invalid ECClass or valid ECClass which is not mapped to `EndTableRelationshipMap`.


MichaelBelousov (approved):

rschili (approved):

aruniverse:

PR #577, opened 2023-12-13
Optimize schema load performance having CustomAttribute from pruned schemas

akshay05n:
This pull request addresses a performance bottleneck in schema loading. The code now checks if a schema, defining custom attributes, is in the pruned schemas list before attempting to locate them.

### DIFF
--- a/iModelCore/ecobjects/src/ECCustomAttribute.cpp
+++ b/iModelCore/ecobjects/src/ECCustomAttribute.cpp
@@ -459,9 +459,11 @@ CustomAttributeReadStatus IECCustomAttributeContainer::ReadCustomAttributes (pug
                 // In EC3 we will fail to load the schema if any invalid custom attributes are found, for EC2 schemas we will skip the invalid attributes and continue to load the schema
                 if (containerSchema.OriginalECXmlVersionAtLeast(ECVersion::V3_0))
                     {
+                    SchemaKey caSchemaKey;
+                    if (ECObjectsStatus::Success != SchemaKey::ParseSchemaFullName(caSchemaKey, ns.c_str()))
+                        return CustomAttributeReadStatus::InvalidCustomAttributes;
                     // In EC3, skip the custom attribute if it belongs to a pruned schema
-                    Utf8String caClassName = ns.substr(0, ns.find('.'));
-                    if (schemaContext.GetSchemasToPrune().end() != std::find(schemaContext.GetSchemasToPrune().begin(), schemaContext.GetSchemasToPrune().end(), caClassName.c_str()))
+                    if (schemaContext.GetSchemasToPrune().end() != std::find(schemaContext.GetSchemasToPrune().begin(), schemaContext.GetSchemasToPrune().end(), caSchemaKey.GetName().c_str()))
                         status = CustomAttributeReadStatus::SkippedCustomAttributes;
                     else
                         status = CustomAttributeReadStatus::InvalidCustomAttributes;
--- a/iModelCore/ecobjects/src/ECCustomAttribute.cpp
+++ b/iModelCore/ecobjects/src/ECCustomAttribute.cpp
@@ -458,7 +458,14 @@ CustomAttributeReadStatus IECCustomAttributeContainer::ReadCustomAttributes (pug
                 {
                 // In EC3 we will fail to load the schema if any invalid custom attributes are found, for EC2 schemas we will skip the invalid attributes and continue to load the schema
                 if (containerSchema.OriginalECXmlVersionAtLeast(ECVersion::V3_0))
-                    status = CustomAttributeReadStatus::InvalidCustomAttributes;
+                    {
+                    // In EC3, skip the custom attribute if it belongs to a pruned schema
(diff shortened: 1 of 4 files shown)


iModelCore/ecobjects/src/ECCustomAttribute.cpp
```diff
@@ -458,7 +458,14 @@ CustomAttributeReadStatus IECCustomAttributeContainer::ReadCustomAttributes (pug
                 {
                 // In EC3 we will fail to load the schema if any invalid custom attributes are found, for EC2 schemas we will skip the invalid attributes and continue to load the schema
                 if (containerSchema.OriginalECXmlVersionAtLeast(ECVersion::V3_0))
-                    status = CustomAttributeReadStatus::InvalidCustomAttributes;
+                    {
+                    // In EC3, skip the custom attribute if it belongs to a pruned schema
+                    Utf8String caClassName = ns.substr(0, ns.find('.'));
```
rschili, line 463: This should be caSchemaName, not caClassName.
Also, substring is probably not a good choice here. Check how ECInstance.cpp does it around line 2600, calling SchemaKey::ParseSchemaFullName.
Looking at ECInstance.cpp line 2675 it seems we also support the ECXML_URI here, though I have never seen that in real life. The xmlns can also be missing completely if it's a local custom attribute. Try and make this check a bit more robust so it doesn't try and look-up schemas if the string format is not valid.

iModelCore/ecobjects/src/ECcontext.cpp
```diff
@@ -399,6 +399,12 @@ struct CustomAttributeInstanceReadContext: public ECInstanceReadContext
         if (it != references.end())
             return it->second.get();
 
+        if (m_schemaContext.GetSchemasToPrune().end() != std::find(m_schemaContext.GetSchemasToPrune().begin(), m_schemaContext.GetSchemasToPrune().end(), key.GetName()))
+            {
+            LOG.errorv("CustomAttributeInstanceReadContext - Custom attribute schema %s is one of the schemas that are to be pruned. Skipping fallback mechanism.", key.GetFullSchemaName().c_str());
```
rschili, line 404: In this case, it's not an error, log as info or warning

akshay05n: Changed to log it as the warning

iModelCore/ecobjects/src/ECCustomAttribute.cpp
akshay05n: Using `SchemaKey::ParseSchemaFullName()` to parse the schema name similar to how ECInstance.cpp does on line 2606 instead of `substring`. Also, we're validating the full schema name format on line 448 in ECCustomAttribute.cpp

iModelCore/ecobjects/src/ECcontext.cpp
rschili: I still think finding hundreds of these warnings in their logs will confuse the connector people a lot.
Better make it clear that this is nothing that needs fixing, it is intended behavior. Somebody requested the schema to be pruned, so we are pruning a custom attribute which uses the schema.
Do not mention the fallback mechanism here. In the error message further down this is necessary because it is an exceptional case where we could not find a CA schema in the referenced schemas, so we are calling locateSchema as a fallback, ideally people would fix their schema by adding a reference.
But in this case with a pruned schema, fallback is not necessary or intended, and the schema is just fine.
I'd say just log an info saying "Skipping loading of a custom attribute because its schema (%s) is being pruned"

rschili (approved):

iModelCore/ecobjects/test/Performance/PerformanceSchemaDeserializationTests.cpp
```diff
@@ -0,0 +1,102 @@
...
+Utf8String GenerateMainSchema(size_t numberOfCAClasses)
+    {
+    Utf8PrintfString schemaName("TestSchema_%zuCA", numberOfCAClasses);
+    Utf8PrintfString refSchemaName("RefSchema_%zuCA", numberOfCAClasses);
+    Utf8PrintfString description("Test schema with %zu Custom Attribute Classes", numberOfCAClasses);
+    Utf8String innerXml = "";
+    for (size_t i = 0; i < numberOfCAClasses; i++) 
+        {
+        Utf8PrintfString caClassName("CustomAttrClass%zu", i);
+        innerXml = innerXml + "<ECCustomAttributes>\n" + "<" + caClassName + " xmlns=\"" + refSchemaName + ".01.01.01\" />\n" + "</ECCustomAttributes>\n";
```
ColinKerr, line 73: There should only be one ECCustomAttributes node on each custom attribute container.  ECSchema will have one ECCustomAttributes node and that node will have all the custom attribute instances as children.  So you should put the ECCustomAttributes node in the fixed part of the schema and just write put the custom attributes here

akshay05n: Updated test so that schema has only 1 `ECCustomAttributes` node.

ColinKerr (approved):

PR #581, opened 2023-12-20
Addressing overuse of INFO messages in ecobjects

akshay05n:


iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -3333,7 +3333,7 @@ SchemaReadStatus ECSchema::ReadFromXmlFile(ECSchemaPtr& schemaOut, WCharCP ecSch
         {
         //We have serialized a schema and its valid.
         timer.Stop();
-        LOG.infov (L"Read (in %.4f seconds) [%3" PRIu64 " ECClasses] %ls", timer.GetElapsedSeconds(), (uint64_t) schemaOut->m_classMap.size(), ecSchemaXmlFile);
+        LOG.tracev (L"Read (in %.4f seconds) [%3" PRIu64 " ECClasses] %ls", timer.GetElapsedSeconds(), (uint64_t) schemaOut->m_classMap.size(), ecSchemaXmlFile);
```
ColinKerr, line 3336: @c-macdonald  do you ever use this timer result?  If not then I think we can just remove the log message and the StopWatch

ColinKerr: Carole does not use the timers, we can remove them

akshay05n: Removed timer log messages

ColinKerr (approved):

PR #591, opened 2024-01-19
Add support for row values commalist

akshay05n:
This pull request enables support for row values constructor lists in ECSQL. Previously, the query `SELECT * FROM (VALUES(1,2), (2,3));` would result in a parsing error (`syntax error`).

With these changes, the ECSQL statement is now prepared successfully: `SELECT [1], [2] FROM VALUES (1, 2) UNION ALL VALUES (2, 3)`. The corresponding SQLite SQL has been optimised to: `SELECT [K0],[K1] FROM (SELECT 1 [K0],2 [K1] UNION ALL SELECT 2,3)`


iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -169,6 +169,16 @@ TEST_F(ECSqlStatementTestFixture, NestedValConstructor) {
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
+TEST_F(ECSqlStatementTestFixture, RowValConstructor) {
+    ASSERT_EQ(DbResult::BE_SQLITE_OK, SetupECDb("RowValConstructor.ecdb"));
+
+    ECSqlStatement stmt;
+    ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb,"SELECT * FROM (VALUES(1,2), VALUES(3,4))"));
```
rschili, line 176: This query in plain sqlite will not parse, instead, the sqlite syntax would be
```sql
SELECT * FROM (VALUES(1,2), (3,4))
```

Also, the first query returns row (1,2) first and then row(3,4) while the native sql below returns the reversed order. (3,4 first, then 1,2)

akshay05n: I've fixed the order of returned values in the generated native SQL.

akshay05n: Updated grammar parser rule to support `SELECT * FROM (VALUES(1,2), (3,4))` instead of `SELECT * FROM (VALUES(1,2), VALUES(3,4))`

rschili (approved):

khanaffan (approved):

PR #595, opened 2024-01-24
Add support to extend system-generated indexes [buildNum:2024.02.26.1]

RohitPtnkr1996:
Refers to https://github.com/iTwin/itwinjs-backlog/issues/840.

We want to add the "MemberPriority" property to bis:ElementRefersToElements and extend the unique index to include this new property.
This will allow us to have duplicate instances of the subclasses of bis:ElementRefersToElements with member priority values being different.
This PR contains the support to handle the soon incoming BisCore changes mentioned above.


khanaffan (approved):

iModelCore/ECDb/ECDb/DbMappingManager.cpp
```diff
@@ -698,47 +698,83 @@ BentleyStatus DbMappingManager::Classes::MapUserDefinedIndexes(SchemaImportConte
...
 
-        if (SUCCESS != MapUserDefinedIndex(importCtx, classMap, indexCA))
+        if (SUCCESS != MapIndex(importCtx, classMap, indexCA, isSystemIndex))
             return ERROR;
+
+        // Special case: Since older imodels ( <= 4.0.0.2) have individual unique indexes for the derived classes as well, just updating the one for the base class is not enough.
+        // As we have already updated the unique index in the base class to handle the uniqueness in the derived classes as well, we can now remove the unique indexes in the derived classes.
+        if (checkDerivedClassIndexes && isSystemIndex && indexCA.IsUnique())
+            {
+            if (const auto allDerivedClasses = importCtx.GetECDb().Schemas().GetAllDerivedClassesInternal(classMap.GetClass()); allDerivedClasses.IsValid())
```
khanaffan, line 760: Nested IFs make code less readable. You can use a short circuit pattern where you `continue`, `break` or `return` when a rule is false. This avoids nesting IFs and thus intending code.

Instead of following

```c
while() {
  if (a) {
    if (b) {
     // work
    }
  }
}
```
use
```c
while () {
  if (!a) continue;
  if (!b) continue;
  // work
}
```

pmconne (approved):

PR #598, opened 2024-02-01
Fix for audit warning 150

naveedkhan8067:
In order to handle the warning 150 from the audit report following changes are performed:
- setup DLM_EXTENSION  to `.node`
- App name has made similar to the file we deliver `imodeljs`

The `warning 150` complains about the absence of `.node` file at the server.

imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/378


naveedkhan8067:
`imodeljs.node` will now symlinked to `imodeljs-addon.node` and the PDB file is `imodeljs-addon.pdb`. @chuckkir shouldn't these two files be named as `imodeljs` ?

chuckkir:
But it seems like it shouldn't be renaming the file in the link. imodeljs.node should be good throughout, unless there's a reason for the rename?

imodeljs.node [D:\imodelNative\out\Winx64\build\imodeljs-addon\imodeljs-addon.node]

naveedkhan8067:
> But it seems like it shouldn't be renaming the file in the link. imodeljs.node should be good throughout, unless there's a reason for the rename?
> 
> imodeljs.node [D:\imodelNative\out\Winx64\build\imodeljs-addon\imodeljs-addon.node]

By rename, I meant the extension of the linked file `imodeljs-addon` changed from `.dll` to `.node`. I am not sure why they have kept this setting where the file being delivered is named as `imodeljs` but the linked file (dll) and the pdb are named differently `imodeljs-addon`. May be we can ask them if it is not required or necessary? then we can update them to have same names (of course excluding extensions) like the other linked files. Probably for that we just have to update `appName = imodeljs-addon` to `appName = imodeljs`.

chuckkir (approved):

pmconne (approved):
> warning 150 from the audit report

That doesn't mean anything to me but deferring to @chuckkir.

chuckkir (approved):

PR #609, opened 2024-02-08
[Experimental] Related instance API allow traversal of instance in forward/backward/both directions

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/8670

**[Experimental]** require specifying `OPTIONS ENABLE_EXPERIMENTAL_FEATURES`

The API allows the generically traverse of a graph of instances using relationships

The API introduces a virtual class that takes an instance key and direction and will return a table of related instances along with relationship ID and direction.

```sql
rel1.related_instances(<instanceid>, <eclassid>, 'forward' | 'backward' | 'both')
```

## Propose ECSQL syntax

1. `related_forward($)`
2. `related_backward($)`
3. `related_both($)`

example

```sql
SELECT * FROM Bis.Element e, realted_forward(e.$) r 
```
 
Related instances use internally cached ECSQL statements to give you related instances and can be slow at first when these statements get prepared. The performance also varies depending on the number of relationships for a given instance key.

It can be used in conjunction with other classes in ECSQL as follows.

```sql
 SELECT
    Bis.Element.ECInstanceId, 
    g.* 
 FROM
     Bis.Element e, 
     rel1.related_instances(Bis.Element.ECInstanceId, Bis.Element.ECClassId) g
 WHERE Bis.Element.ECINstanceId = 0x30000001ddf OPTIONS ENABLE_EXPERIMENTAL_FEATURES;
```
 Table

|`ECInstanceId` |`ECInstanceId`   |`ECClassId`  |`RelECClassId`    |`Direction`|         
|-----------------|-----------------|-----|---|-------------|
|0x30000001ddf|1                        |244|91 |1|
|0x30000001ddf|3298534891276|252|160|1|
|0x30000001ddf|3298534890841|211|164|1|

### A related fix is to query fk relationship by its class name

The fix here is that if `RelECClassId` is null then ECSQL will return root `ECClassId` for the relationship to make this behavior consistent with navigation property


rschili:
Hey, this looks very interesting, and I have some high level design comments, which I would have liked to discuss before the implementation is done :( Did I miss a meeting where this was discussed? Now it feels too late to make comments, but here are them anyways...

> example
> 
> ```sql
> SELECT * FROM Bis.Element e, realted_forward(e.$) r 
> ```

Using an alias introduced in the same FROM clause (e here) isn’t something SQL normally allows — the alias isn’t fully in scope yet while the FROM clause is being built.
Typically, the alias becomes available only later (in the SELECT, WHERE, ORDER BY, etc.). So this pattern would require a custom parser behavior that diverges from standard SQL semantics.

If we want something that feels more object-graph–like, we could make it look like a call on the instance itself, e.g.:

```sql
SELECT * FROM Bis.Element e, e->realted_forward() r 
```

PostgreSQL’s LATERAL JOIN provides a clean, standards-aligned way to express this idea — it lets you reference earlier tables in the same FROM clause.
That would look like:

```sql
SELECT e.id, r.*
  FROM Element e
  JOIN LATERAL related_forward(e.id) AS r ON TRUE
```

This keeps SQL semantics intact and avoids parser complications.

> It can be used in conjunction with other classes in ECSQL as follows.
> 
> ```sql
>  SELECT
>     Bis.Element.ECInstanceId, 
>     g.* 
>  FROM
>      Bis.Element e, 
>      rel1.related_instances(Bis.Element.ECInstanceId, Bis.Element.ECClassId) g
>  WHERE Bis.Element.ECINstanceId = 0x30000001ddf OPTIONS ENABLE_EXPERIMENTAL_FEATURES;
> ```

Personally, I’d love something like this, which combines the clarity of LATERAL joins with the readability of the $ dereference syntax we already use:

```sql
SELECT e.ECInstanceId, r.$->Length
FROM Element e
JOIN LATERAL related_forward(e.ECInstanceId, e.ECClassId) AS r
WHERE e.ECInstanceId = 0x30000001ddf
  AND r.$->Length > 50;
```

This would return all related instances having Length > 50 for the given element.
It’s intuitive, stays close to SQL, and doesn’t introduce ambiguous scoping.

aruniverse:
@khanaffan , Is this still needed with https://github.com/iTwin/imodel-native/pull/1400 ?  If not, lets close this PR

khanaffan:
> @khanaffan , Is this still needed with #1400 ? If not, lets close this PR

No. I will close it.

khanaffan:
Closing as this feature has been implemented in another PR https://github.com/iTwin/imodel-native/pull/1400

Issue #620, opened 2024-02-15
Presentation: Models tree duplicates Subject nodes when requesting with hierarchy level size limit


grigasp:
Repro:

1. Start VCR test-viewer with Models tree config:
   ```tsx
   {
     id: ModelsTreeComponent.id,
     getLabel: ModelsTreeComponent.getLabel,
     render: () => (
       <ModelsTreeComponent
         selectionPredicate={() => true}
         selectionMode={SelectionMode.Extended}
         hierarchyLevelConfig={{ isFilteringEnabled: true, sizeLimit: 100 }}
       />
     ),
   },
   ```

2. Open http://localhost:3000/?iTwinId=8b3d0653-02b5-4a76-832e-7c16ed50ff21&iModelId=69d20703-222b-46da-af46-891a5c6de153

3. Open the Models tree, expand the `ExtonMaster` Subject node.
4. Refresh the web page.
5. Expand the `ExtonMaster` node again. Result:
   ![Image](https://github.com/iTwin/imodel-native/assets/35135765/881187ff-c15e-448f-a03b-cd722cdf0c24)

PR #621, opened 2024-02-15
Presentation: Fix nodes getting duplicated when using hierarchy level limiting with same label instance grouping under hidden nodes

grigasp:
Fixes https://github.com/iTwin/imodel-native/issues/620.

In addition, there was a problem that same-label grouping nodes were returned in wrong order when requesting in pages.


saskliutas (approved):

grigasp:

PR #644, opened 2024-02-28
Make SchemaMerger more robust

6ar8nas:
## Changes

Solves the issue of OB Connector crashing when a new merger is used, as the schemas had wrong versions. This PR allows to bypass the unwanted behavior by setting a `DoNotMergeReferences` flag. The flag enables not inserting missing schema references and copying source references instead of locating it. Additionally, schema lists validation and null checks were added for the merge operation to make it more robust and more descriptive in case of an error.

### Miscellaneous

- Unified `CompareIUtf8Ascii` across the solution to be only in one place, rather than scattered across different files.
- Fixed `DbSchema.cpp`.`PrimaryKeyDbConstraint::Create` which had only partial uniqueColNames logic present.
- Updated `SchemaMerger.MergeSchemas` method documentation.
- Unified `BeStringUtilities` test fixtures into one file.
- Fixed an issue where SchemaMerger would treat different cased schema name strings as different schemas.
- Fixed an issue where SchemaMerger would sometimes merge schema versions wrong depending on schema names contents.

Closes iTwin/itwinjs-backlog#1038


iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -218,8 +236,30 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
     //Make a copy of the input vectors so we don't modify the original (given to us as const anyways)
     bvector<ECSchemaCP> left(rawLeft);
     bvector<ECSchemaCP> right(rawRight);
-    ECSchema::SortSchemasInDependencyOrder(left);
-    ECSchema::SortSchemasInDependencyOrder(right);
+    auto doNotMergeReferences = options.DoNotMergeReferences();
+    if(!doNotMergeReferences)
+        {
+        ECSchema::SortSchemasInDependencyOrder(left);
```
rschili, line 242: I believe we should also sort the provided schemas even if the flag is false.
Better add a new boolean parameter to SortSchemasInDependencyOrder which gets passes along to BuildDependencyOrderedSchemaList() and indicates whether referenced schemas should be included in that list or not

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -218,8 +236,30 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
...
+        ECSchema::SortSchemasInDependencyOrder(right);
+        }
+    else
+        {
+        bool isValid = true;
+        if(SchemaMerger::ValidateUniqueSchemaNames(left) != BentleyStatus::SUCCESS)
+            {
+            result.Issues().ReportV(IssueSeverity::Fatal, IssueCategory::BusinessProperties, IssueType::ECSchema, ECIssueId::EC_0060,
+                "At least a single existing schema name was non-unique.");
```
rschili, line 251: This can probably be improved, the message as it stands will not be very useful for troubleshooting. I'd say make it log the side (left or right) and also make it mention the name of the schema which was non unique. (ValidateUniqueSchemaNames could return an optional string, or it could directly log any duplicate schema names if there are multiple) That way the callers can much more easily track the issue down.
Then you do not even need the isValid flag, just call Validate for both sides in a single line, the method itself logs any violations and if one of the two calls fails, you return error here.

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -277,6 +317,13 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
     if(failedToFillSchemas)
         return BentleyStatus::ERROR;
 
+     if(SchemaMerger::ValidateUniqueSchemaNames(result.GetResults()) != BentleyStatus::SUCCESS)
```
rschili, line 320: I do not think this should ever happen, or can you hit it with a test?

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -277,6 +317,13 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
     if(failedToFillSchemas)
         return BentleyStatus::ERROR;
 
+     if(SchemaMerger::ValidateUniqueSchemaNames(result.GetResults()) != BentleyStatus::SUCCESS)
+        {
+        result.Issues().ReportV(IssueSeverity::Fatal, IssueCategory::BusinessProperties, IssueType::ECSchema, ECIssueId::EC_0060,
```
rschili, line 322: The ECIssueIds should be unique for every log message call. So you will have to add more and not just reuse 0060 for multiple similar issues

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -343,6 +390,13 @@ BentleyStatus SchemaMerger::MergeItems(SchemaMergeResult& result, ECSchemaP left
         if (opCode == ECChange::OpCode::New)
             { //This may actually work
             auto newSchemaItem = (right->*getItemCP)(itemName);
+            if (newSchemaItem == nullptr)
+                {
+                result.Issues().ReportV(IssueSeverity::Fatal, IssueCategory::BusinessProperties, IssueType::ECSchema, ECIssueId::EC_0059,
+                    "The schema is dirty. Another reference of the item with name %s already exists in the schema.", right->GetFullSchemaName().c_str());
```
rschili, line 396: I do not think callers would understand this log message.
I suggest something a long the lines of "Failed to find item with name %s in right schema with name %s. This usually indicates a dirty schema graph where multiple memory references of the same schema with different contents are provided"

iModelCore/ecobjects/test/NonPublished/SchemaMergerTests.cpp
```diff
@@ -40,6 +40,21 @@ ECSchemaReadContextPtr InitializeReadContextWithAllSchemas(bvector<Utf8CP> const
     return readContext;
     }
 
+ECSchemaReadContextPtr InitializeReadContextWithAllSchemas(bvector<BeFileName> const& fileNames, bvector<ECSchemaCP>* loadedSchemas = nullptr)
```
rschili, line 43: Remove or comment this as it's not intended for the final code

iModelCore/ecobjects/test/NonPublished/SchemaMergerTests.cpp
```diff
@@ -5504,55 +5519,442 @@ TEST_F(SchemaMergerTests, TestBaseClassAdditionAndRemoval)
...
+        )schema"
+    };
+    ECSchemaReadContextPtr rightContext = InitializeReadContextWithAllSchemas(rightSchemasXml);
+    bvector<ECN::ECSchemaCP> rightSchemas = rightContext->GetCache().GetSchemas();
+
+    // Merge the schemas
+    SchemaMergeResult result;
+    MergerTestIssueListener issues;
+    result.AddIssueListener(issues);
+    EXPECT_EQ(BentleyStatus::ERROR, SchemaMerger::MergeSchemas(result, leftSchemas, rightSchemas));
```
rschili, line 5557: I believe you found a new bug here :-) Schema Item and Schema Names are supposed to ignore case. So the result should never treat this input differently as if they had the identical names.... Let's fix this while we're at it. The intended result for this test is Success with the right schema xml as the output.

rschili (changes requested):

iModelCore/ecobjects/src/SchemaMerger.cpp
6ar8nas: The reworked behavior will only require one issue.

iModelCore/ecobjects/src/SchemaMerger.cpp
6ar8nas: I could, but from what [this comment](https://github.com/iTwin/imodel-native/pull/644#discussion_r1517532994) suggests that was the cause of a bug.

iModelCore/ecobjects/test/NonPublished/SchemaMergerTests.cpp
6ar8nas: The ECSchema name comparison utility function `CompareByName` explicitly notes case-sensitive checking to be the expected behavior. However. as per what we've discussed, SchemaMerger will now disregard case when merging schemas.

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -2466,7 +2466,7 @@ struct SchemaKeyLessThan
 };
 
 typedef bmap<SchemaKey , ECSchemaPtr> SchemaMap;
-typedef std::function<bool(const std::pair<SchemaKey, ECSchemaPtr>)> SchemaKeyMatchFnPredicate;
+typedef std::function<bool(SchemaKeyCR)> SchemaKeyMatchFn;
```
rschili, line 2469: Just very minor. I checked our codebase and we never use the Fn suffix. Let's not introduce it now. Some existing functions use xCallback, xPredicate (though there already is a SchemaKeyMatchPredicate) or xHandler but Handler is not a good match here. I suggest "SchemaKeyMatchCallback".

rschili (approved):

swwilson-bsi (approved):

PR #646, opened 2024-02-28
Update OpenSSL from 3.1.4 to 3.1.5

jiaruiz717:
Update OpenSSL from 3.1.4 to 3.1.5


nick4598 (approved):
Could you modify imodel-native/iModelCore/libsrc/openssl/BentleyVersionString.c from 3.1.1 to 3.1.5? Looks like we missed that the past couple of times.

WindaZ:
> Could you modify imodel-native/iModelCore/libsrc/openssl/BentleyVersionString.c from 3.1.1 to 3.1.5? Looks like we missed that the past couple of times.

Done

Issue #640, opened 2024-02-26
Presentation: Large amounts of logs block the main thread


grigasp:
- Presentation library emits large amounts of logs at `INFO` level when creating content. Should look into reducing severity of some (or all) of these messages.
  ```csv
  ...
  "event-46e2e920325608dca5677b102c000000","0x55ec9816","2024-02-20T20:55:31.7620000Z","Information","Content is built from 1 queries",""
  "event-46e2e920325608dca6677b102c000000","0x60d3f12b","2024-02-20T20:55:31.7620000Z","Information","Query: `SELECT 1 AS [/ContractId/], GetInstanceKey(IFNULL([rel_RVT13704_ProfileRoof_0].[ECClassId],0),[rel_RVT13704_ProfileRoof_0].[ECInstanceId]) AS [/ECInstanceKeys/], CAST(null AS TEXT) AS [/InputECInstanceKeys/], '[]' AS [/RelatedInstanceInfo/], [rel_RVT13704_ProfileRoof_0].[Timestamp] AS [pc_RVT13704_RevitIdPropertyMixinClass_Timestamp/41] FROM [RVT13704].[ProfileRoof] [rel_RVT13704_ProfileRoof_0] INNER JOIN [bis].[GraphicalElement3dRepresentsElement] [rel_bis_GraphicalElement3dRepresentsElement_40] ON [rel_RVT13704_ProfileRoof_0].[ECInstanceId] = [rel_bis_GraphicalElement3dRepresentsElement_40].[TargetECInstanceId] INNER JOIN [generic].[Graphic3d] [related] ON [related].[ECInstanceId] = [rel_bis_GraphicalElement3dRepresentsElement_40].[SourceECInstanceId] WHERE ([related].[ECInstanceId] IN (?))`",""
  "event-46e2e920325608dca7677b102c000000","0x9bb2e6af","2024-02-20T20:55:31.7620000Z","Information","Read content items: 0",""
  "event-46e2e920325608dca8677b102c000000","0x55ec9816","2024-02-20T20:55:31.7620000Z","Information","Content is built from 1 queries",""
[...]

PR #648, opened 2024-02-29
Non blocking logging

grigasp:
Closes https://github.com/iTwin/imodel-native/issues/640.
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6477

This is a significant change to how logs, emitted on threads other than main, are handled. Previously, they were accumulated in a `vector` and flushed using an explicit function call. This has two problems in cases when there are many logs:
1. Possible memory spikes due to accumulated logs.
2. Possible main thread blocking while the logs are being synchronously flushed on the main thread.

With this PR we start using `Napi::ThreadSafeFunction` to immediately queue the message for emitting on the main thread as soon as it's logged on a worker one.

Additionally, this PR makes it clear what the interface of JS logger should be. Previously it said it's `Logger` (from `@itwin/core-bentley`), but really it was expecting `typeof Logger` and was using its private static members. I defined an interface clearly stating what we expect and adjusted `Logger` in `@itwin/core-bentley` to match the requirements.

Finally, I reduced severity of a number of `ECPresentation` log messages from `LOG_INFO` to `LOG_TRACE`.


saskliutas (approved):

iModelJsNodeAddon/api_package/ts/src/test/index.ts
```diff
@@ -36,7 +36,23 @@ chaiuse(chaiAsPromised);
 
 Logger.initializeToConsole();
 Logger.setLevelDefault(LogLevel.Error);
-iModelJsNative.logger = Logger;
+iModelJsNative.logger = {
+  // note: using private Logger fields is temporary until the version of `@itwin/core-bentley` is updated
```
pmconne, line 40: Will you file an issue and ensure we address it ASAP?

grigasp: Filed an issue for tracking: https://github.com/iTwin/imodel-native/issues/649.

I don't think there's a need to hurry - we were using those private members in our native code for a while anyway. I was planning to fix this at some point when a new `itwinjs-core` nightly build, using this new addon, is released.

PR #653, opened 2024-03-04
Fixed geomlib runtime crashes for BytesToXXX API

saeeedtorabi:


iModelCore/GeomLibs/PublicAPI/GeomSerialization/GeomLibsFlatBufferApi.h
```diff
@@ -22,25 +22,55 @@ BEGIN_BENTLEY_GEOMETRY_NAMESPACE
...
 public:
 
 // test if the bytes have the signature for BentleyGeomFlatBuffer.
-static BGFBIMPEXP bool IsFlatBufferFormat (bvector <Byte> &buffer);
-static BGFBIMPEXP bool IsFlatBufferFormat (Byte const *buffer);
+static BGFBIMPEXP bool IsFlatBufferFormat(bvector <Byte> &buffer);
+static BGFBIMPEXP bool IsFlatBufferFormat(Byte const *buffer);
+
+static BGFBIMPEXP IGeometryPtr BytesToGeometry(bvector <Byte> &buffer, bool applyValidation = true);
+//! @deprecated in 4.x. Use `BytesToGeometrySafe`.
```
pmconne, line 36: Why deprecate? imodel-native APIs are not subject to public API support policies.

saeeedtorabi: (my bad; I should have mentioned the reason in the commint message; will add)

The reason for deprecation is those functions are not safe and crash at run time if they receive bad binary files as input. We had change the API argument list so had to deprecate them. We (Dave and I) are working with Mike Belousov to reveal and fix those kind of run time crashes.

pmconne: imodel-native APIs are not public. You are free to delete them, or change their implementations and/or argument lists.

pmconne: For imodel02 I would not expect so. Connectors would update to new version of imodel02 and adjust their code accordingly if needed. right @jffmarker?

Can't comment on PP.

dassaf4: This is good to know. Last year we stripped a bunch of stuff from imodel-native geomlibs in order to make it more palatable for open sourcing, so I was operating under some assumptions about its API evolution. We have dozens of deprecations in native geomlibs over the years, but unfortunately no process to remove the deprecated methods. So what I'm reading here is that we could sweep away the deprecations in imodel-native and imode02. PPBase geomlibs is more restrictive: I've been told not to change signatures in the PublicAPI there, and I'm not aware of a process to clean up deprecations there.

jffmarker: @pmconne @saeeedtorabi  Correct, for imodel02 consumers, there's no strict API lifecycle. If you make a breaking change, you will be expected to help consumers port or refactor when they get your changes from biweekly builds, but you are generally free to make changes, especially for security concerns.

pmconne: > we could sweep away the deprecations in imodel-native and imode02

Yep.

iModelCore/GeomLibs/PublicAPI/GeomSerialization/GeomLibsFlatBufferApi.h
```diff
@@ -22,25 +22,55 @@ BEGIN_BENTLEY_GEOMETRY_NAMESPACE
 class BentleyGeometryFlatBuffer
 {
 private:
-BentleyGeometryFlatBuffer (){} // no intances.
+BentleyGeometryFlatBuffer (){} // no instances.
+// template <class PtrName, class ReaderMethodName>
+// PtrName BytesToXXXSafe(Byte const *buffer, size_t const bufferSize, bool applyValidation);
```
MichaelBelousov, line 27: remove dead comments

MichaelBelousov (approved):
cursory review, LOKTM

Avoid too many white space/formatting changes in PRs that aren't explicitly about that, because it makes reading for what you actually changed here more difficult.

iModelCore/GeomLibs/PublicAPI/GeomSerialization/GeomLibsFlatBufferApi.h
saeeedtorabi: Thanks, will remove the dead code. The PR was not 100% ready to push. That's why it's a draft.
Also the white space removal is because I have this feature enabled automatically on save.

iModelCore/GeomLibs/PublicAPI/GeomSerialization/GeomLibsFlatBufferApi.h
saeeedtorabi: Removed deprecation and applied API chanegs to the existing API. Please "resolve conversation" if you are satisfied with the change.

saeeedtorabi: @pmconne

pmconne (approved):

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -1918,89 +1922,92 @@ static void ReadVariantGeometry(const BGFB::VariantGeometry * fbGeometry, bvecto
...
-    auto fbRoot = flatbuffers::GetRoot <BGFB::VariantGeometry>(fbStart);
-    if (nullptr != fbRoot
-        && fbRoot->has_geometry_type ()
-        && fbRoot->has_geometry ()
-        )
+        return nullptr;
+    auto fbRoot = flatbuffers::GetRoot<BGFB::VariantGeometry>(fbStart);
+    auto myVerifier = flatbuffers::Verifier(fbStart, bufferSize - s_prefixBufferSize);
+    if (!fbRoot->Verify(myVerifier))
+        return nullptr;
```
dassaf4, line 1954: Move the new verifier lines below, inside the if (nullptr != fbRoot && ... ) scope.

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -1918,89 +1922,92 @@ static void ReadVariantGeometry(const BGFB::VariantGeometry * fbGeometry, bvecto
...
-        auto result = FBReader::ReadGeometry (fbRoot);
-        if (applyValidation && s_readValidator.IsValid())
+        PtrType result = readerMethod(fbRoot);
+        if (result.IsValid())
             {
-            if (!result->IsValidGeometry(s_readValidator))
-                return nullptr;
+            if (applyValidation && s_readValidator.IsValid())
+                if (!result->IsValidGeometry(s_readValidator))
+                    return nullptr;
```
dassaf4, line 1962: Style tip: if an `if` statement is followed by a block of more than one line, use curly braces:

```suggestion
            if (applyValidation && s_readValidator.IsValid())
                {
                if (!result->IsValidGeometry(s_readValidator))
                    return nullptr;
                }
```

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -2011,10 +2018,10 @@ bvector<IGeometryPtr> *invalidGeometry)
     if (invalidGeometry)
         invalidGeometry->clear ();
     auto fbRoot = flatbuffers::GetRoot <BGFB::VariantGeometry>(fbStart);
-    if (nullptr != fbRoot
-        && fbRoot->has_geometry_type ()
-        && fbRoot->has_geometry ()
-        )
+    auto myVerifier = flatbuffers::Verifier(fbStart, buffer.size() - s_prefixBufferSize);
+    if (!fbRoot->Verify(myVerifier))
+        return false;
```
dassaf4, line 2023: Move the new verifier lines below, inside the if (nullptr != fbRoot && ... ) scope.

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -2024,20 +2031,27 @@ bvector<IGeometryPtr> *invalidGeometry)
...
-        return false;        
+        return false;
     auto fbRoot = flatbuffers::GetRoot <BGFB::VariantGeometry>(fbStart);
-    if (nullptr != fbRoot
-        && fbRoot->has_geometry_type ()
-        && fbRoot->has_geometry ()
-        )
+    auto myVerifier = flatbuffers::Verifier(fbStart, bufferSize - s_prefixBufferSize);
+    if (!fbRoot->Verify(myVerifier))
+        return false;
```
dassaf4, line 2050: Move the new verifier lines below, inside the if (nullptr != fbRoot && ... ) scope.

dassaf4 (approved):
Looks good up to this point. Keeping open since we have more fuzz crashes and test changes forthcoming.

iModelCore/GeomLibs/geom/test/SerializationTest/t_imodeljson.cpp
```diff
@@ -121,44 +111,62 @@ TEST(IModelJson,BytesToXXX)
...
+        if (Check::True(ReadIModelJson (
+            geometry, L"IModelJson", L"IModelJsonFromNative", fileName, nullptr, L"imjs", nullptr, L"fbjs"),
+            "read imjs file"))
             {
-            for (auto const& g : geometry)
+            for (IGeometryPtr const& g : geometry)
                 {
                 if (Check::True(g.IsValid(), "Imported geometry is valid"))
-                    Check::NearRoundTrip(*g, 0.0, name);
+                    Check::NearRoundTrip(*g, 0.0, (char*) fileName);
```
dassaf4, line 133: This is not good practice, even if it works in this case. We are stuck with a plethora of char* types in the geomlibs test API, but this type should be avoided in general, in favor of WCharCP (defined to be wchar_t const*) for representing strings. A better solution would be to create an overload of `NearRoundTrip` that takes WCharCP; then the existing method can call the new overload passing in `WString(pString, BentleyCharEncoding::Locale).c_str()`. Otherwise, if you keep the char* cast, you should at least warn about the bad practice with a comment.

dassaf4 (approved):

iModelCore/GeomLibs/geom/test/SerializationTest/t_imodeljson.cpp
saeeedtorabi: As discussed will address this in the next PR.

saeeedtorabi: Due to too many merge conflicts, I decided to delay the whiteSpace PR. Therefore, I want to address this comment in this PR.

Because we are allowed to change imdoel-native and imodel02 APIs, I will update the existing API to take `WCharCP` instead of `char const*` in these 2 repos. For PPBase I will deprecate the old API and add a new overload. Note that there is only 1 other call to `Check::NearRoundTrip` in imodel-native geomlibs.

Please let me know if this is a good plan.

dassaf4: Sounds good.  Actually, I'd just change the arg to WCharCP in _both_ repos, since I don't think the Check class is public API (being in test directory).

saeeedtorabi: I need to discuss this with you. If we change API in imodel-native and imodel02 but deprecate in PPBase we end up having different names in different repos for API with the same behaviour. So I was thinking even if we dont deprecate in imodel-native and imodel02, we should use the same API name as PPBase.

dassaf4: Yes, same names for `Check::NearRoundTrip` in all 3 native repos. No deprecation needed for changes to `Check` since it isn't public API.

saeeedtorabi: I dont meant just `Check::NearRoundTrip`
I meant all other functions that we changes line `BytesToGeometry` or `BytesToCurvePrimitive`
I dont think it makes sense to have `BytesToGeometry` in 2 repos and then `BytesToGeometrySafe` in another repo when both do the same exact thing.

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -1621,6 +1658,8 @@ static ICurvePrimitivePtr ReadCurvePrimitive (const BGFB::VariantGeometry * fbGe
             {
             BGFB::TransitionSpiral const *fbTransitionSpiral
                 = reinterpret_cast <const BGFB::TransitionSpiral *> (fbGeometry->geometry ());
+            if (!fbTransitionSpiral)
+                return nullptr;
```
dassaf4, line 1662: This change is fine. But please also add an early out after the next line when `detail` is nullptr.

dassaf4 (approved):
Review of next batch of changes...

iModelCore/GeomLibs/geom/test/SerializationTest/t_imodeljson.cpp
dassaf4: Those methods are in the PPBase public API, unfortunately.

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -1566,13 +1597,17 @@ static ICurvePrimitivePtr ReadCurvePrimitive (const BGFB::VariantGeometry * fbGe
...
             int  order = fbInterpolationCurve->order ();
             bool closed = fbInterpolationCurve->closed () != 0;
             auto fbFitPoints = fbInterpolationCurve->fitPoints ();
             auto fbKnots = fbInterpolationCurve->knots ();
             auto fbStartTangent = fbInterpolationCurve->startTangent ();
             auto fbEndTangent = fbInterpolationCurve->endTangent ();
-            int numPoles = fbFitPoints->Length () / 3;
+            if (!fbFitPoints || !fbKnots || !fbStartTangent || !fbEndTangent)
+                return nullptr;
```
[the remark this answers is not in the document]
saeeedtorabi: ok will update PPBase

iModelCore/GeomLibs/geom/test/SerializationTest/t_imodeljson.cpp
```diff
@@ -44,9 +72,7 @@ WCharCP flatbufferUInt8Extension = nullptr   // optionally write flatbuffer to s
         {
         if (Check::True (IModelJson::TryIModelJsonStringToGeometry (string, geometry), "json string to geometry"))
             {
-            Utf8String stringB;
-            if (IModelJson::TryGeometryToIModelJsonString (stringB, geometry))
-                GTestFileOps::WriteToFile (stringB, outputDirectory, nameB, nameC, extension);
+            writeGeometryToFile(geometry, outputDirectory, nameB, nameC, extension);
```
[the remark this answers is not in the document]
saeeedtorabi: I added check for return value.

iModelCore/GeomLibs/geom/test/SerializationTest/t_imodeljson.cpp
saeeedtorabi: Yes, I know. I mean in imodel02 and imodel-native, rather than changing the body and keeping the API name, we also change the API name to match with PPBase. For example, in PPBase the new method would be called `BytesToGeometrySafe` so in imodel02 and imodel-native we also should call the safe method `BytesToGeometrySafe` and not `BytesToGeometry` to be consistent with PPBase.

dassaf4: Ah, I see. I guess since all three native repos are going forward for the time being, it would be nice for their APIs to be the same.  How about this: add simple "Safe"-suffixed wrapper methods to imodel02/imodel-native that just call the updated `BytesToGeometry` methods. This way there is still a 1-1 correspondence of methods in the FB APIs of the three repos.  Callers in imodel02/imodel-native can continue to use the newly updated BytesToGeometry methods; no imodel02/imodel-native methods will actually call the new "Safe" wrappers, but this will make it apparent to which imodel02/imodel-native methods the PPBase "Safe" methods map. You don't even have to put the new "Safe" wrappers in the imodel02/imodel-native API .h file. You might want to write a test that covers them, and verifies that they produce the same output as the methods they wrap.

saeeedtorabi: yes, I like this idea. I will add the safe wrappers. 

FYI, I replaced `char const*` with `WCharCP` in `Check::NearRoundTrip` args so I can resolve this conversation.

iModelCore/GeomLibs/geom/test/src/checkers.cpp
```diff
@@ -1839,14 +1839,12 @@ void Check::TearDown()
...
+bool Check::NearRoundTrip(IGeometryCR g, double tolerance, WCharCP pString)
     {
+    std::wstring ws(pString);
     std::string myString;
-    if (pString)
-        {
-        myString.append(pString);
-        myString.append(": ");
-        }
+    std::transform(ws.begin(), ws.end(), std::back_inserter(myString), [](wchar_t c){ return (char)c; });
```
dassaf4, line 1846: This casting of wchar_t to char is bad practice, because it won't always work.

This could be the tip of the iceberg, but the cleanest change would be to propagate the WCharCP arg to the other Check::XXX methods called by `Check::NearRoundTrip`. In other words, change those called methods' signatures too.

If you don't want to do that, the conventional way to turn `pString` into a `char` string is using `BeStringUtilities::WCharToCurrentLocaleChar`.  See WString.h for example usage.

WString("B-spline surface: ", BentleyCharEncoding::Locale).c_str()

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -1618,11 +1618,11 @@ static ICurvePrimitivePtr ReadCurvePrimitive (const BGFB::VariantGeometry * fbGe
             if (nullptr != fbEndTangent)
                 endTangent = DVec3d::From (fbEndTangent->x (), fbEndTangent->y (), fbEndTangent->z ());
 
-            int numKnots = fbKnots ? fbKnots->Length () : 0;
+            int numKnots = fbKnots->Length ();
             double const * pKnots = numKnots > 0 ? (double const*)fbKnots->GetStructFromOffset(0) : nullptr;
 
             MSInterpolationCurve curve;
-            curve.Populate (
+            if(curve.Populate (
```
dassaf4, line 1625: space after if.  (And no space after Populate.)

dassaf4 (approved):
Review of today's changes:

PR #665, opened 2024-03-08
Special case root subject remapping

jiaruiz717:
As https://github.com/iTwin/imodel-transformer/issues/149
Make root subject special cased when remapping from source to target


pmconne:
Tests?

nick4598:
I think it could be worth adding some tests into the TS part of the imodel-native repo (see DgnDb.test.ts for an example test file).

Perhaps those should validate certain things after calling cloneElement:

- ParentId of non root-subject == rootSubjectId == scope on the code of non root-subject
- A remap rule of IModel.rootSubjectId -> IModel.rootSubjectId works as expected. (Note this is kind of a forced example but someone could do this)
      - I.e. rootSubjectId's parent still doesn't exist after the clone, and the scope on its code is still itself.
-   Remapping a non root-subject to a non root-subject keeps its parent and code as expected

You and I can coordinate on that

PR #674, opened 2024-03-13
Concurrent query locked up when a long running query was about to be cancelled

khanaffan:
Concurrent query by default attempts to cancel a query that running for more than 60 seconds. A long-running query here means any query that when stepped spends all the time in a single step. This prevents the normal quota base cancel from happening. A separate monitor thread kicks in and it tries to interrupt the query. But it was unable to interrupt and spend time waiting for a lock held by a long-running query thread. This lockup prevents new query requests from being assigned a thread resulting in the full queue and eventually not accepting new requests at all. 

The lock was misplaced for a block of code and not for the whole time the query was running.

An example of long running query that takes forever to run in single steps is below. We attempt to use MATCH with bbox that includes all elements surprisingly it takes a long time for a very large model. 

It was also found `imodel_spatial_overlap_aabb` was not blocking the thread it run in parallel on all threads it is just that for a very large bbox it can take very long to return matching elements. Which appears as if it is blocking but it was an attempt to cancel it that locked up the concurrent query. 

```sql
      WITH sequence(n) AS (
        SELECT 1 UNION ALL SELECT n + 1 FROM sequence WHERE n < ${iteration}
      )
      SELECT COUNT(*) FROM bis.SpatialIndex i, sequence s
        WHERE i.ECInstanceId MATCH
          imodel_spatial_overlap_aabb(
            (SELECT imodel_bbox(MIN(Minx),MIN(MinY), MIN(MinZ), MAX(MaxX),MAX(MaxY), MAX(MaxZ)) FROM bis.SpatialIndex))
```


pmconne (approved):

rschili (approved):

aruniverse (approved):

aruniverse:
@backport release/4.x

aruniverse:

PR #679, opened 2024-03-20
Disable schema sync when doing profile/domain schema upgrades

RohitPtnkr1996:
Pulling/pushing changes with schema sync when doing a profile upgrade on a briefcase and will leave the sync db's be_prop out of sync even when the ec_ tables are updated.
With the new workflow, we will disable schema sync internally when doing a profile/domain schema upgrade and re-enable it after the upgrade is done.
The upgraded briefcase will then need to push the changes, which the other briefcases will have to pull in order to sync up.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6536


khanaffan (approved):

nick4598:
Noticed this test is failing:
1) ImportContext
       should successfully change type of property during ImportContext.cloneElement even though the two schemas have the same classId within their own database:
     Error: wrong class
      at Context.<anonymous> (D:\vsts_a\2\s\imodel-native\iModelJsNodeAddon\api_package\ts\lib\test\ImportContext.test.js:53:35)
      
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2755684&view=logs&jobId=e7aeb5d1-10c0-5cf0-50bd-ace52912433b&j=e7aeb5d1-10c0-5cf0-50bd-ace52912433b&t=f7454ded-b844-5713-3c8d-1d8dd13c1041

Even though that test is not merged into main. @chuckkir any ideas? I guess below agent might be in a bad state?
pool: [iModelTechCI](https://dev.azure.com/bentleycs/2c48216e-e72f-48b4-a4eb-40ff1c04e8e4/_settings/agentqueues?poolId=&queueId=280)
Agent: BUILDMINION09A

chuckkir:
What indication is there that it’s the machine? We have plenty of flakey tests. I went ahead and rebooted but I don’t see any reason to believe that will affect this test.

nick4598:
> What indication is there that it’s the machine? We have plenty of flakey tests. I went ahead and rebooted but I don’t see any reason to believe that will affect this test. 

The reason I'm leaning towards the machine is because this test does not exist in his branch at all. I feel like we have had similar problems in the past where the built TS tests seem to not get cleaned up and rebuilt properly, but thats just a feeling and I dont have any concrete examples

nick4598:
> You could be correct. We don’t delete the source tree, so if the test is writing into the source tree rather than the output tree that could be an issue with the tests.


The tests live in a path like:

- src\imodel-native\iModelJsNodeAddon\api_package\ts\src\test\DgnDb.test.ts

And when built I believe:
   - src\imodel-native\iModelJsNodeAddon\api_package\ts\lib\test\DgnDb.test.js
   
That is just what my local tree looks like, but the machines probably look pretty much identical to this. That might explain why we have these issues. Idk what it'd look like if we fixed this, I guess we could move the lib folder over to the out tree but I assume you'd have to copy other pieces over as well such as the node_modules. Currently those are 280 MB in size on my machine which doesn't seem too bad.

chuckkir:
That would make sense for the cause. Generally we symlink rather than copy if possible. Probably the simplest fix would be to make ts/lib a symlink into the output tree so that the contents will get removed on clean. Because the symlink will remain after cleanup there may be some nuances to the implementation like checking if it exists before creating the symlink but I think it should work.

pmconne (approved):

khanaffan (approved):

PR #685, opened 2024-03-22
Restart query trigger unrelated queries to be cancelled.

khanaffan:
When restart token is provided,  attempt is made to cancel any previous query with that restart token. But doing so cause any currently running query to be get cancelled.


pmconne:
Testable?

khanaffan:
> Testable?

Yes but will take time. I do not want to add a flaky test. It needs a few running queries and a query in the queue and then sends a new query with a restart token. 

I have a test right now but that sends a batch of queries and takes half a minute.  I am trying to reduce it so it can run reliability in our test suite in under a few seconds.

This PR is needed by the pineapple team as they cannot deploy a new version so I differed the test. and created the PR.

pmconne (approved):

aruniverse:

PR #688, opened 2024-03-25
Allow setting composite EC properties to null

6ar8nas:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6884

The changes test for different updating to null struct and array behaviors through various updating mechanisms and fixes updating a prefilled ECInstance from a JSON value through handling null cases. 

Closes iTwin/itwinjs-backlog#1057


iModelCore/ECDb/ECDb/ECSql/ECInstanceECSqlSelectAdapter.cpp
```diff
@@ -320,7 +317,7 @@ BentleyStatus ECInstanceECSqlSelectAdapter::SetPrimitiveValue(ECValueR val, ECN:
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-BentleyStatus ECInstanceECSqlSelectAdapter::SetStructArrayElement(ECValueR val, ECClassCR structType, IECSqlValue const& value) const
+BentleyStatus ECInstanceECSqlSelectAdapter::SetStructElement(ECValueR val, ECClassCR structType, IECSqlValue const& value) const
```
rschili, line 320: I'm a bit confused. You renamed the method from StructArray to just Struct, but the method contents didn't change, and it is called for struct arrays, yet it does not contain any logic regarding arrays.
Does it work with arrays or not?  Is SetStruct(nullptr) the appropriate call for a struct array?
If it all works fine, let's at least add a comment explaining that this method works for both, arrays and normal structs.

6ar8nas: This has been called for each of the array elements, not struct array itself. Seeing the same logic (just the non-array struct handling was lacking a null check) implemented twice, I just realized we could reuse the same SetStructArrayElement, which seems to not have anything to do with arrays inside of it.

khanaffan:
Following test are failing
```
[  FAILED  ] ECInstanceInserterTests.InsertSingleInstanceOfComplexClass
[  FAILED  ] ECInstanceInserterTests.InsertSingleRuleInstance
[  FAILED  ] ECSqlStatementTestFixture.StructUpdateWithDotOperator
[  FAILED  ] ECSqlStatementTestFixture.WriteCalculatedECProperty
```

6ar8nas:
ECInstance doesn't seem to support the requested functionality. Instead, an alternative approach of eliminating ECInstance from the workflow altogether has been selected. Follow-up: #719

iModelCore/ECDb/ECDb/ECSql/ECInstanceECSqlSelectAdapter.cpp
6ar8nas: Resolving this, as an alternative approach has been selected.

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -11800,4 +11800,460 @@ TEST_F(ECSqlStatementTestFixture, SelectAnySomeAll)
...
+    ecsql.Sprintf("INSERT INTO ts.TestClass (array_i) VALUES (?)");
+    ECSqlStatement stmt;
+    ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, ecsql.c_str())) << ecsql.c_str();
+    stmt.GetBinder(1).AddArrayElement().BindInt(i);
+
+    ECInstanceKey key;
+    ASSERT_EQ(BE_SQLITE_DONE, stmt.Step(key)) << stmt.GetECSql();
+    stmt.Finalize();
+
+    ecsql.Sprintf("UPDATE ONLY ts.TestClass SET array_i=? WHERE ECInstanceId=%s", key.GetInstanceId().ToString().c_str());
```
rschili, line 11900: For anything that goes into SQL we use SqlPrintfString instead

rschili: Oh, and for specifying an ECInstanceId like you do here and several times more in the method, it's better to not printf it into the statement but use stmt.BindId() instead

khanaffan (approved):

rschili (approved):

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
rschili: As discussed stip the printf calls which do not have any parameter, then this conversation can be marked as resolved. Thanks

6ar8nas: Sorted that out as well.

PR #690, opened 2024-03-25
Improve performance of drop schema

khanaffan:
Dropping a very large number of schemas takes a very long time. With a specific iModel dropping `922` schemas takes 1 hour and 25 minutes. 

Most of the time is spent trying to figure out if the schema has an instance or has FK references.

To improve performance the check now limits the instance search to specific schema classes previously it used root relationships and classes to search for instances. This causes unrelated relationships to be searched for schema classes that are going to be dropped. 

This improvement for the given data set brings overall time to 15 min from over an hour.


ColinKerr (approved):
This change looks fine but we still need perf tests for this improvement.

PR #693, opened 2024-03-27
Presentation: Fix a memory leak

grigasp:
`NativeECPresentationManager` was leaking `DefaultPropertyFormatter` instance - `ECPresentationManager` was taking a raw pointer with the intention that its creator would own the property formatter, however `NativeECPresentationManager` never deleted the object. 

While this was the only leak I found, I went ahead and changed a number of other raw pointers to smart ones, making such mistakes harder to do in the future.


saskliutas (approved):

pmconne (approved):

grigasp:

PR #696, opened 2024-03-28
Removed unsafe BytesToXXX API

saeeedtorabi:
imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/442


iModelCore/GeomLibs/geom/test/CurvePrimitiveTest/t_IGeometry.cpp
```diff
@@ -113,13 +113,13 @@ void DoRoundTrip (IGeometryPtr g0, bool emitGeometry, int serializerSelect)
...
         PolyfaceQueryCarrier carrier0 (0, false, 0, 0, nullptr, nullptr);
-        Check::True(BentleyGeometryFlatBuffer::BytesToPolyfaceQueryCarrier(&buffer0[0], buffer0.size(), carrier0));
+        Check::True(BentleyGeometryFlatBuffer::BytesToPolyfaceQueryCarrierSafe(&buffer0[0], buffer0.size(), carrier0));
 
         bvector<Byte> buffer1;
         BentleyGeometryFlatBuffer::GeometryToBytes (*polyface, buffer1);
 
         PolyfaceQueryCarrier carrier1 (0, false, 0, 0, nullptr, nullptr);
-        Check::True(BentleyGeometryFlatBuffer::BytesToPolyfaceQueryCarrier(&buffer1[0], buffer0.size(), carrier1));
+        Check::True(BentleyGeometryFlatBuffer::BytesToPolyfaceQueryCarrierSafe(&buffer1[0], buffer0.size(), carrier1));
```
dassaf4, line 122: Make sure you apply all the _later_ commits from imodel02.

```suggestion
        Check::True(BentleyGeometryFlatBuffer::BytesToPolyfaceQueryCarrierSafe(&buffer1[0], buffer1.size(), carrier1));
```

iModelCore/GeomLibs/geom/test/CurvePrimitiveTest/t_IGeometry.cpp
```diff
@@ -1126,7 +1126,7 @@ TEST(FlatBuffer, MSBsplineSurface)
     }
     template <typename GeometryTypePtr>
     void TestSpecialReader(GeometryTypePtr &validGeometry, GeometryTypePtr &invalidGeometry,
-    GeometryTypePtr (*bytesToGeometry)(Byte const *bytes, size_t const bufferSize, bool applyValidation))
+    GeometryTypePtr (*bytesToGeometrySafe)(Byte const *bytes, size_t const bufferSize, bool applyValidation))
```
dassaf4, line 1129:
```suggestion
    GeometryTypePtr (*bytesToGeometrySafe)(Byte const *bytes, size_t bufferSize, bool applyValidation))
```

iModelCore/GeomLibs/serialization/src/FlatBuffer/FixedStructs.cpp
```diff
@@ -2062,69 +2062,39 @@ PtrType BytesToXXX(Byte const *buffer, size_t const bufferSize, bool applyValida
...
 
-IGeometryPtr BentleyGeometryFlatBuffer::BytesToGeometry(Byte const *buffer, size_t const bufferSize, bool applyValidation)
+IGeometryPtr BentleyGeometryFlatBuffer::BytesToGeometrySafe(Byte const *buffer, size_t bufferSize, bool applyValidation)
     {
-    return BytesToXXX<IGeometryPtr>(buffer, bufferSize, applyValidation, FBReader::ReadGeometry);
+    return BytesToXXXSafe<IGeometryPtr>(buffer, bufferSize, applyValidation, FBReader::ReadGeometry);
     }
 
-IGeometryPtr BytesToGeometrySafe(Byte const *buffer, size_t const bufferSize, bool applyValidation)
+IGeometryPtr BentleyGeometryFlatBuffer::BytesToGeometry(bvector<Byte> &buffer, bool applyValidation)
```
dassaf4, line 2070: If a const arg can be passed in, it should be const in the signature.

```suggestion
IGeometryPtr BentleyGeometryFlatBuffer::BytesToGeometry(bvector<Byte> const& buffer, bool applyValidation)
```

dassaf4 (approved):

saeeedtorabi:
> What am I overlooking? The net change appears to have been to append "Safe" to a bunch of function names, and adjust every caller in reaction to that change. If the unsafe functions no longer exist, why not reuse the original name for the now-safe functions and leave all callers alone?

Initially we only made the unsafe API safe and did not rename them. We just added a `size_t` arg to the argument list as the second arg to make them safe. Third arg is `bool`. The problem with that was users could still keep the old code and the compiler would cast `bool` to `size_t` which would lead to a run time crash and we did not want that to happen. We changed all the API calls we could find but still there might be API calls that we are unaware of that. Therefore, we renamed the API so users code which still kept the old code does not compile rather than run time crash.

Another reason for rename was, in PBBase we cannot rename. We have to deprecate unsafe API and add new safe API. Renaming in imodel-native and imodel02 would keep all 3 repos consistent.

pmconne (approved):

PR #702, opened 2024-04-04
Enable SQLite ICU feature

khanaffan:
**NOTE:  This PR by itself only enable ICU support it does not make any breaking change. But the feature it enable when used to solve following problem can lead to a breaking change for file as it cannot be read with older code.**

## CodeValue uniqueness
Sqlite `NOCASE` is an ASCII collation. So CodeValue `Löwe` and `LÖWE` are still treated as two different `CodeValue`s.  If lower/upper casing is appplied to this string it only affects ASCII characters and the non-ascii Latin character set is skipped or left as is.

### Possible Solution
```
m_db.SetNoCaseCollation(NoCaseCollation::Latin1)
```
and it can be switched back

```
m_db.SetNoCaseCollation(NoCaseCollation::ASCII)
```
Here is how we can do this.

•	We enable Latin1 by default.
•	For applying changeset we switch back to ASCII as we do not want it to fail for existing data.

One main consequence is if we have a UNIQUE index on that column or if that column is part of a UNIQUE index, then that index has duplicate values. 
•	First, we cannot call REINDEX to rebuild the index as it will fail with UNIQUE_CONSTRAINT.
•	When searching for Latin1-wise duplicate value the accent character must have the same casing though the ASCII part can be anything. This means even after enabling LATIN1 any old duplicate values will be still ASCII case insensitive.

Besides if an application was already putting out the duplicate value in a non-ascii sense then that application might see failure as e.g À, Á, Â, Ã, Ä, Å, à, á, â, ã, ä, å and a are under LATIN1 case insensitive is treated as the same letter a.


kabentley:
i don't see why we need to switch back to ASCII to apply changesets.

khanaffan:
> I don't see why we need to switch back to ASCII to apply changesets.

if we do have duplicates in our existing data, it will fail to apply the changeset. We do not want the already pushed changeset to fail. 

This is temporary for existing data on the cloud. At some point in the future, we can remove this logic.

kabentley:
> if we do have duplicates in our existing data, it will fail to apply the changeset. We do not want the already pushed changeset to fail.

Why is applying a changeset is any different from any other direct change? If the changeset introduces a new conflict is *should* fail.

khanaffan:
Their is some issue with this branch where i am unable to push changes. I have another PR https://github.com/iTwin/imodel-native/pull/722

PR #707, opened 2024-04-08
Fix logger crash on app exit

grigasp:
Fixes https://github.com/iTwin/imodel-native/issues/706

### DIFF
--- a/iModelJsNodeAddon/JsLogger.cpp
+++ b/iModelJsNodeAddon/JsLogger.cpp
@@ -110,7 +110,7 @@ void JsLogger::LogMessage(Utf8CP category, SEVERITY sev, Utf8CP msg)
         {
         logToJs(category, sev, msg);
         }
-    else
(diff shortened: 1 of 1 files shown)


pmconne (approved):
extract-api is probably switching the order of things when run locally vs CI...clean, build, extract-api should fix. EDIT: uh except you don't have any TypeScript changes...

grigasp:
> extract-api is probably switching the order of things when run locally vs CI...clean, build, extract-api should fix. EDIT: uh except you don't have any TypeScript changes...

yes, this one's interesting - I contacted @DanRod1999 regarding this

DanRod1999:
looking at some other runs, to see if they also fail. Don't really understand what's going on, but I would assume that if this problem keeps happening master branch in itwinjs-core should also be broken, but based on its run history it seems fine. Might just have to wait for the outcome of this current run.

grigasp:
@DanRod1999 this is still failing

PR #711, opened 2024-04-09
Add api to decompress be_Prop binary data field from changeset

khanaffan:


iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -31,6 +31,44 @@ describe("basic tests", () => {
...
+    iModelDb.saveFileProperty({ namespace: "test", name: "test" }, "hello", propData);
+    iModelDb.saveChanges();
+    const stmt = new iModelJsNative.SqliteStatement();
+    stmt.prepare(iModelDb, "SELECT [RawSize], [Data] FROM [be_Prop] WHERE [NameSpace] = 'test' AND [Name] = 'test'");
+    assert.equal(stmt.step(), DbResult.BE_SQLITE_ROW);
+    const rawSize = stmt.getValueInteger(0);
+    assert.notEqual(rawSize, 0);
+    const blob = stmt.getValueBlob(1);
+    stmt.dispose();
+    const propDecompressData = iModelJsNative.DgnDb.zlibDecompress(blob, rawSize);
```
nick4598, line 66: Based off of this test, I'm assuming that saveFileProperty itself compresses the propData passed to it? Is that correct? How would a user know that this is the case? Feels like we should document that on saveFileProperty in itwinjs-core even if it is internal.

khanaffan: Yes, and it's conditionally compressed by checking if compression increases the size of the buffer. If it does compress it sets another property called rawSize.

The addon method I added is not supposed to be called by the user. It's meant to be used by internal code in the conflict handler to record the result before/after value for conflict and its resolution in be_prop table. At the end of pull merge application can enumerate what the conflicts were detected and what was the decision taken by default and that app can then decide to do nothing or override the default resolution picking one of the old/new values and calling db.SaveFileProperty(). Note be_prop has no lock so two user can change it at the same time.

nick4598 (approved):

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -31,6 +31,44 @@ describe("basic tests", () => {
...
   });
 
+  it("compress/decompress", () => {
+    const assertCompressAndThenDecompress = (sourceData: Uint8Array) => {
+      const compressData = iModelJsNative.DgnDb.zlibCompress(sourceData);
+      const decompressData = iModelJsNative.DgnDb.zlibDecompress(compressData, sourceData.length);
+      assert.deepEqual(sourceData, decompressData);
+    };
+    const randomData = (n: number) => {
+      return Uint8Array.from(Array.from({ length: n }, () => Math.floor(Math.random() * 10)));
```
pmconne, line 41: Please don't use random data in tests.

khanaffan: Yes, I thought about it and agree not to use random numbers. I had difficulty compressing a small array as it resulted in a bigger buffer than the source buffer and did not want to paste a large int array in test code. So I used random but generated a less dense array as random numbers are just between 0-10 and not 0-255. This allows data to be always compressed significantly as its zlib which uses Huffman compression (assigning small code to more frequent characters)

But now that I think it could be just an array of single numbers [1,1,1...] which will be even better as the buffer can be small and the encoder will just assign a bit to each digit.

Let me fix it.

nick4598: Out of curiosity, what is the concern with random data? Some permutation of data might cause problems?

pmconne: Tests should be deterministic. Random data introduces potential flakiness and even if flakiness is impossible, it's better to be explicit about your inputs and expected outputs. It's fine to generate a random data set ahead of time and then hard-code that as your test data - just don't generate new random data each time the test runs.

khanaffan: Just to avoid pasting the 1K array into the test file. Any compression algorithm poorly compresses a randomized buffer as it takes advantage of patterns in data. With no repeating pattern, it cannot compress. So when I use a randomized buffer it has to be 1K to be compressed a little bit. But as I said I can just have to repeat data for it to be compressible.

pmconne (approved):

PR #731, opened 2024-04-22
Fix default ECDb schema version numbers for profile 4005

rschili:
Was an oversight during the recent version increment.


pmconne (approved):
I don't suppose there's a unit test that could be written to catch such oversights, right?

rschili:
> I don't suppose there's a unit test that could be written to catch such oversights, right?

In this case it's tricky. The code goes through LocateSchema, which returns the latest compatible. So if 2.0.4 is available, it doesn't matter if we try to locate 2.0.3 or 2.0.4, the result is the same. There are tests that ensure 2.0.4 is being used eventually. We could potentially make the locater use the explicit version, but that would just create a new source of error with no real advantage.

PR #741, opened 2024-04-29
ConcurrentQuery ignored QueryQuota specified by the client

atharvadeshpande14:
Fixes https://github.com/iTwin/itwinjs-core/issues/6610


iModelJsNodeAddon/api_package/ts/src/test/ECDb.test.ts
```diff
@@ -126,6 +126,51 @@ describe("concurrent query tests", () => {
...
+      kind: DbRequestKind.ECSql,
+      query: "with cnt(x) as (values(0) union select x+1 from cnt where x < 1000 ) select x, CAST(randomblob(1000) AS BINARY) from cnt",
+      quota: {
+        memory: 85464865465,
+        time: 6198,
+      },
+    } as DbQueryRequest);
+
+    expect(rc3.stats.memLimit).eq(0x800000);
+    expect(rc3.stats.timeLimit).eq(60000);
```
khanaffan, line 171: it seems if one of the quota (memory/time) exceeds, we reset both. Maybe we should reset only one that exceeded.

atharvadeshpande14: I have confirmed that for individual quota exceeding, only that particular value is reset to globalQuota. I have added tests for same.

ColinKerr (approved):

khanaffan (changes requested):

khanaffan (approved):

PR #744, opened 2024-05-02
Enhancements to SchemaSync for iModel Hub Forking, Initialization, and Profile Upgrades

khanaffan:
### IModel Hub initializing & forking usecase
1. `SyncId` now mean `ContainerId` which of type string (instead of BeGuid) and must be provided by caller when initializing schema sync container. Previously it was random Guid that was problem when forking iModel from specific changeset.
2. For any given iModel, a schema sync container can be reinitialized with a new container. In this scenario, SchemaSync will allocate a new `containerId` to `SyncId` and will also transfer the `DataVer` from the iModel to the new container.

### Profile upgrades
When an iModel undergoes a profile upgrade, it is now also pushed to the schema sync container, whereas previously it was not.
1. Under the new rule profile, versions of schema sync containers and iModels must adhere to the following rules.
   * **On pull**: The iModel ECDb profile version should be less than or equal to the schema sync container ECDb profile version.
   * **On push**: The iModel ECDb profile version should be greater than or equal to the schema sync container ECDb profile version.
2. The `IModelDb.upgradeSchemas()` function in iTwin ensures that the profile upgrade is distributed to the users.

### Other important changes
1. When merging local transactions, we now include all DDLs, which was not the case previously for some reason. We only included first DDL txn and then ignore rest.
2. When applying a changeset, apply 'ec_*' changes before data changes, and after applying 'ec_*' changes, synchronize the database schema to ensure that subsequent data can be inserted correctly. This is necessary when the schema and data are in the same changeset.
3. In the past, only changesets with DDL were internally recognized as schema changesets when applying a changeset. However, changes to 'ec_*' tables without DDL were not considered as schema changesets. This meant that schema caches were sometimes not cleared when there were no DDL changes. Now, we use ChangesetType to ensure that both EC and DDL changes are treated as schema changes.
### Changes

- [x] Added `Db::GetTotalModifiedRowCount64()`. This is use by SchemaSync to see if `SchemaSync::Pull()`. or `SchemaSync::Push()` changes any thing at all.
- [x] Added `DDLChanges::GetDDls()`. This is used by `TxnManager::StartCreateChangeset()` to merge DDLs. Since locally there can be more then one DDL changeset as we sync and import schemas.
- [x] Remove `ChangeStream::ApplyChanges()` paramter `ignoreNoop`. Its now set by default.
- [x] Added proxy ECDbImpl::ExecuteDDL() method for `Db::ExecuteDdl()`. This is to able to disable DDL tracking.
- [x] `ECDb::_AfterSchemaChangeSetApplied()` disable DDL tracking.
- [x] Fix spelling errors in code.
- [x] `ProfileManager::GetProfileVersion()` automatically load version if not loaded before.
- [x] `SyncId` changed form `BeGuid` -> `Utf8String`
- [x] SchemaSync will update container profile version.
- [x] SchemaSycn will check profile version before synchronizing to ensure we do not mess data up.
- [x] Allow overrding SchemaSync container for a iModel.
- [x] DataVersion for SchemaSync is updated at the end of sync operation to esnure we do not update it an fail later.
- [x] Added addon message `DgnDb.schemaSyncPush()`
- [x] C++ changeset conflict handler updated to match typescript one. 
  - [x] On conflict for ec_* table of type DATA or PRIMARY KEY we skip.
  - [x] Replace be_Prop row when `namespace=ec_Db` and `name=localDbInfo`.
- [x] `DgnDb::SchemaSyncPull()` should mark current session txn as schema changeset if `DgnDb::SchemaSyncPull()` modified the schema.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6692


iModelCore/ECDb/ECDb/ECDbImpl.cpp
```diff
@@ -233,7 +238,18 @@ void ECDb::Impl::OnInit() const
     RegisterECSqlPragmas();
     }
 
+//--------------------------------------------------------------------------------------
+// @bsimethod
+//---------------+---------------+---------------+---------------+---------------+------
+DbResult ECDb::Impl::ExecuteDDL(Utf8CP ddl) const {
```
rschili, line 244: Seems a very counter intuitive place to put this logic. I was wondering why so many calls changes from ECDb::ExecuteDdl() to this method. Why not bake this logic directly into the ExecuteDdl() method?

There is no guarantee callers will use this instead of the method on ECDb. Actually, it's more likely people will directly call the other method and not this one. It also seems strange to directly call an Impl method which in turn will call back through the public ECDb interface again...

pmconne: I am curious about this as well @khanaffan.

pmconne (approved):

iModelCore/ECDb/ECDb/ECDbImpl.cpp
khanaffan: _I thought i answered this few days back but I guess I did not pressed comment._

The objective is to utilize the metadata stored in the ec_* table to update the database schema once a schema changeset has been applied and schema synchronization is active. The solution I've implemented involves disabling DDL tracking during the database schema synchronization post-application of a schema changeset.

As a principle, change tracking should never be disabled unless applying an incoming changeset. This is precisely why the current change tracker in BeSQlite::Db is not accessible via a function. It can only be enabled or disabled within the TxnManager, which acts as a change tracker itself. Therefore, we should not introduce a method in Db to disable it, preventing any assumptions by programmers that it can be disabled at will.

My implementation aims to restrict access to ECDb for a very specific case, and even within ECDb, tracking cannot be disabled unless the `DisableDdlTracking` scope is used on the stack. Therefore, my main goal is to narrow the circumstances under which DDL tracking can be disabled, ensuring that any attempt to do so will be evident during a peer review.

rschili (approved):

ColinKerr (approved):

PR #748, opened 2024-05-03
Allow App to specify GCS asset Dir

DanRod1999:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6825
resolves: https://github.com/iTwin/itwinjs-backlog/issues/1134

Add logging for GCS when workspaces are disabled, also add log for time taken to collect gcs data.

Be able to specify the location for GCS data in the native code when gcs workspaces are disabled. Visualization/GPB want to use it when loading up GPB in the kubernetes clusters. This is related to one of the experiments during the pineapple viewer hackathon.


iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -8946,7 +8946,7 @@ struct GeoCoordWorkspaces {
             Logging::LogMessageV("GeoCoord", LOG_ERROR, "Unable to read data for GCS file %s from workspace %s, rc=%d", path, row.db->GetDbFileName(), rc);
             delete resource; // weren't able to read from row.
         }
-        Logging::LogMessageV("GeoCoord", LOG_WARNING, "Unable to find GCS file %s in any workspace", path);
+        Logging::LogMessageV("GeoCoord", LOG_INFO, "Unable to find GCS file %s in any workspace", path);
```
DanRod1999, line 8949: Should this log messaged be removed entirely in favor of the later message stating if the file was not found in a workspace or locally?

nick4598: A bit nitpicky but, when we log 'in workspace or locally' are we actually searching locally always? The log message implies that we checked both, but I imagine we could fail to have it in our workspace and then we'd log 'in workspace or locally' without searching locally right?

DanRod1999: In the case of that particular log message, we are checking both. It failed to find anything in a workspace, so it looks locally, if it finds it in neither than that msg is logged

nick4598: Sounds good,

Regarding your original question: Should this log messaged be removed entirely in favor of the later message stating if the file was not found in a workspace or locally?

We should keep the log message but change warning to info and vice versa based off of if loadLocalFiles is true. If loadlocalFiles is false we'd log nothing at all if we removed the message.

nick4598 (approved):

wgoehrig (approved):

pmconne:
> Visualization/GPB want to use it when loading up GPB in the kubernetes clusters. This is related to one of the experiments during the pineapple viewer hackothon.

The `Initialize` function takes a data directory. Silently overriding that input with an undocumented environment variable is gross and surprising. Did you consider permitting the application to configure `PlatformLib::IKNownLocationsAdmin::_PGetGeoCoordinateDataDirectory`?

kabentley:
I disagree with the whole concept of reading local data files. That capability should be removed entirely - it is not necessary and will only cause problems. 

If somebody *really* wants to deliver GCS data with an application (I still don't understand why you can't use the cloud workspaces), they should use the existing api and add a local WorkspaceDb.

PLEASE DON'T DO THIS!

PR #751, opened 2024-05-07
Backport ECDb profile 4.0.0.5 to release/3.x

rschili:
Still work in progress, lots of merge problems during cherry pick.


pmconne:
Can you edit the description to describe the motivation for updating 3.x this late in its maintenance cycle? It reaches end of life in 2 weeks.

ColinKerr:
> Can you edit the description to describe the motivation for updating 3.x this late in its maintenance cycle? It reaches end of life in 2 weeks.

Our policy is to push profile upgrades back to the oldest supported version.  There is an update to BisCore that uses this profile update and there are applications that are waiting on the features enabled by the profile + BisCore update.  

The goal of doing this update so late in cycle is to increase the number of apps that can easily support the new profile.  It'd be nice if they were all updated to 4.x but we know some are not and we do not want that to delay the use of the new features enabled by this profile.

rschili:
After discussion with Colin, we will not do this. It seems too risky so close to end of life of the branch.

PR #753, opened 2024-05-07
Add function findElementClassId

jiaruiz717:
as [issue 173](https://github.com/iTwin/imodel-transformer/issues/173), we need to export findRemappedElementClass, this PR is for the corresponding part on native side


khanaffan:
You do not need a function for it. You can do it with ECSQL

```
iModelConsole> SELECT FORMAT('0x%x', ec_classId('BisCore.Element'));

FORMAT('0x2f',ec_classId('BisCore.Element'))
--------------------------------------------
0x50
```

nick4598:
> You do not need a function for it. You can do it with ECSQL
> 
> ```
> iModelConsole> SELECT FORMAT('0x%x', ec_classId('BisCore.Element'));
> 
> FORMAT('0x2f',ec_classId('BisCore.Element'))
> --------------------------------------------
> 0x50
> ```

Would this work in the cases where transformations define their own remappings from sourceClassFullName to targetClassFullName?

The most recent example we saw was with an iModel which had some linearDraft changesets which introduce a lot of unnecessary schemas.

The source iModel would have a schema like IUQYJHLAL:WireSupport, which would be remapped to the target iModel ConsolidatedWireSupport:WireSupport.

khanaffan:
> > You do not need a function for it. You can do it with ECSQL
> > ```
> > iModelConsole> SELECT FORMAT('0x%x', ec_classId('BisCore.Element'));
> > 
> > FORMAT('0x2f',ec_classId('BisCore.Element'))
> > --------------------------------------------
> > 0x50
> > ```
> 
> Would this work in the cases where transformations define their own remappings from sourceClassFullName to targetClassFullName?
> 
> The most recent example we saw was with an iModel which had some linearDraft changesets which introduce a lot of unnecessary schemas.
> 
> The source iModel would have a schema like IUQYJHLAL:WireSupport, which would be remapped to the target iModel's schema ConsolidatedWireSupport:WireSupport.

No it will not do that. I meant if you want to convert className to id from typescript side. So if you have map for className available on typescript side then you can do the conversion their instead of adding native method. Our goal should be bringing more logic into typescript when possible and slim out native layer.

nick4598:
> No it will not do that. I meant if you want to convert className to id from typescript side. So if you have map for className available on typescript side then you can do the conversion their instead of adding native method. Our goal should be bringing more logic into typescript when possible and slim out native layer.

The map for className available on the typescript side doesn't exist. Right now it goes directly to the native side by calling 'transformer.context.remapElementClass'. Native side converts the classFullName into a classId.

I suppose we could add a method to transformer in TS `transformer.remapElementClass` which maintains its own map and also calls `transformer.context.remapElementClass` so that cloneElement continues to work properly in native. We'd just have to ask services team to switch to calling `transformer.remapElementClass`. Do you view this as a better option?

khanaffan:
> > No it will not do that. I meant if you want to convert className to id from typescript side. So if you have map for className available on typescript side then you can do the conversion their instead of adding native method. Our goal should be bringing more logic into typescript when possible and slim out native layer.
> 
> The map for className available on the typescript side doesn't exist. Right now it goes directly to the native side by calling 'transformer.context.remapElementClass'. Native side converts the classFullName into a classId.
> 
> I suppose we could add a method to transformer in TS `transformer.remapElementClass` which maintains its own map and also calls `transformer.context.remapElementClass` so that cloneElement continues to work properly in native. We'd just have to ask services team to switch to calling `transformer.remapElementClass`. Do you view this as a better option?

make sense. Your remap is in context on native side.

khanaffan (approved):

PR #757, opened 2024-05-14
Fix PolyfaceAuxData index blocking flatbuffer format

dassaf4:
Addresses https://github.com/iTwin/itwinjs-core/issues/6583.

TypeScript FlatBuffer serialization of `PolyfaceAuxData` indices lacked the 0-blocking of the other `Polyface` index arrays, which led to crashes during mesh processing. Now TypeScript FlatBuffer serialization outputs the correct index blocking, and both TypeScript and native FlatBuffer APIs read both old and new PolyfaceAuxData index formats. The old format is detected using heuristics.

Although there is a similar [TypeScript PR](https://github.com/iTwin/itwinjs-core/pull/6726), the fixes in each repo are independent of the other.

Also:
* Consolidate code in native FlatBuffer mesh writers. Ensure only meshes with indices are written (all FlatBuffer readers require them). Non-indexed mesh styles are converted to variable-sized index blocking first, like the native JSON mesh writer does.
* New t_crossPlatform.cpp verifies that native API can read TypeScript API geometry serializations. (The same test exists in iTwinjs to verify the TypeScript API can read native API geometry serializations.) Each test case consists of at least four files: FlatBuffer and JSON serializations of a single geometry, written by both APIs. The test verifies that a given API deserializes each file in a test case into the same geometry. Initial test cases include `Polyface` variations in `numPerFace` value, `PolyfaceAuxData` format, and all combinations thereof.
* `PolyfaceHeader::CloneAsVariableSizeIndexed` didn't actually clone the instance. Deprecate it in favor of an overload that does.
* New `PolyfaceHeader::CreateIndexedMeshSwap` avoids copying the input arrays for efficiency.
* Document that `PolyfaceHeader::ClearTags` sets `m_twoSided = true`, contrary to other mesh initializers. This side effect was discovered by the new cross-platform tests. Can't change it for fear of breaking Civil users. Work around it in known callers.
* New "Polyface.SphereMeshSanity" test verifies that the standard sphere mesh has unique facets and outward normals.
* Fix a few typos in the 1220geomodeler JSON test files discovered by stricter mesh index checking.
* Improve documentation for `PolyfaceAuxData`. In particular emphasize that the indices must have the same blocking as the other `Polyface` index arrays.
* Constify GTestFileOps and add FlatBuffer file helpers.


iModelCore/GeomLibs/geom/test/PolyfaceTest/t_polyface.cpp
```diff
@@ -4604,6 +4605,126 @@ TEST (PolyfaceQuery, FacetOrientation)
...
+        if (y0 > y1)
+            return false;
+        }
+    if (!DoubleOps::WithinTolerance(z0, z1, tol))
+        {
+        if (z0 < z1)
+            return true;
+        if (z0 > z1)
+            return false;
+        }
```
saeeedtorabi, line 4633:
```suggestion
    if (!DoubleOps::WithinTolerance(x0, x1, tol))
        return x0 < x1;
    if (!DoubleOps::WithinTolerance(y0, y1, tol))
       return y0 < y1;
    if (!DoubleOps::WithinTolerance(z0, z1, tol))
        return z0 < z1;
```

saeeedtorabi: Also what do you return if x0 == x1?

dassaf4: > Also what do you return if x0 == x1?

false. But that's a good simplification. Thanks!

Issue #750, opened 2024-05-06
Presentation.getManager().getContent() hangs backend after db edit operations


arnoldas-v:
**Describe the bug**
Sometimes after create/update operations `Presentation.getManager().getContent()` call starts responding very slowly or hangs completely. As a result application becomes unusable (no properties can be loaded or no other further edit operations can be started).

**To Reproduce**

Reproducing this issue is very difficult. For normal use it appears seemingly randomly and not on every imodel.
In our application we have a sample tool that would try to force this issue by looping these operations for a few times on specific imodel:
-createElements, saveChanges
-Presentation.getManager().getContent()
-updateElements, saveChanges
-Presentation.getManager().getContent()
-deleteElements, saveChagnes

Every call to getContent() might take multiple minutes or never complete at all.
saveChanges call is very important because without it the issue never happens.

Will provide dumps/steps to build the app and reproduce as needed.

**Expected behavior**
PresentationManager always responds in a reasonable amount of time.

**Screenshots**
![image](https://github.com/iTwin/itwinjs-core/assets/81693692/dfdff0ec-f2d7-4aef-9305-ee083b2ab4fa)
Properties widget timed out and crashed.

**Desktop (please complete the applicable information):**

- OS: windows11
- Electron
- Version 25.8.1
- iTwin.js Version 4.3.3

**Additional context**
Interesting observation:
I noticed that during this hang the whole electron backend becomes not responsive.
For example if you have:
```ts
const timeout () => new Promise(resolve => setTimeout(() => resolve("timed-out"), 1000 * 60 * 1));
await Promise.race([Presentation.getManager().getContent(), timeout()]);
```
The Promise.race will never resolve. Indicating that native does not even give back control to node/electron somehow?

grigasp:
Deadlock call stack:
![image](https://github.com/iTwin/imodel-native/assets/35135765/e8b351a2-791e-4451-babb-1b53fa2995b8)

grigasp:
The fix was released with `itwinjs-core` packages `@4.6.1`

GintV:
@grigasp We're still able to reproduce this issue with Core 4.6.1

grigasp:
> @grigasp We're still able to reproduce this issue with Core 4.6.1

Are you seeing a hang or a slowdown with this version?

GintV:
We got a hang for 10 min for the whole app. After that, properties grid stopped loading properties and we started to get BackendTimeout continuously

[...]

PR #758, opened 2024-05-16
Fixes to avoid getting nodes cache into invalid state

saskliutas:
https://github.com/iTwin/imodel-native/issues/750 issue is difficult to repeatedly reproduce but managed to spot some problems that could lead to backend hang when nodes cache ends up in invalid state. Made some changes to avoid that:

- Always check for locked children levels when locking hierarchy level. Managed to end up in a situation where parent and child hierarchy levels were lock and updated at the same time. This change also allowed to remove additional locks held during hierarchy update after iModel change.
- Moved hierarchy level caching in `SameLabelGroupingNodesPostProcessorDeprecated` to be done in a transaction after lock is aquired. Also added additional check before caching hierarchy level and data source in case they were cached while checking for child nodes.


iModelCore/ECPresentation/Source/UpdateHandler.cpp
```diff
@@ -596,6 +557,7 @@ struct HierarchyRefresher
 
         AffectedHierarchyLevelIdentifier grandparentLevel(parentHierarchyLevelIdentifier, grandParent.IsValid() ? grandParent->GetKey().get() : nullptr);
 
+        newProvider.GetContext().GetHierarchyLevelLocker().Unlock();
         std::shared_ptr<IHierarchyLevelLocker> grandParentHierarchyLock = m_nodesCache->CreateHierarchyLevelLocker(grandparentLevel.GetHierarchyLevelIdentifier(), true);
```
grigasp, line 561: Do we really need to lock here? It looks like a lock gets created by `RefreshHierarchyLevel` -> `RefreshNavNodesProvider`

iModelCore/ECPresentation/Source/Hierarchies/NavNodeProviders.cpp
```diff
@@ -3538,10 +3558,9 @@ NavNodesProviderPtr SameLabelGroupingNodesPostProcessorDeprecated::_PostProcess(
 
     // attempt to find cached merged nodes provider - success means the whole hierarchy level is already post-processed and in cache.
     // it's more efficient to use the cached version compared to loading and merging everything again, so just return the cached provider.
-    DataSourceIdentifier mergedDatasourceIdentifier(GetHierarchyLevelIdentifier(*context).GetId(), {}, context->GetInstanceFilterPtr());
-    mergedDatasourceIdentifier.SetResultSetSizeLimit(context->GetResultSetSizeLimit());
-    DataSourceInfo mergedDatasourceInfo = context->GetNodesCache().FindDataSource(mergedDatasourceIdentifier, context->GetRulesetVariables());
-    if (mergedDatasourceInfo.GetIdentifier().IsValid())
+    HierarchyLevelIdentifier hlIdentifier = GetHierarchyLevelIdentifier(*context);
+    DataSourceInfo mergedDatasourceInfo = GetMergedDataSourceInfo(*context, hlIdentifier.GetId());
+    if (mergedDatasourceInfo.GetIdentifier().IsValid() && mergedDatasourceInfo.IsInitialized())
```
grigasp, line 3563: Is it possible that `mergedDatasourceInfo.IsInitialized()` returns `true`, but `mergedDatasourceInfo.GetIdentifier().IsValid()` - not? Maybe only one check is enough?

iModelCore/ECPresentation/Source/UpdateHandler.cpp
saskliutas: We don't need. Removed

iModelCore/ECPresentation/Source/Hierarchies/NavNodeProviders.cpp
saskliutas: Yes, it's enough to check `IsInitialized()`

grigasp (approved):

grigasp:

PR #759, opened 2024-05-16
Backport latest ECDb code to release/3.x

rschili:
Experimental, not sure yet, if this is going to work. WIP


rschili:
After discussion with Colin, we will not do this. It seems too risky.

PR #761, opened 2024-05-16
CSMAP Read-only directory fix for ios

tcobbs-bentley:
Fix was suggested by Alain Robert


aruniverse (approved):

aruniverse:

AlainRobertAtBentley:
I have made all tests run correctly on a local build. The failing tests must be related to something in the pipeline. I also noted the presence of a few windows slashes in tests paths. I think these should be changed from \\ to / specifically at:https://github.com/iTwin/imodel-native/blob/64949e58b45b056e5943bbf7fe101745634dc641/iModelCore/GeoCoord/Tests/NonPublished/GeoCoordTestCommon.cpp#L31

tcobbs-bentley:
> I have made all tests run correctly on a local build. The failing tests must be related to something in the pipeline. I also noted the presence of a few windows slashes in tests paths. I think these should be changed from \ to / specifically at:
> 
> https://github.com/iTwin/imodel-native/blob/64949e58b45b056e5943bbf7fe101745634dc641/iModelCore/GeoCoord/Tests/NonPublished/GeoCoordTestCommon.cpp#L31

As far as I can tell, BeFileName automatically fixes paths to be correct on Windows and non-Windows. (On Windows, it converts all forward slashes to backslashes, and on non-Windows, it converts all backslashes to forward slashes.) There are four total places in the GeoCoord C++ code that use `Assets\\`, and none that use `Assets/`, so it seems like leaving it alone is the best thing here.

tcobbs-bentley:
Replaced with https://github.com/iTwin/imodel-native/pull/946

PR #771, opened 2024-05-24
Enable ICU by default on new IModels

khanaffan:
New IModel will have Latin1 by default. Older IModel open with new code will still use ASCII.


rschili (approved):
Approving in Draft mode so we can merge after coordination with hub as mentioned in previous comment

PR #773, opened 2024-05-29
enable debug logging for cleanDeletedBlocks

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6771

Using an options object for cleanDeletedBlocks. May in the future also add a callback option as part of discussions with SQLite team: https://sqlite.org/cloudsqlite/forumpost/a3847355a0.

Main motivatgion for this PR is to allow 'debugLogging' which gives us an idea of how long it takes to build a hash table for the all the blocks in the manifest and how long it takes to iterate all the contents of the container in blob storage. This is to potentially support / go against the argument that we need to be able to interrupt the hash table building and the container contents listing in order to give the write lock to more important operations such as checkpoint creation.


iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -471,10 +471,18 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
 
     Napi::Value CleanDeletedBlocks(NapiInfoCR info) {
         RequireWriteLock();
-        OPTIONAL_ARGUMENT_INTEGER(0, nSeconds, (60*60)); // default to one hour
+
+        bool debugLogging = false;
+        int nSeconds = 3600; // default is 1 hour.
+        if (info[0].IsObject()) {
```
pmconne, line 477: If you want to allow an object as an argument, can you update the TS function signature?

nick4598: Sorry.. I forgot to push that change

pmconne (approved):

PR #781, opened 2024-06-11
Version compare change summary performance improvements

diegopinate:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6826

Changes to fix service case regarding high memory usage for 400 MB changeset. Summary of changes:

1. Allow skipping bounding box calculation and loading elements altogether.
2. Filter out relationship instances as those are handled by path traversal.
3. Simplify passing options.

Bounding box calculation for changed models was never used in version compare. The original reasoning behind it was to improve performance of viewed iModels by leveraging the volumes that contain changes and limit the tiles downloaded. However, as this feature was never leveraged by the frontend, and is just a waste of computation, allow it to be turned off via an option.

Relationships are also not necessary to be processed as part of the initial instance processing, so lower amount of cached data by simply skipping them.


pmconne (changes requested):
You updated itwinjs-core tests to always pass true for the new option but didn't add any tests to verify what happens when you pass false.

diegopinate:
> You updated itwinjs-core tests to always pass true for the new option but didn't add any tests to verify what happens when you pass false.

Good catch, thanks @pmconne, I added a new test to itwinjs-core that test both variations of wantBoundingBoxes

pmconne (approved):

diegopinate:
@mergiyio backport release/4.x

PR #782, opened 2024-06-12
[Experimental] Add getInstance() to NativeDgnDb

khanaffan:
## Performance of getElement() vs getInstance()
`getInstance()` use `InstanceReader` which does not use ECSqlStatement and have far less moving parts (_it does not do any prepare per class. it just prepare one sqlite stmt per table)_ then when using ECSqlStatement. InstanceReader also get called when we run `SELECT $ FROM BisCore.Element`. 

1. DateTime::ToString() now use similar alog as sqlite which is very performant. This function as hot spot in profiler when generating JSON.
2. Added ExecuteSql() to help run pragmas for testing quickly without preparing and holding statements.
3. Added GetInstance() allowing bulk read of ecinstances from iModel.

% change : $`(V_2 - V_1) / |V_1| * 100 = ?`$

Screenshot 2024-06-25 155152


iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -17,6 +17,42 @@ import { copyFile, dbFileName, getAssetsDir, getOutputDir, iModelJsNative } from
 if (os.platform() === "linux")
   process.env.LINUX_MINIDUMP_ENABLED = "yes";
 
+describe.only("performance tests", () => {
```
6ar8nas, line 20:
```suggestion
describe("performance tests", () => {

```

6ar8nas:
Allowing to pass on `InstanceReader` options would seem to be greatly beneficial.

iModelCore/Bentley/Bentley/DateTimeConverter.cpp
```diff
@@ -369,6 +369,136 @@ BentleyStatus DateTimeConverter::ComputeLocalTimezoneOffsetFromUtcTime(int64_t&
 std::regex* DateTimeStringConverter::s_dtRegex = nullptr;
 std::regex* DateTimeStringConverter::s_todRegex = nullptr;
 
+//---------------------------------------------------------------------------------------
```
ColinKerr, line 372: We should have tests comparing these new methods to the existing printf methods

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
```diff
@@ -12,11 +12,84 @@ import * as path from "path";
 import { openDgnDb } from ".";
 import { IModelJsNative, SchemaWriteStatus } from "../NativeLibrary";
 import { copyFile, dbFileName, getAssetsDir, getOutputDir, iModelJsNative } from "./utils";
+import { Suite } from "mocha";
 
 // Crash reporting on linux is gated by the presence of this env variable.
 if (os.platform() === "linux")
   process.env.LINUX_MINIDUMP_ENABLED = "yes";
 
+describe.skip("performance tests", function (this: Suite) {
```
ColinKerr, line 21: we should have some proper tests and not these local only perf test

iModelCore/Bentley/Bentley/DateTimeConverter.cpp
khanaffan: yes there is a lot of test. It was written by Krischan so well tested.

iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts
khanaffan: Yes we should have that but not in this repo.

pmconne:
> Azure Pipelines failed to run 1 pipeline(s).

@DanRod1999 any idea what's wrong?

DanRod1999:
> > Azure Pipelines failed to run 1 pipeline(s).
> 
> @DanRod1999 any idea what's wrong?
Looks like an error in the yml file. I'll take a look

DanRod1999:
found the issue caused by a recent change in imodel-native-internal. Will fix now

ColinKerr:
where are the performance tests used to get the table comparing performance?

khanaffan:
> where are the performance tests used to get the table comparing performance?

Its out of here and likely to go some where else. I was thinking more of itwin-cli where we can write performance test and run it against a given dataset (files in directory) and create standard report out of it that can be compare over runs.

rschili (approved):

pmconne (approved):

PR #783, opened 2024-06-13
Presentation: Fix duplicated related property fields 

grigasp:
Fix duplicated related property fields when input contains multiple classes and related property spec has an instance filter. This fixes https://github.com/iTwin/viewer-components-react/issues/936.


saskliutas (approved):

grigasp:

PR #787, opened 2024-06-13
Allow a callback to be passed to cleanDeletedBlocks making it interruptible.

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6842

The motivation for this change is to allow the hub team's deletion jobs to be interrupted quickly if they detect that a creation job needs to be run on the same iModelId.


khanaffan (approved):

pmconne:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1127

PR #788, opened 2024-06-14
Presentation: Fix missing related property values

grigasp:
A fix for https://github.com/iTwin/viewer-components-react/issues/929.

We were skipping the load of nested related properties, because their field's source classes set was wrongly pointing to the root source classes rather than source classes of the field. Since we have no cheap way of knowing those classes for nested related fields, I changed the set to be nullable and made sure they're not set if we nest those fields.


saskliutas (approved):

grigasp:

PR #796, opened 2024-06-25
Fix Comment regarding Workspace Priority Param

DanRod1999:
I believe this comment is incorrect unless I'm miss understanding it. When working with GCS workspaces using 0 does give it the lowest priority, as in workspaces with greater numbers are used first.


pmconne:
The [schema](https://github.com/iTwin/itwinjs-core/blob/19871b33d7b16ca7a299da5c2b39a6b5d783f353/core/backend/src/assets/Settings/Schemas/Gcs.schema.json#L16) agrees with the original comment - GCS WorkspaceDbs with higher priority values are supposed to be searched first.

pmconne:
I misread the comment. It is incorrect, the implementation+schema are correct.

pmconne (approved):

PR #799, opened 2024-07-01
Add an extra version check for importing of modified schemas

6ar8nas:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6919

Closes iTwin/itwinjs-backlog#1101


iModelCore/ECDb/ECDb/SchemaWriter.cpp
```diff
@@ -286,6 +286,15 @@ BentleyStatus SchemaWriter::ImportSchema(Context& ctx, ECN::ECSchemaCR ecSchema)
                 {
                 if (schema->GetName().Equals(schemaChange->GetChangeName()))
                     {
+                    if (schemaChange->VersionMinor().GetOld() == schemaChange->VersionMinor().GetNew() &&
```
rschili, line 289: We should be able to use SchemaKey::CompareByVersion or ::Matches here instead of putting our own check.

rschili (approved):

rschili (changes requested):
Please add one test on the native side, too

iModelCore/ECDb/ECDb/SchemaWriter.cpp
rschili: @ColinKerr please have a quick look. I am surprised the SchemaComparer detects this as "modified" I thought we had an optimization so we would not compare schemas if version numbers are identical, so we don't do unnecessary work.... But seems like we do compare them.

Are you okay changing the behavior in general (so ECDb refuses to import schemas with matching version number but different contents)?

6ar8nas: > We should be able to use SchemaKey::CompareByVersion or ::Matches here instead of putting our own check.

This is a `SchemaChange` class which doesn't expose SchemaKeys. However I tackled it how it's usually done given the needed circumstances.

6ar8nas:
Prerequisite: #810 (TestLogger is used here)

iModelCore/ECDb/ECDb/SchemaWriter.cpp
6ar8nas: > @ColinKerr please have a quick look. I am surprised the SchemaComparer detects this as "modified" I thought we had an optimization so we would not compare schemas if version numbers are identical, so we don't do unnecessary work.... But seems like we do compare them.
> 
> Are you okay changing the behavior in general (so ECDb refuses to import schemas with matching version number but different contents)?

We had discussed it over a call last week. It was decided to note the user of this error by logging an error, but still continue processing so that it doesn't disrupt any current workflows.

I also have noticed that if some major version bump requiring change is present, the schema import alreqdy fails, so some guards are already present - it would just alert about all of the expected minor increments as wel.

iModelCore/ECDb/Tests/NonPublished/SchemaManagerTests.cpp
```diff
@@ -3184,6 +3184,45 @@ TEST_F(SchemaManagerTests, DuplicateInMemorySchemaTest)
...
     }
 
+/*---------------------------------------------------------------------------------**//**
+* @bsitest
++---------------+---------------+---------------+---------------+---------------+------*/
+TEST_F(SchemaManagerTests, SchemaWithChangesButSameVersionTest)
+    {
+    Utf8CP originalSchemaXml =
+        "<?xml version='1.0' encoding='utf-8' ?>"
+        "<ECSchema schemaName='std' nameSpacePrefix='std' version='1.0' xmlns='http://www.bentley.com/schemas/Bentley.ECXML.3.0'>"
```
rschili, line 3194: Ad mentioned, please update to EC 3.2

rschili (approved):

PR #805, opened 2024-07-08
Presentation: Fix red squiggles when using Visual Studio 2022

Yato333:
Presentation headers are usually included as `<ECPresentation/...>`. Since `ECPresentation` folder is not currently under the include search path the IDE couldn't always find the correct path to the included file.
Added `$(SrcRoot)imodel-native\iModelCore` to include search path, where `ECPresentation` folder can be located.

This change should only affect the IDE behavior and shouldn't affect the actual build.


grigasp (approved):

Yato333:
This seems to help only when using Resharper

PR #808, opened 2024-07-11
Add new pragma function to purge orphaned link table relationships

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1097

### DIFF
--- a/iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
+++ b/iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
@@ -729,5 +729,63 @@ DbResult PragmaIntegrityCheck::Write(PragmaManager::RowSet& rowSet, ECDbCR ecdb,
 }
 
 
+//=======================================================================================
+// PurgeOrphanedRelationships
+//=======================================================================================
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+DbResult PragmaPurgeOrphanRelationships::Read(PragmaManager::RowSet& rowSet, ECDbCR ecdb, const PragmaVal& val, PragmaManager::OptionsMap const& options)
+	{
+	if (!isExperimentalFeatureAllowed(ecdb, options))
+		{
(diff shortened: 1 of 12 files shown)


iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
```diff
@@ -729,5 +729,63 @@ DbResult PragmaIntegrityCheck::Write(PragmaManager::RowSet& rowSet, ECDbCR ecdb,
...
+	if (!isExperimentalFeatureAllowed(ecdb, options))
+		{
+		ecdb.GetImpl().Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSQL, ECDbIssueId::ECDb_0729, "'PRAGMA %s' is an experimental feature and is disabled by default.", GetName().c_str());
+		return BE_SQLITE_ERROR;
+		}
+
+	rowSet = std::make_unique<StaticPragmaResult>(ecdb);
+	rowSet->FreezeSchemaChanges();
+
+	bmap<Utf8CP, bvector<Utf8String>> orphanRelationshipRows;
```
rschili, line 749: Why keep the instance IDs as strings? We use them in an IN(), and we don't know how many there may be. I think a virtual set would be the perfect fit for this (BindVirtualSet())

iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
```diff
@@ -729,5 +729,63 @@ DbResult PragmaIntegrityCheck::Write(PragmaManager::RowSet& rowSet, ECDbCR ecdb,
...
+		return true;
+		};
+
+	if (const auto check = IntegrityChecker(ecdb).CheckLinkTableFkIds(callBack); check != BE_SQLITE_OK)
+		return check;
+
+	for (const auto& [className, instanceIds] : orphanRelationshipRows)
+		{
+		ECSqlStatement stmt;
+		stmt.Prepare(ecdb, SqlPrintfString("delete from %s where ECInstanceId in (%s)", className, BeStringUtilities::Join(instanceIds, ",").c_str()).GetUtf8CP());
```
rschili, line 763: I like how it's using the existing code from IntegrityChecker, but given that the number of orphan rows is unknown... maybe it would be better to not hold all of them in memory and instead use a combined statement that just finds and deletes them in one swoop?

rschili (approved):

khanaffan (approved):

PR #811, opened 2024-07-15
Presentation: Show display label for nested content values

Yato333:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7003


grigasp (changes requested):
I'd like:
- at least one native integration test where related instances' content has a custom label,
- at least one full stack test in `itwinjs-core` which does the same.

iModelCore/ECPresentation/Source/Content/ContentSpecificationsHandler.cpp
```diff
@@ -51,7 +51,7 @@ int ContentSpecificationsHandler::GetDefaultContentFlags(Utf8CP displayType, Con
     if (DISPLAY_TYPES_EQUAL(ContentDisplayType::Grid, displayType))
         flags |= ENUM_FLAG(ContentFlags::ShowLabels);
     else if (DISPLAY_TYPES_EQUAL(ContentDisplayType::PropertyPane, displayType))
-        flags |= ENUM_FLAG(ContentFlags::MergeResults);
+        flags |= ENUM_FLAG(ContentFlags::MergeResults) | ENUM_FLAG(ContentFlags::ShowLabels);
```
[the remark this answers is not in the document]
Yato333: Removed this. Everything seems to work without it

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/Content_RelatedPropertiesTests.cpp
```diff
@@ -1300,6 +1300,11 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, LoadsXToManyRelatedInstance
             "MergedFieldNames": []
             }],
         "%s": [{
+            "DisplayLabel": {
```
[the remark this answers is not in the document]
Yato333: Still looking into this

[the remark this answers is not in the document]
Yato333: Fixed the issue by adding a display label column to the related content queries

Yato333:
- 1 full-stack test is failing from core. The test checks hierarchy node path, which contains hashes of nodes. The first node in the path contains its' query and since a change was made to display label clause (`IF(classId IS NULL, ...)`) the hash now differs. The snapshot can be simply overwritten.
- 2 tests were added from the native side (Is that enough?)
- 3 full-stack tests have been created from the core's side (all passing locally).
I suggest merging https://github.com/iTwin/itwinjs-core/pull/7003
then we can successfully merge this PR, and reenable the Core tests afterwards

iModelCore/ECPresentation/Source/Content/ContentQueryResultsReader.cpp
```diff
@@ -150,7 +150,8 @@ static void SkipValues(int& sqlColumnIndex, bvector<ContentDescriptor::Field*> c
             if (!contract.ShouldHandleRelatedContentField(relatedContentField))
                 continue; // don't increase sqlColumnIndex even for this field
 
-            SkipValues(sqlColumnIndex, relatedContentField.GetFields(), contract);
+            // Skip display label column 
+            SkipValues(++sqlColumnIndex, relatedContentField.GetFields(), contract);
```
grigasp, line 154:
```suggestion
            // Skip display label column 
            ++sqlColumnIndex;
            // Skip related content select fields
            SkipValues(sqlColumnIndex, relatedContentField.GetFields(), contract);
```

iModelCore/ECPresentation/Source/Serialization/DefaultECPresentationSerializer.cpp
```diff
@@ -607,7 +607,7 @@ rapidjson::Document DefaultECPresentationSerializer::_AsJson(ContextR ctx, Conte
             rapidjson::Value nestedValuesJson(rapidjson::kArrayType);
             for (auto const& nestedItem : nestedContentEntry.second)
                 {
-                static const int valueSerializationFlags = ContentSetItem::SERIALIZE_PrimaryKeys | ContentSetItem::SERIALIZE_Values | ContentSetItem::SERIALIZE_DisplayValues | ContentSetItem::SERIALIZE_MergedFieldNames;
+                static const int valueSerializationFlags = ContentSetItem::SERIALIZE_PrimaryKeys | ContentSetItem::SERIALIZE_Values | ContentSetItem::SERIALIZE_DisplayValues | ContentSetItem::SERIALIZE_MergedFieldNames | ContentSetItem::SERIALIZE_DisplayLabel;
```
grigasp, line 610: I don't see a similar change in `IModelJsECPresentationSerializer` - is it not necessary there?

iModelCore/ECPresentation/Source/Shared/Queries/QueryBuilderHelpers.cpp
```diff
@@ -963,48 +1000,36 @@ PresentationQueryContractFieldPtr QueryBuilderHelpers::CreateDisplayLabelField(U
...
+            instanceIdField,
+            defaultPropertyValueField,
+            PresentationQueryContract::CreateRelatedInstanceInfoField(relatedInstancePaths),
+            });
+        }, name, schemaHelper, selectClass, classIdField, instanceIdField, labelOverrideValueSpecs, labelRequestsStack);
+
+    // If class ID select clause results in NULL (i.e. when class ID is taken from a joined table with no overlaps), then label query might fail.
+    // NOTE: It's assumed that class ID field and instance ID fields are non-null constants or they're properties of the same instance that will both be defined or NULL.
+    // This should result into the following clause:
+    // IIF(classId IS NULL, CAST(NULL AS TEXT), ...)
```
grigasp, line 1025: I don't understand the described situation - could you provide the full ECSQL as an example?

iModelCore/ECPresentation/Source/Shared/Queries/QueryBuilderHelpers.cpp
```diff
@@ -963,48 +1000,36 @@ PresentationQueryContractFieldPtr QueryBuilderHelpers::CreateDisplayLabelField(U
...
+    static auto const SimpleField = [](Utf8String clause) { return PresentationQueryContractSimpleField::Create("", clause, false); };
```
grigasp, line 1026: I'd rename this to `CreateSimpleField`

iModelCore/ECPresentation/Source/Content/ContentQueryContracts.h
```diff
@@ -25,6 +26,8 @@ struct ContentQueryContract : PresentationQueryContract
     ECClassCP m_relationshipClass;
     Utf8String m_relationshipClassAlias;
     IQueryInfoProvider const& m_queryInfo;
+    ECSchemaHelper const& m_schemaHelper;
+    IRulesPreprocessor& m_rulesPreprocessor;
```
grigasp, line 30: What do you think about passing a lambda to get label override specs rather than schema helper and rules preprocessor? IMO, these two have way more than we need and I'd like contracts to have as little dependencies as possible.

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/Content_RelatedPropertiesTests.cpp
```diff
@@ -1396,53 +1466,104 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, LoadsXToManyRelatedInstance
...
-        <ECNavigationProperty propertyName="A" relationshipName="A_B" direction="Backward" />
+        <ECProperty propertyName="StringProperty" typeName="string" />
+    </ECStructClass>
+    <ECEntityClass typeName="ChildClass2">
+        <ECNavigationProperty propertyName="Parent" relationshipName="ParentHasChildren2" direction="Backward" />
+        <ECProperty propertyName="IntProperty" typeName="int" />
+        <ECProperty propertyName="StringProperty" typeName="string" />
+        <ECArrayProperty propertyName="ArrayProperty" typeName="int" />
+        <ECStructProperty propertyName="StructProperty" typeName="MyStruct" />
+        <ECStructArrayProperty propertyName="StructArrayProperty" typeName="MyStruct" />
```
grigasp, line 1483: do we need all these different kinds of properties if we're testing display labels in this test?

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/Content_RelatedPropertiesTests.cpp
```diff
@@ -1452,16 +1573,133 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, LoadsXToManyRelatedInstance
...
+                }
+            }]
+        })",
+        FIELD_NAME(childClass2, "Parent"), FIELD_NAME(childClass2, "IntProperty"), FIELD_NAME(childClass2, "StringProperty"),
+        FIELD_NAME(childClass2, "ArrayProperty"), FIELD_NAME(childClass2, "StructProperty"),
+        FIELD_NAME(childClass2, "StructArrayProperty")).c_str());
+    rapidjson::Document actualChildClass2FieldType = descriptor->GetVisibleFields()[2]->GetTypeDescription().AsJson();
+    EXPECT_EQ(expectedChildClass2FieldType, actualChildClass2FieldType)
+        << "Expected: \r\n" << BeRapidJsonUtilities::ToPrettyString(expectedChildClass2FieldType) << "\r\n"
+        << "Actual: \r\n" << BeRapidJsonUtilities::ToPrettyString(actualChildClass2FieldType);
```
grigasp, line 1702: do these type assertions have any virtue for this test?

iModelCore/ECPresentation/Tests/NonPublished/Unit/Content/ContentQueryResultsReaderTests.cpp
```diff
@@ -25,6 +25,12 @@ struct ContentQueryResultsReaderTests : QueryExecutorTests
         {
         return ContentDescriptor::Create(*m_connection, *m_ruleset, RulesetVariables(), *NavNodeKeyListContainer::Create(), ContentDisplayType::Undefined, contentFlags, contentFlags);
         }
+    
+    ContentQueryContractPtr CreateContract(uint64_t id, ContentDescriptorPtr descriptor, ECClassCP ecClass, ComplexQueryBuilderPtr queryInfo,
+        bool skipCompositePropertyFields = false, bool skipXToManyRelatedContentFields = false) const
+        {
+        return ContentQueryContract::Create(id, *descriptor, ecClass, *queryInfo, *m_schemaHelper, *m_rulesPreprocessor, nullptr, {}, skipCompositePropertyFields, skipXToManyRelatedContentFields);
+        }
```
grigasp, line 33: A function should prefer taking a reference rather than a pointer, if it expects the value to never be NULL. I suggest changing `descriptor` to `ContentDescriptorCR` and `queryInfo` to `ComplexQueryBuilderR`.

iModelCore/ECPresentation/Source/Shared/Queries/QueryBuilderHelpers.cpp
Yato333:
```sql
SELECT 1 AS [/ContractId/],
       GetInstanceKey(IFNULL([this].[ECClassId], 0), [this].[ECInstanceId]) AS [/ECInstanceKeys/],
       CAST(NULL AS TEXT) AS [/InputECInstanceKeys/],
       '[]' AS [/RelatedInstanceInfo/],
       GetInstanceKey(IFNULL([RelatedClass_1 ].[ECClassId], 0),
       [RelatedClass_1].[ECInstanceId]) AS [/key/RelatedClass_1],      
       GetECInstanceDisplayLabel(([RelatedClass_1].[ECClassId]), ([RelatedClass_1].[ECInstanceId]), '', '[]')) AS [/DisplayLabel/RelatedClass_1],

FROM bis.Element [this]
LEFT JOIN RelatedClass_1 ON ...
```

When computing value for `/DisplayLabel/RelatedClass_1`, it calls `GetECInstanceDisplayLabel`, which throws an error if 1 or 2 argument is null.

grigasp: maybe the `IIF(classId IS NULL, ...` should be placed on the call to `GetECInstanceDisplayLabel` function call then, rathen than the whole label field?

Yato333: DisplayLabel clause may consist of `COALESCE` with many fallback specifications. For example:
`COALSESCE(GetDisplayLabel(classId, instanceId, ...), [RelatedClass_1].[Name], [RelatedClass_1].[UserLabel], ...)`
I think, it is easier (and probably more efficient) not to invoke `COALESCE` at all if there are no related instances found than to ensure that each fallback specification can work with NULL.

iModelCore/ECPresentation/Source/Serialization/DefaultECPresentationSerializer.cpp
Yato333: You are right. Deeply nested content doesn't serialize properly. Fixed and added more tests from the core side

iModelCore/ECPresentation/Source/Shared/Queries/QueryBuilderHelpers.cpp
Yato333: Renamed

iModelCore/ECPresentation/Source/Content/ContentQueryContracts.h
Yato333: Changed to `std::function`

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/Content_RelatedPropertiesTests.cpp
Yato333: Removed

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/Content_RelatedPropertiesTests.cpp
Yato333: Removed

iModelCore/ECPresentation/Source/Shared/Queries/QueryBuilderHelpers.cpp
grigasp: In theory, some of the fallbacks in `COALESCE` could be static and not require the instance to exist to have a value. Now that I think of it, maybe the best approach would be to modify the `GetECInstanceDisplayLabel` custom function to return `NULL` if class or instance IDs are `NULL`?

Yato333: Made it so now `ContentQueryBuilder` wraps the column with `IIF` and not the `QueryBuilderHelpers::CreateDisplayLabelField`. As a result, hierarchy node hashes are not affected anymore.
Also extracted related instance display label field creation to `QueryBuilderHelpers`, to avoid repetition in `ContentQueryBuilder` tests.

grigasp (approved):

PR #814, opened 2024-07-18
Fix a bug that cause INTERRUPT when preparing query in concurrent query manager

khanaffan:
The issue involves:
1. A long-running query that is repeatedly canceled by the monitor.
2. The monitor calls interrupt on the SQLite connection running the query.
3. The monitor interrupts again just before the long-running query is about to exit, while another query is being prepared on the same connection.
4. This leads to the monitor interrupting the preparation of a new request, causing it to fail.

The solution is to interrupt the request only when it's running and to do so only once. SQLite ignores subsequent interrupt requests. The query is marked as interrupted, and the interruption is only successful if the query is assigned a connection. If not, it is not interrupted.

Additionally, queries on the primary connection are not interrupted.


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.h
```diff
@@ -208,27 +208,30 @@ struct RunnableRequestBase {
         std::atomic_bool m_cancelled;
         uint32_t m_executorId;
         uint32_t m_connId;
+        bool m_interrupted;
```
pmconne, line 211: Not sure about threading scenario here - does this need to be atomic?

khanaffan: The variable is only accessed by a single thread through a monitor call. Since no other thread accesses this value, declaring it as atomic may not be necessary. However, for the sake of code readability and to future-proof against potential access by other threads, I will declare it as atomic.

pmconne (approved):

PR #815, opened 2024-07-22
Fix issue with importing schemas when XML is ill-formed

akshay05n:


iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -905,7 +905,8 @@ DbResult JsInterop::ImportSchemas(DgnDbR dgndb, bvector<Utf8String> const& schem
         if (SchemaReadStatus::Success != schemaStatus)
             return BE_SQLITE_ERROR;
 
-        schemas.push_back(schema.get());
+        if (schema.IsValid())
```
rschili, line 908: Hmm I feel like we should not just quietly procceed here...
The lower layers write something to the log, but another log message here seems appropriate, and I'd even argue that returning "Success" from this method call is not good. We would be returning Success, if we got multiple input schemas and only one was malformed, no?

ColinKerr: Does this mean we are returning a SchemaReadStatus of success when the schema is not valid?

rschili: Oh, right, to get to this point the SchemaReadStatus had to be success, which seems wrong.... OR, looking at the code, if it goes through LocateSchema it seems the schemaStatus is not set at all, so it would be Success ?!

akshay05n: That's right, we're returning Success even if all of the schemas were malformed. We can update the earlier check to something like - `if (SchemaReadStatus::Success != schemaStatus || schema.IsNull())` to return Error even if one of the schemas is malformed.
Alternatively, we can return Error for NULL schema and log an error message like "Error occurred while reading schema from XML string"

rschili: Oh wait, but LocateSchema does have a reference to SchemaReadstatus (I missed the last parameter). So what I think is the correct way to handle this is make it so that status is correctly set to an error.

rschili (approved):

pmconne (approved):

PR #819, opened 2024-07-23
check if current user still holds the write lock when upload changes

jiaruiz717:
iTwin/itwinjs-backlog#1170
 
There is a possible scenario that two users try to upload local changes to the same cloud container:
 
1. User1 grabs the write lock and sit there for a very long time. User1 does not release the write lock even after it has expired
2. User2 grabs the write lock after User1's write lock expires, and upload something to the cloud container
3. User1 then tries to upload changes,  we should disable User1 from uploading his changes because the write lock has expired
 
However, if there are no other users grab the write lock after User1, we allow User1 to upload changes(and refresh the write lock) even if the write lock has expired. 

Tests are added in itwinjs-core
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7000


iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -256,7 +274,7 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
     }
 
     Napi::Value UploadChanges(NapiInfoCR info) {
-        RequireWriteLock();
```
nick4598, line 259: Maybe everywhere we call RequireWriteLock, we should be calling CheckWriteLockHeldByCurrentUser()? Or potentially replace requriewritelock with your implementation?

khanaffan:
If CloudSqlite throws an exception and the operation fails (unable to upload changes), it is likely that iModelDb will execute abandonChanges(). However, the CloudSqlite local cache may still contain changed blocks. 

If an exception is thrown and the application exits, then restarts with the same iModel, the local cache might still have dirty pages. 

To abandon changes in the local cache when abandoning changes on the iModel, you would need to ensure that the local cache is also cleared or synchronized to reflect the abandonment of changes. 

Is there a method on ClouldSqlite that i can call in SchemaSync to discard any changes after sync operation but before upload?

jiaruiz717:
> If CloudSqlite throws an exception and the operation fails (unable to upload changes), it is likely that iModelDb will execute abandonChanges(). However, the CloudSqlite local cache may still contain changed blocks.
> 
> If an exception is thrown and the application exits, then restarts with the same iModel, the local cache might still have dirty pages.
> 
> To abandon changes in the local cache when abandoning changes on the iModel, you would need to ensure that the local cache is also cleared or synchronized to reflect the abandonment of changes.
> 
> Is there a method on ClouldSqlite that i can call in SchemaSync to discard any changes after sync operation but before upload?

Not sure if I understand this correctly, but there is another abandonChanges() in CloudContainer(JsCloudSqlite.cpp/IModelJsNative/JsCloudContainer/AbandonChanges()) which is used to abandon/revert all changes in the local cache. In SchemaSync we have nativeDb object so we can all nativeDb.cloudContainer?.abandonChanges() to discard all changes in local cache.

iModelJsNodeAddon/JsCloudSqlite.cpp
jiaruiz717: Resolved in commit [commit 35f01fd017118730a226782e7d4cbebc0752b487](https://github.com/iTwin/imodel-native/pull/819/commits/35f01fd017118730a226782e7d4cbebc0752b487)

khanaffan:
Thanks for pointing me to that method. I will update schema import code to call it when schema import fails.

iModelJsNodeAddon/JsCloudSqlite.cpp
nick4598: Thanks, are there other places in the code where we potentially rely on m_writeLockHeld and perhaps we shouldn't be checking that?

jiaruiz717: There's only one more place that is using m_writeLockHeld, `void Connect(NapiInfoCR info)`. I think that one is fine according to the comments
image

nick4598: what if we connected our container, grabbed the write lock, disconnected our container, let user2 grab the write lock, then connected our container again?

I imagine we'd end up in the catch within ResumeWriteLock, but I do wonder if m_writeLockHeld could still be true in this case?

iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -191,8 +191,20 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
...
+        ReadWriteLock(lockedBy);
+        auto lockedByGuid = lockedBy[JSON_NAME(guid)].asString();
+        // check if it's the same guid  
+        if (lockedByGuid != m_cache->m_guid) {
+            // another user grabbed the write lock after the current user's write lock expiration time, disable current user from operating
+            BeNapi::ThrowJsException(Env(), Utf8PrintfString("Container [%s] is currently locked by another user.", m_containerId.c_str()).c_str());
+        } else {
+            // Refresh the write lock for the current user no matter if the write lock expires
+            // 1. If the write lock expires, refresh it for the current user to operate further actions
+            // 2. If not, also refresh it in case the current user's write lock is about to expire when they call operations that require write lock to be present, and somehow expires during the operations
```
nick4598, line 204: Could you also potentially write a test on itwinjs side that asserts the write lock gets refreshed for a user? 

You might have to read bcv_kv similar to how you modified it in azuriteTest in order to achieve this.

iModelJsNodeAddon/JsCloudSqlite.cpp
jiaruiz717: Since none of these actions call (AbandonChanges | ReleaseWriteLock | ClearWriteLock) so m_writeLockHeld will remain true in this case. I will write a test to validate this

iModelJsNodeAddon/JsCloudSqlite.cpp
jiaruiz717: added in [commit ee5297e0e535456bd7b00797255309cbbeca0734](https://github.com/iTwin/itwinjs-core/pull/7000/commits/ee5297e0e535456bd7b00797255309cbbeca0734)

iModelJsNodeAddon/JsCloudSqlite.cpp
jiaruiz717: User1 and User2 are in different processes so they don't share the same m_writeLockHeld. User2's m_writeLockHeld is false if user2 didn't call acquireWriteLock(). User1's m_writeLockHeld is still true in this case.

However, since User2 has grabbed the write lock, we want User1 to AbandonChanges() when User1 reconnect, right?

nick4598: Yes, correct

jiaruiz717: Resolved in [commit c32f13881e48297481c3903552c2ed6109a6a921](https://github.com/iTwin/imodel-native/pull/819/commits/c32f13881e48297481c3903552c2ed6109a6a921)

iModelJsNodeAddon/JsCloudSqlite.cpp
```diff
@@ -191,8 +191,20 @@ struct JsCloudContainer : CloudContainer, Napi::ObjectWrap<JsCloudContainer> {
 
     void RequireWriteLock() {
         RequireConnected();
-        if (!m_writeLockHeld)
-            BeNapi::ThrowJsException(Env(), Utf8PrintfString("container [%s] is not locked for write access", m_containerId.c_str()).c_str());
+        BeJsDocument lockedBy;
```
nick4598, line 194: Does the requireWriteLock function still throw an error 'the container is not locked for write access'? If the user never calls acquireWriteLock?

nick4598: To me, it seems like the answer is no. We should have a test on itwinjs side that asserts we get that error if we never held the write lock in the first place.

PR #820, opened 2024-07-25
Convert any numeric textureIds to a hexadecimal string while loading a rendermaterial element.

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/6933

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
@@ -264,6 +264,24 @@ DgnElementCPtr DgnElements::LoadElement(DgnElement::CreateParams const& params, 
     if (DgnDbStatus::Success != el->_LoadFromDb())
         return nullptr;
 
+    if (elHandler->GetClassName().Equals("RenderMaterial")) {
+        if (el->m_jsonProps.hasMember("materialAssets") && el->m_jsonProps["materialAssets"].hasMember("renderMaterial") && el->m_jsonProps["materialAssets"]["renderMaterial"].hasMember("Map")) {
+            BeJsValue map = el->m_jsonProps["materialAssets"]["renderMaterial"]["Map"];
+            BeJsDocument mapToModify;
+            mapToModify.toObject();
+            map.ForEachProperty([&](Utf8CP memberName, BeJsConst memberJson) {
+                if (memberJson.isNumericMember("TextureId")) {
+                    int textureId = memberJson["TextureId"].GetInt();
+                    BeInt64Id id(textureId);
+                    Utf8String hexStr = id.ToHexStr();
+                    mapToModify[memberName]["TextureId"] = hexStr.c_str();
+                }
+                return false;
+            });
+            map.From(mapToModify);
+        }
+    }
+
     el->_OnLoadedJsonProperties();
 
     if (makePersistent) {


nick4598:
I could move this logic over to a private function like CheckRenderMaterialTextureIds... but not sure if thats the correct thing to do?

One issue I see with this is that LoadElement is considered const which means I would also need to make CheckRenderMaterialTextureIds const, but it feels wrong considering I am explicitly modifying the jsonProps in this case. That also means that maybe LoadElement is the wrong place for me to be doing this.

Open to any suggestions.

pmconne:
Consider fixing it up in `_OnLoadedJsonProperties` instead. Apparently I made just the same sort of [fix](https://github.com/iTwin/imodel-native/blob/b652775fbe778f2e411fe36435cf8ed4f2de6f36/iModelCore/iModelPlatform/DgnCore/ViewDefinition.cpp#L1568) for display styles years ago.

nick4598:
Thanks for the suggestion, pushed the changes.

iModelCore/iModelPlatform/DgnCore/DgnMaterial.cpp
```diff
@@ -13,6 +13,33 @@ namespace dgn_ElementHandler
...
+            mapToModify.toObject();
+            map.ForEachProperty([&](Utf8CP memberName, BeJsConst memberJson)
+                {
+                if (memberJson.isNumericMember("TextureId")) 
+                    {
+                    // Fix IDs that were previously stored as 64-bit integers rather than as ID strings.
+                    int textureId = memberJson["TextureId"].GetInt();
+                    BeInt64Id id(textureId);
+                    Utf8String hexStr = id.ToHexStr();
+                    mapToModify[memberName]["TextureId"] = hexStr.c_str();
```
pmconne, line 35: Why are you extracting it as a 32-bit signed integer? The briefcase Id is in the upper 32 bits, so unless the briefcase Id is zero the Id you get back will be wrong.

You already know it's a number, so ask `GetId64` to convert that to an Id and then assign it back as a hex string.

```suggestion
                    auto textureId = memberJson["TextureId"].GetId64<DgnTextureId>;
                    mapToModify[memberName]["TextureId"] = textureId;

```

iModelCore/iModelPlatform/DgnCore/DgnMaterial.cpp
```diff
@@ -13,6 +13,33 @@ namespace dgn_ElementHandler
...
+                    }
+                return false;
+                });
+            if (mapToModify.size() > 0)
+                map.From(mapToModify);
```
pmconne, line 40: Doesn't this obliterate any other properties in `map`?

nick4598: Ah yeah.. youre totally right

nick4598: Thanks for catching that, I switched over to your approach of casting to (BeJsValue&) in order to be able to modify it in place.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -271,16 +273,9 @@ struct BeJsConst {
             Utf8String::Sscanf_safe(str, fmt, &val);
             return val;
         }
-        if (isNumeric()) {
-            // Casting a double to a uint produces undefined results while casting a double to int is well defined.
-            // Similarly, casting int to uint is well defined.
-            // On ARM casting a negative double to uint results in 0.
-            // On Intel, casting a negative double to uint produces the same result as below.
-            return uint64_t(int64_t(GetDouble((double)defVal)));
```
nick4598, line 279: Little unsure about getting rid of this entire snippet outright.. Should this potentially be part of the implementation in napi.h of Getuint64?

pmconne: Yes, that's the proposal: let the napi implementation deal with the JS precision constraints, don't impose them on the other implementations.

node-addon-api/Napi/napi.h
```diff
@@ -131,6 +131,27 @@ struct NapiValueRef : JsValueRef {
     virtual bool GetBoolean(bool defVal) const override { return isNull() ? defVal : m_napiVal.ToBoolean(); }
     virtual int32_t GetInt(int32_t defVal) const override { return isNumeric() ? m_napiVal.ToNumber() : defVal; }
     virtual uint32_t GetUInt(uint32_t defVal) const override { return isNumeric() ? m_napiVal.ToNumber() : defVal; }
+    virtual int64_t GetInt64(int64_t defVal) const override {
+        if (isNumeric()) {
+            double number = m_napiVal.ToNumber();
+            if (number >= std::numeric_limits<int64_t>::min() && number <= std::numeric_limits<int64_t>::max())
```
pmconne, line 137: Don't you want to test against [Number.MIN/MAX_SAFE_INTEGER](https://github.com/iTwin/imodel-native/blob/76e41a24831d794d0b382057909c210d5dfc3c94/iModelCore/libsrc/jsoncpp/include/json/value.h#L320)? Any value outside of that range cannot be trusted, because it was lossily converted to a double.

node-addon-api/Napi/napi.h
```diff
@@ -131,6 +131,27 @@ struct NapiValueRef : JsValueRef {
...
+            double number = m_napiVal.ToNumber();
+            if (number >= std::numeric_limits<int64_t>::min() && number <= std::numeric_limits<int64_t>::max()) 
+                return static_cast<int64_t>(number);
+        }
+        return defVal;
+    }
+    virtual uint64_t GetUInt64(uint64_t defVal) const override { 
+        if (isNumeric()) {
+            double number = m_napiVal.ToNumber();
+            if (number >= std::numeric_limits<uint64_t>::min() && number <= std::numeric_limits<uint64_t>::max()) {
```
pmconne, line 145: Reject if < 0 or >= Number.MAX_SAFE_INTEGER.

node-addon-api/Napi/napi.h
nick4598: Yep, made that change thanks.

node-addon-api/Napi/napi.h
nick4598: Made that change

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
nick4598: Moved the snippet over to GetUint64 in napi.h

iModelCore/ecobjects/test/NonPublished/ECJsonUtilitiesTests.cpp
```diff
@@ -42,17 +88,17 @@ TEST_F(ECJsonUtilitiesTestFixture, JsonToId)
...
+    EXPECT_EQ(BeInt64Id(0), ECJsonUtilities::JsonToId<BeInt64Id>(jsonDoc)) << jsonStr;
 
     ASSERT_EQ(SUCCESS, ParseJson(rapidJson, jsonStr));
-    EXPECT_EQ(BeInt64Id(uint64_t(-10)), ECJsonUtilities::JsonToId<BeInt64Id>(rapidJson)) << jsonStr;
+    EXPECT_EQ(BeInt64Id(0), ECJsonUtilities::JsonToId<BeInt64Id>(rapidJson)) << jsonStr;
 
     jsonStr = "3.14";  // Not a valid value for an Id, but failing the method is not worth the overhead. So the test documents the behavior of the API
     ASSERT_EQ(SUCCESS, ParseJson(jsonDoc, jsonStr));
-    EXPECT_EQ(BeInt64Id(3), ECJsonUtilities::JsonToId<BeInt64Id>(jsonDoc)) << jsonStr << " floating numbers are rounded";
+    EXPECT_EQ(BeInt64Id(0), ECJsonUtilities::JsonToId<BeInt64Id>(jsonDoc)) << jsonStr << " floating numbers are given defaultValue of 0.";
```
nick4598, line 98: I decided to make the JsonToId function more strict and return invalidIds for these types of numbers. It seemed more correct to me to turn a 3.14 into a 0 when trying to convert it to an ID, same with the string "3.14".

This all originated from me starting to hit RAPIDJSON_ASSERTS when calling m_val->GetUint64 on something like 3.14 when its a BeRapidJsonValue. This is because originally i was just checking that the value `isNumeric()` before calling `GetUint64()` on it. I temporarily switched over to `IsUint64` before calling `GetUint64` but I was worried that was too restrictive. which caused me to add some more test cases to this specific test where I casted to double,float,int64,int32. This also led to the addition of IsLosslessUint64 which takes double or float.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -834,6 +837,28 @@ struct BeRapidJsonValue : JsValueRef {
...
+        if (m_value->IsDouble() && IsLosslessUint64(m_value->GetDouble())) {
+            return static_cast<uint64_t>(m_value->GetDouble());
+        }
+        if (m_value->IsFloat() && IsLosslessUint64(m_value->GetFloat())) {
+            return static_cast<uint64_t>(m_value->GetFloat());
+        }
+        return m_value->IsUint64() ? m_value->GetUint64() : defVal; 
+    }
+    bool IsLosslessUint64(double d) const {
+        if (d < 0.0 || d > static_cast<double>(std::numeric_limits<uint64_t>::max())) {
```
pmconne, line 851: I don't understand what definition of "lossless" you are using here. A double cannot losslessly hold every uint64_t value. Don't you want to compare against JS_MAX_SAFE_INTEGER?

nick4598: If it is a double that fits into the range of uint64_t then it can be converted to a uint64 safely was the thought process I had. I added this because when I casted something like 2 to a double, it would not return an id of 2 which I think it should. (Note I don't know 2 being a double is ever realistic)

pmconne: The bug you are trying to accommodate is somebody did the equivalent of:
```
ElementId id;
Json::Value v; // or RapidJson value
v = Json::Value(id.GetValue()); // -> uint64_t
```

That is always stored internally as a uint64_t, losslessly, not as a double. So these `IsLossless` functions will never be involved in your new code path for doubles.

If for some reason somebody's trying to convert a value stored as a **double** into a uint64_t, then that conversion is only accurate if the double value is <= Number.MAX_SAFE_INTEGER, aka the maximal extent of the range of contiguous integer values exactly representable with a 53-bit significand. That isn't a JavaScript thing, it's a IEEE 754 thing.

Our API intentionally makes it difficult to make this mistake (e.g., by making `Json::Value(uint64_t)` constructor explicit). All we want to do in this PR is prevent data corruption due to people having made that mistake.

nick4598: I was assuming that a double of size 2^55 for example could be converted to a uint without any loss, but that is not true?

I think I understand it more the other way around where all uints in the range 2^55-2^56 could not be converted to a double without loss because a double could not represent every single possible number between the range of 2^55-2^56.

Its true in both directions? I assumed that all doubles in the range 2^54->2^63 was a subset of the full range of numbers in the range 2^54->2^63. I also assumed that uint64 can represent the full range of numbers in the range 2^54->2^63).

pmconne: Ok, I see where you are coming from now.

iModelCore/ecobjects/test/NonPublished/ECJsonUtilitiesTests.cpp
```diff
@@ -71,17 +117,17 @@ TEST_F(ECJsonUtilitiesTestFixture, JsonToId)
 
     jsonStr = R"json("-10")json";  // Not a valid value for an Id, but failing the method is not worth the overhead. So the test documents the behavior of the API
```
pmconne, line 118:
```suggestion
    jsonStr = R"json("-10")json";  // Negative numbers are invalid Ids.

```

iModelCore/ecobjects/PublicApi/ECObjects/ECJsonUtilities.h
```diff
@@ -359,18 +372,35 @@ struct ECJsonUtilities
...
+            } 
+        else if (json.IsDouble() && IsLosslessUint64(json.GetDouble())) 
+            {
+            return TBeInt64Id((uint64_t) json.GetDouble());
+            }
+
+        if (stringCheckFailed || json.IsFloat() || json.IsDouble() || SUCCESS != JsonToInt64(val, json) || val < 0)
             {
             TBeInt64Id invalidId;
             invalidId.Invalidate();
```
pmconne, line 401:
```suggestion

```

Nit and not your change, but default constructor produces an invalid Id already.

pmconne (approved):

pmconne:
@ColinKerr @rschili or @calebmshafer please review.

iModelCore/iModelPlatform/DgnCore/DgnMaterial.cpp
```diff
@@ -13,6 +13,28 @@ namespace dgn_ElementHandler
...
+        BeJsValue materialAssets = GetMaterialAssetsR();
+        if (materialAssets.hasMember("renderMaterial") && materialAssets["renderMaterial"].hasMember("Map"))
+            {
+            BeJsValue map = materialAssets["renderMaterial"]["Map"];
+            map.ForEachProperty([&](Utf8CP memberName, BeJsConst memberJson)
+                {
+                if (memberJson.isNumericMember("TextureId")) 
+                    {
+                    // Fix IDs that were previously stored as 64-bit integers rather than as ID strings.
+                    auto textureId = memberJson["TextureId"].GetId64<DgnTextureId>();
```
ColinKerr, line 30: If we get an invalid id (0) here I think we should log and maybe assert.  While this shouldn't happen with the current problem we should be defensive so we find other issues out before they hit us in a transformation workflow or something like that

nick4598: https://github.com/iTwin/imodel-native/pull/820/commits/d5f65fa7dd3c1aa3cd60a71fe942b1c3666f4d5e
How does this look?

pmconne: Use [PRIu64](https://stackoverflow.com/questions/16859500/what-is-priu64-in-c) to format unsigned 64-bit integers.

pmconne: Or, make life easy and use `GetElementId().ToHexStr()` and `%s`.

nick4598: switched to ToHexStr() and %s

pmconne: LGTM will let @ColinKerr resolve.

ColinKerr (approved):

PR #821, opened 2024-07-26
Support in Formatting API for Bearing, Azimuth and Ratio

rschili:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1197
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1196

Bearing, Azimuth direction support and ratio formatting support, both with parsers.
EC-Schema integration is still missing, this is only support on the quantity framework side.


mathieu-fournier:
@rschili @ColinKerr 
I think itwin-core quantity parser is not yet ready to parse these types. So it will likely result in an error when showing or editing them in the user interface. 
We should coordinate this with itwin-core Quantity team. Add these to [QuantityFormatter](https://www.itwinjs.org/reference/core-frontend/quantityformatting/).

rschili:
> @rschili @ColinKerr I think itwin-core quantity parser is not yet ready to parse these types. So it will likely result in an error when showing or editing them in the user interface. We should coordinate this with itwin-core Quantity team. Add these to [QuantityFormatter](https://www.itwinjs.org/reference/core-frontend/quantityformatting/).

We are aware. Afaik there is no dedicated itwin-core Quantity team, that task also falls to us. So once the native updates are complete and fully fleshed out, I will go and replicate the same thing in the itwin repo. We will roll out both at the same time.

iModelCore/Units/PublicAPI/Formatting/FormattingApi.h
```diff
@@ -370,6 +408,7 @@ friend struct Format;
 
         bool UpdateProblemCode(FormatProblemCode code) { return m_problem.UpdateProblemCode(code); }
```
rschili, line 409: This is a very strange method. The second parameter returns a pointer to the first parameter, and the method always returns an empty Quantity when we probably have a quantity object with a more precise error value available. I don't like this method, and I don't mind the bit of redundancy which it fixes. Let's just replace the calls, for example

return UpdateAndSetProblemCode(FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed, probCode);

By the full bloated version, like

probCode = FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed;
return converted; //quantity has error assigned

And then at the outmost caller (GetQuantity()) we can check if a problemCode has been set to the pointer and if so, call m_problem.UpdateProblemCode(code);

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
```diff
@@ -144,6 +144,11 @@ struct FormatParsingSet
     //! Returns error codes when FUS does not match the expression.
     //! The input expression signature code mus be provided by the caller
     BEU::Quantity ComposeColonizedQuantity(Formatting::FormatSpecialCodes cod, FormatCP fusP = nullptr);
+    BEU::Quantity ParseAndProcessTokens(Formatting::FormatSpecialCodes cod, FormatCP fusp, BEU::UnitCP inputUnit);
+    // parsing helper methods
+    BEU::Quantity ParseBearingFormat(FormatProblemCode* probCode, FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity ParseRatioFormat(FormatProblemCode* probCode,FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity UpdateAndSetProblemCode(FormatProblemCode code, FormatProblemCode* probCode);
```
rschili, line 151: This is a very strange method. The second parameter returns a pointer to the first parameter, and the method always returns an empty Quantity when we probably have a quantity object with a more precise error value available. I don't like this method, and I don't mind the bit of redundancy which it fixes. Let's just replace the calls, for example

return UpdateAndSetProblemCode(FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed, probCode);

By the full bloated version, like

probCode = FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed;
return converted; //quantity has error assigned

And then at the outmost caller (GetQuantity()) we can check if a problemCode has been set to the pointer and if so, call m_problem.UpdateProblemCode(code);

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
```diff
@@ -144,6 +144,11 @@ struct FormatParsingSet
     //! Returns error codes when FUS does not match the expression.
     //! The input expression signature code mus be provided by the caller
     BEU::Quantity ComposeColonizedQuantity(Formatting::FormatSpecialCodes cod, FormatCP fusP = nullptr);
+    BEU::Quantity ParseAndProcessTokens(Formatting::FormatSpecialCodes cod, FormatCP fusp, BEU::UnitCP inputUnit);
+    // parsing helper methods
+    BEU::Quantity ParseBearingFormat(FormatProblemCode* probCode, FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity ParseRatioFormat(FormatProblemCode* probCode,FormatCP fusP, BEU::UnitCP unit);
```
rschili, line 150: @ColinKerr  I would like to change this API, but what I have in mind is breaking. Check the exported method GetQuantity() on line 160 in this same file.
It's about the FormatProblemCode* parameter.
You can just call format->GetProblem() to obtain the exact same value.
What's even worse, you can also call FormatParsingSet->GetProblemCode() (this same class) to obtain the exact same value for the last call to ParseQuantity.
That's 3 mechanisms to obtain the same piece of information which is probably not used by any consumer other than our own tests.

Also, I recently added Quantity->GetProblemCode() which makes these mechanisms unnecessary. The returned Quantity can just indicate the error. Though, it uses a UnitsProblemCode other than a FormatProblemCode, so it's not the exact same thing. FormatProblemCode has been used as a dumping ground for all sorts of issues, including unit conversion problems, which are unrelated to format.

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -1541,4 +1543,443 @@ TEST_F(FormattingTestFixture, FormatsUsingDefaultNumericFormatSpecWhenItsNotSet)
...
+        {0.0,     "N00:00:00E", "N00°00'00\"E", "N00.000°E", "00:00:00", "00.000"},
+        {5.0,     "N05:00:00E", "N05°00'00\"E", "N05.000°E", "05:00:00", "05.000"},
+        {45.0,    "N45:00:00E", "N45°00'00\"E", "N45.000°E", "45:00:00", "45.000"},
+        {45.5028, "N45:30:10E", "N45°30'10\"E", "N45.503°E", "45:30:10", "45.503"},
+        {90.0,    "N90:00:00E", "N90°00'00\"E", "N90.000°E", "90:00:00", "90.000"},
+        {135.0,   "S45:00:00E", "S45°00'00\"E", "S45.000°E", "135:00:00", "135.000"},
+        {180.0,   "S00:00:00E", "S00°00'00\"E", "S00.000°E", "180:00:00", "180.000"},
+        {225.0,   "S45:00:00W", "S45°00'00\"W", "S45.000°W", "225:00:00", "225.000"},
+        {234.4972,"S54:29:50W", "S54°29'50\"W", "S54.497°W", "234:29:50", "234.497"},
+        {270.0,   "S90:00:00W", "S90°00'00\"W", "S90.000°W", "270:00:00", "270.000"},
```
rschili, line 1577: Expected result is N:90...

naronchen: just fixed

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -920,42 +930,54 @@ TEST_F(FormatParsingSetTest, CompositeFormats_IGNORED)
...
+    fps.GetQuantity(nullptr, &format);
     EXPECT_TRUE(fps.HasProblem()) << "Should be overflow because the number of digits in the floating part is more than 15";
     EXPECT_TRUE(fps.GetProblemCode() == FormatProblemCode::TooManyDigits);
 
     TestValidParseToQuantity("0.000000000000001 m", meter, 0.000000000000001);
-    fps = FormatParsingSet("0.0000000000000001 m", meter);
+    // fps = FormatParsingSet("0.000000000000001 m", meter);
+    // fps.GetQuantity(nullptr, &format);
+    FormatParsingSet fps2("0.000000000000001 m", meter);
+    fps2.GetQuantity(nullptr, &format);
```
ColinKerr, line 950: what's the deal here.  you create fps2 but don't validate it

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -1541,4 +1556,519 @@ TEST_F(FormattingTestFixture, FormatsUsingDefaultNumericFormatSpecWhenItsNotSet)
...
+        {580.0,   "S40:00:00W", "S40°00'00\"W", "S40.000°W", "220:00:00", "220.000"},
+        {640.0,   "N80:00:00W", "N80°00'00\"W", "N80.000°W", "280:00:00", "280.000"},
+    };
+
+
+    auto unitDegree = s_unitsContext->LookupUnit("ARC_DEG");
+    auto unitRadian = s_unitsContext->LookupUnit("RAD");
+    auto revolution = s_unitsContext->LookupUnit("REVOLUTION");
+    //auto unitMinute = s_unitsContext->LookupUnit("ARC_MINUTE");
+    //auto unitSecond = s_unitsContext->LookupUnit("ARC_SECOND");
```
ColinKerr, line 1600: clean up commented code.  either it should be there or it should be removed

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -1541,4 +1556,519 @@ TEST_F(FormattingTestFixture, FormatsUsingDefaultNumericFormatSpecWhenItsNotSet)
...
+    formatRatioTest("-0.5:1", -0.5, v_h, NtoOne, v_h);
+    formatRatioTest("1:-2", -0.5, v_h, valueBased, v_h);
+
+    formatRatioTest("-0.5:1", -2, v_h, NtoOne, h_v);
+    formatRatioTest("1:-2", -2, v_h, valueBased, h_v);
+    }
+
+    // really large/small numbers
+    {
+    formatRatioTest("0:1", 0.00000001, v_h, NtoOne, v_h);
```
ColinKerr, line 2027: I don't get this one

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
naronchen: just fixed it

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
naronchen: the format has precision:3
input 0.00000001 gets rounded to 0, 
so the output string is 0:1

ColinKerr (approved):

PR #831, opened 2024-08-13
Ratio formatting support

rschili:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1197


iModelCore/Units/PublicAPI/Formatting/FormattingApi.h
```diff
@@ -294,6 +294,10 @@ struct NumericFormatSpec
     void SetCounterClockwiseAngle(bool setTo) { SetTraitsBit(FormatTraits::CounterClockwiseAngle, setTo);}
     bool IsCounterClockwiseAngle() const {return GetTraitBit(FormatTraits::CounterClockwiseAngle);}
 
+    // TODO - Naron: this should be somewhere else?
```
rschili, line 297: It's an alright place, because the other numeric formatting code is in this class, too. I think it's fine. Another possible location for this would be in format directly (alongside formatQuantity)

iModelCore/Units/PublicAPI/Formatting/FormattingEnum.h
```diff
@@ -35,6 +35,8 @@ enum class PresentationType
     Station, // Civil Engineering Stationing (ie 1+00)
     Bearing, // Bearing angle e.g. N05:00:00E. Requires provided quantities to be of the angle phenomenon
     Azimuth, // Azimuth angle e.g. 45°30'00". Requires provided quantities to be of the angle phenomenon
+    FractionalRatio,
```
rschili, line 38: let's add some descriptions on how it works

iModelCore/Units/src/Formatting/Format.cpp
```diff
@@ -281,6 +281,23 @@ Utf8String Format::FormatQuantity(BEU::QuantityCR qty, BEU::UnitCP useUnit, Utf8
         additionalFormatting = true;
         }
 
+    if (fmtP->GetPresentationType() == PresentationType::FractionalRatio || fmtP->GetPresentationType() == PresentationType::IntegerRatio)
+        {
+        double dval = temp.GetMagnitude();
+        BEU::UnitCP unit = qty.GetUnit();
+
+        if (dval == 0)
+            if (unit->GetName() == "VERTICAL_PER_HORIZONTAL")
```
rschili, line 290: get rid of hard coded units, this should work in a generic way

rschili: this could be handled by the FormatToXXXRatio methods

iModelCore/Units/src/Formatting/NumericFormatSpec.cpp
```diff
@@ -959,4 +963,46 @@ double NumericFormatSpec::RoundDouble(double dval, double roundTo)
     return (dval < 0.0) ? -rnd : rnd;
     }
 
+Utf8String NumericFormatSpec::FormatToFractionalRatio(double dval, Utf8String unitName) const
+{
+    double reciprocal = 1.0 / dval;
+
+    Utf8String result;
+    if (unitName == "VERTICAL_PER_HORIZONTAL")
```
rschili, line 971: again, needs to work without hard coded units, the result should entirely depend on dval and nothing else. unitName should not be needed as a parameter

iModelCore/Units/src/Formatting/NumericFormatSpec.cpp
```diff
@@ -959,4 +963,46 @@ double NumericFormatSpec::RoundDouble(double dval, double roundTo)
...
+    while (b != 0) {
+        int temp = b;
+        b = a % b;
+        a = temp;
+    }
+    
+    numerator /= a;
+    denominator /= a;
+
+    if (unitName == "VERTICAL_PER_HORIZONTAL")
```
rschili, line 1000: same as before no hard coded units.
Maybe it helps to imagine it with different units than slope. You could have the input 25 for the "kilometers_per_hour" unit. I think the output should then be "25:1" and if you have an inverted unit for that, the magnitude would be converted into 0.04 (1 / 25), so the generated result should be "1:25"
I believe you can make this depend on whether the dval is bigger or smaller than 1.

rschili: But let's sync up with Simon about this. He can provide more and better examples. I'm not the expert for ratios...

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -1738,4 +1738,87 @@ TEST_F(FormattingTestFixture, AzimuthWithVariousBases) {
...
+
+        Units::Quantity quantity(value, *unit);
+        return fractionalRatioFormat.FormatQuantity(quantity);
+    };
+
+    // Vertical per Horizontal
+    {
+        auto verticalPerHorizontalUnit = s_unitsContext->LookupUnit("VERTICAL_PER_HORIZONTAL");
+
+        ASSERT_STREQ("0:1", formatFractionalRatio(0.0, verticalPerHorizontalUnit).c_str());
```
rschili, line 1758: The way Simon described it on the main story, @ColinKerr derived the following conclusion: https://github.com/iTwin/itwinjs-backlog/issues/1147#issuecomment-2217975255
So for integer ratio, you should not always have a "1" on one side of the ratio.
It's confusing, researching the web I find different definitions for integer/fractional ratio. Can you start a discussion and make sure we have some good test data to work with? We need to better understand the expected solution.

iModelCore/Units/src/Formatting/Format.cpp
```diff
@@ -281,6 +281,15 @@ Utf8String Format::FormatQuantity(BEU::QuantityCR qty, BEU::UnitCP useUnit, Utf8
...
 
+    if (fmtP->GetPresentationType() == PresentationType::FractionalRatio || fmtP->GetPresentationType() == PresentationType::IntegerRatio)
+        {
+        double dval = temp.GetMagnitude();
+        if (fmtP->GetPresentationType() == PresentationType::FractionalRatio)
+            return fmtP->FormatToFractionalRatio(dval);
+        else
+            return fmtP->FormatToIntegerRatio(dval);
+        }
+
```
rschili, line 292:
```suggestion
    if (fmtP->GetPresentationType() == PresentationType::FractionalRatio)
        return fmtP->FormatToFractionalRatio(temp.GetMagnitude());
    else if (fmtP->GetPresentationType() == PresentationType::IntegerRatio)
        return fmtP->FormatToIntegerRatio(temp.GetMagnitude());
```
This is shorter. Also I think giving the Format to that method may be a good idea because you may want to respect some of the format's properties (minWidth, ShowLabels and such)

iModelCore/Units/src/Formatting/NumericFormatSpec.cpp
```diff
@@ -959,4 +963,44 @@ double NumericFormatSpec::RoundDouble(double dval, double roundTo)
...
 
+Utf8String NumericFormatSpec::FormatToFractionalRatio(double dval) const
+{   
+    double reciprocal;
+    if (dval == 0.0)
+        reciprocal = 0.0;
+    else
+        reciprocal = 1.0 / dval;
+        
+    return "1:" + Format(reciprocal);
```
rschili, line 974: Should this be reversed if dval is greater than 1.0? The example you linked on the github issue says 0.2857 should turn into "1:3.5" but how about a value greater than 1? I believe in that case the second number should be 1 and the first should be variable.

iModelCore/Units/src/Formatting/NumericFormatSpec.cpp
```diff
@@ -959,4 +963,44 @@ double NumericFormatSpec::RoundDouble(double dval, double roundTo)
...
+}
+
+Utf8String NumericFormatSpec::FormatToIntegerRatio(double dval) const
```
rschili, line 977: Part of this file was migrated from old C-code, so it uses some unnecessary c practices. We should not do that for new code. dVal is such a thing. Name the variable by what it represents, not by what it is (meaning value instead of dval)

iModelCore/Units/src/Formatting/NumericFormatSpec.cpp
```diff
@@ -959,4 +963,44 @@ double NumericFormatSpec::RoundDouble(double dval, double roundTo)
...
+    else
+        reciprocal = 1.0 / dval;
+
+    double precisionFactor = GetDecimalPrecisionFactor();
+    reciprocal = RoundDouble(reciprocal, 1/precisionFactor);
+
+    int numerator = static_cast<int>(reciprocal * precisionFactor);
+    int denominator = static_cast<int>(precisionFactor);
+
+    // int gcd = std::gcd(numerator, denominator);
```
rschili, line 991: std::gcd seems like the best choice here. Is there a problem with it?

iModelCore/Units/tests/NonPublished/FormattingTests.cpp
```diff
@@ -1738,4 +1738,51 @@ TEST_F(FormattingTestFixture, AzimuthWithVariousBases) {
...
+            Format integerRatioFormat(integerRatioSpec);
+
+            Units::Quantity quantity(value, *unit);
+            return integerRatioFormat.FormatQuantity(quantity);
+        };
+
+    {
+        auto verticalPerHorizontalUnit = s_unitsContext->LookupUnit("VERTICAL_PER_HORIZONTAL");
+
+        ASSERT_STREQ("1:0", formatDoubleToRatio(0.0, verticalPerHorizontalUnit, PresentationType::FractionalRatio).c_str()); // special case
```
rschili, line 1757: these calls could be easier to read if you define a lambda that takes a string and double input, and then reuse it for all these examples

iModelCore/Units/PublicAPI/Formatting/FormattingApi.h
```diff
@@ -294,6 +299,10 @@ struct NumericFormatSpec
     void SetCounterClockwiseAngle(bool setTo) { SetTraitsBit(FormatTraits::CounterClockwiseAngle, setTo);}
     bool IsCounterClockwiseAngle() const {return GetTraitBit(FormatTraits::CounterClockwiseAngle);}
 
+    // Utf8String FormatToFractionalRatio(double dval) const;
+    // Utf8String FormatToIntegerRatio(double dval) const;
```
rschili, line 303: Remove unused code

.vscode/launch.json
```diff
@@ -110,7 +110,7 @@
             "program": "${env:OutRoot}Winx64/Product/Units-Gtest/UnitsTest.exe",
             "args": [
                 "--gtest_break_on_failure",
-                "--gtest_filter=CompositeValueSpecJsonTest.TestDefaultAndEmptySpacerRoundTrips"
+                "--gtest_filter=FormattingTestFixture.DoubleToRatio"
```
rschili, line 113:
```suggestion
                "--gtest_filter=CompositeValueSpecJsonTest.TestDefaultAndEmptySpacerRoundTrips"
```
remove change here, keep local

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
```diff
@@ -144,6 +144,11 @@ struct FormatParsingSet
     //! Returns error codes when FUS does not match the expression.
     //! The input expression signature code mus be provided by the caller
     BEU::Quantity ComposeColonizedQuantity(Formatting::FormatSpecialCodes cod, FormatCP fusP = nullptr);
+    BEU::Quantity ParseAndProcessTokens(Formatting::FormatSpecialCodes cod, FormatCP fusp, BEU::UnitCP inputUnit);
+    // parsing helper methods
+    BEU::Quantity ParseBearingFormat(FormatProblemCode* probCode, FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity ParseRatioFormat(FormatProblemCode* probCode,FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity UpdateAndSetProblemCode(FormatProblemCode code, FormatProblemCode* probCode);
```
rschili, line 151: This is a very strange method. The second parameter returns a pointer to the first parameter, and the method always returns an empty Quantity? What's the point of this?

iModelCore/Units/src/Formatting/FormatParsing.cpp
```diff
@@ -1220,4 +1149,249 @@ BEU::Quantity  FormatParsingSet::ComposeColonizedQuantity(Formatting::FormatSpec
...
+        // for input of "N:0" without reversed unit
+        return UpdateAndSetProblemCode(FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed, probCode);
+    }
+
+    BEU::Quantity qty = BEU::Quantity(numerator / denominator, *inputUnit);
+    converted = qty.ConvertTo(m_unit);
+
+    // for input of "0:N" with reversed unit
+    if (converted.GetProblemCode() == BEU::UnitsProblemCode::InvertingZero){
+        return UpdateAndSetProblemCode(FormatProblemCode::QT_MathematicOperationFoundButIsNotAllowed, probCode);
```
rschili, line 1379: Better return "converted" here, it has a problem code which is accurate.

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
```diff
@@ -144,6 +144,11 @@ struct FormatParsingSet
     //! Returns error codes when FUS does not match the expression.
     //! The input expression signature code mus be provided by the caller
     BEU::Quantity ComposeColonizedQuantity(Formatting::FormatSpecialCodes cod, FormatCP fusP = nullptr);
+    BEU::Quantity ParseAndProcessTokens(Formatting::FormatSpecialCodes cod, FormatCP fusp, BEU::UnitCP inputUnit);
+    // parsing helper methods
+    BEU::Quantity ParseBearingFormat(FormatProblemCode* probCode, FormatCP fusP, BEU::UnitCP unit);
+    BEU::Quantity ParseRatioFormat(FormatProblemCode* probCode,FormatCP fusP, BEU::UnitCP unit);
```
rschili, line 150: @ColinKerr I would like to change this API, but what I have in mind is breaking. Check the exported method GetQuantity() on line 160 in this same file.
It's about the FormatProblemCode* parameter.
You can just call format->GetProblem() to obtain the exact same value.
What's even worse, you can also call FormatParsingSet->GetProblemCode() (this same class) to obtain the exact same value for the last call to ParseQuantity.
That's 3 mechanisms to obtain the same piece of information which is probably not used by any consumer other than our own tests.

Also, I recently added Quantity->GetProblemCode() which makes these mechanisms unnecessary. The returned Quantity can just indicate the error. Though, it uses a UnitsProblemCode other than a FormatProblemCode, so it's not the exact same thing. FormatProblemCode has been used as a dumping ground for all sorts of issues, including unit conversion problems, which are unrelated to format.

PR #840, opened 2024-08-22
Do not search for the class from the pruned schema

akshay05n:


iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2687,14 +2687,22 @@ struct  InstanceXmlReader
                 return InstanceReadStatus::ECSchemaNotFound;
                 }
 
+            // do not search for the class if its schema is going to be pruned.
+            SchemaKey key;
+            ECObjectsStatus schemaNameParseStatus;
+            if (ECObjectsStatus::Success == (schemaNameParseStatus = SchemaKey::ParseSchemaFullName(key, m_fullSchemaName.c_str())) && m_context.IsSchemaToBePruned(key))
```
rschili, line 2693:
```suggestion
            ECObjectsStatus schemaNameParseStatus = SchemaKey::ParseSchemaFullName(key, m_fullSchemaName.c_str());
            if (ECObjectsStatus::Success == schemaNameParseStatus) && m_context.IsSchemaToBePruned(key))
```

Same code but I think it's easier to read.

iModelCore/ecobjects/src/ECcontext.cpp
```diff
@@ -357,6 +359,15 @@ struct ECSchemaReadContextBackedInstanceReadContext: public ECInstanceReadContex
...
         return schema.get();
         }
+
+    /*---------------------------------------------------------------------------------**//**
+    * @bsimethod
+    +---------------+---------------+---------------+---------------+---------------+------*/
+    virtual bool IsSchemaToBePruned(SchemaKeyCR key) const
+        {
+        const auto& schemasToPrune = m_schemaReadContext.GetSchemasToPrune();
+        return std::find(schemasToPrune.begin(), schemasToPrune.end(), key.GetName()) != schemasToPrune.end();
```
rschili, line 369: I think we will want to match by name only, and not by full schema key

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2687,14 +2687,22 @@ struct  InstanceXmlReader
                 return InstanceReadStatus::ECSchemaNotFound;
                 }
 
+            // do not search for the class if its schema is going to be pruned.
```
ColinKerr, line 2690: This doesn't make sense to me.  If the schema has been pruned it won't be found so how would it even get to this code?

I suggest you write a test case where some schemas are pruned so you understand the problem before you write a solution.

iModelCore/ecobjects/src/ECcontext.cpp
ColinKerr: This is matching by name only `key.GetName()`

ColinKerr:
Overall this change seems wrong to me.  CustomAttributeInstanceReadContext already checks to see if the schema is pruned but it doesn't communicate this information back out to the deserialization code so it doesn't know how to handle it.

I think you could find some way to wiggle some better status information out from the `_FindSchemaCP` method to the `InstanceXmlReader.GetInstance` method.  That way you would only have to parse the schema name and scan the pruned schema list once per CA instead of twice.  Haven't thought through the right way to do it but it seems possible with some refactoring.

iModelCore/ecobjects/src/ECInstance.cpp
akshay05n: We already have a test for this scenario: SchemaDeserializationTest.PruneCAFromPrunedEC32Schemas.
In this test, we are pruning the referenced schema. The problem occurs because when ECInstanceReadContext::FindSchemaCP() can’t locate the schema, it returns a fallback schema, which, in this situation, ends up being the actual schema that relies on the reference schema.

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2606,6 +2606,12 @@ struct  InstanceXmlReader
             if (ECObjectsStatus::Success != SchemaKey::ParseSchemaFullName(key, m_fullSchemaName.c_str()))
                 return NULL;
 
+            if (m_context.IsSchemaToBePruned(key))
```
ColinKerr, line 2609: I don't think you have changed the behavior at all.  This method will still return null and then GetInstance will log an error because GetSchema returned a nullptr.

iModelCore/ecobjects/src/ECcontext.cpp
```diff
@@ -385,39 +402,55 @@ struct CustomAttributeInstanceReadContext: public ECInstanceReadContext
...
-            return it->second.get();
+            {
+            schema = it->second.get();
+            return ECObjectsStatus::Success;
+            }
 
-        if (m_schemaContext.GetSchemasToPrune().end() != std::find(m_schemaContext.GetSchemasToPrune().begin(), m_schemaContext.GetSchemasToPrune().end(), key.GetName()))
+        if (IsSchemaToBePruned(key))
             {
             LOG.infov("Skipping loading of the custom attribute because its schema %s is being pruned.", key.GetFullSchemaName().c_str());
```
c-macdonald, line 430: Can this be made debug or trace instead?

iModelCore/ecobjects/PublicApi/ECObjects/ECObjects.h
```diff
@@ -191,6 +191,8 @@ enum class ECObjectsStatus
     InvalidConstantUnit,
     ClassTypeNotCorrect,
     InvalidUnit,
+    PruneSchema,
```
rschili, line 194: More consistent with the other items in this enum I'd call this "SchemaIsPruned".

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2687,6 +2698,12 @@ struct  InstanceXmlReader
                 return InstanceReadStatus::ECSchemaNotFound;
                 }
 
+            if (ECObjectsStatus::PruneSchema == schemaStatus)
+                {
+                LOG.infov("Skipping finding of the class '%s' because its schema '%s' is being pruned.", m_className.c_str(), m_fullSchemaName.c_str());
+                return InstanceReadStatus::ECSchemaNotFound;
```
c-macdonald, line 2704: Can this be made debug or trace instead?

rschili, line 2704: I think this can be improved. The calling code in `iModelCore\ecobjects\src\ECCustomAttribute.cpp` checks if that return value is not success, and then goes to check if the schema is being pruned.
We already know that it is being pruned, and we could return that information here, so the caller doesn't have to repeat that check. Just add a new value InstanceReadStatus::ECSchemaPruned

akshay05n: Changed logging level to debug

iModelCore/ecobjects/PublicApi/ECObjects/ECObjects.h
akshay05n: Updated the enum value to `SchemaIsPruned`

rschili (approved):

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2625,7 +2631,11 @@ struct  InstanceXmlReader
                 return nullptr;
                 }
 
-            ECSchemaCP constraintSchema = Utf8String::IsNullOrEmpty(constraintSchemaName.c_str()) ? defaultSchema : GetSchema(constraintSchemaName);
+            ECSchemaCP constraintSchema = Utf8String::IsNullOrEmpty(constraintSchemaName.c_str()) ? defaultSchema : nullptr;
```
ColinKerr, line 2634: While we're here looking at this what happens if a relationship constraint is in a pruned schema.  I don't think we will hit this case with the standard set of schemas on the prune list but it's worth checking if we've got a test for that

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -4444,13 +4462,13 @@ bool IECInstance::SaveOnlyLoadedPropertiesToXml() const
...
 +---------------+---------------+---------------+---------------+---------------+------*/
-ECSchemaCP ECInstanceReadContext::FindSchemaCP(SchemaKeyCR key, SchemaMatchType matchType) const
+ECObjectsStatus ECInstanceReadContext::FindSchemaCP(SchemaKeyCR key, SchemaMatchType matchType, ECSchemaCP& schema) const
     {
-    ECSchemaCP schema = _FindSchemaCP(key, matchType);
-    if (NULL != schema)
-        return schema;
+    ECObjectsStatus status = _FindSchemaCP(key, matchType, schema);
+    if (ECObjectsStatus::Success != status)
+        schema = &m_fallBackSchema;
```
ColinKerr, line 4469: It seems wrong to return the fallback schema when the schema we're looking for has been pruned

iModelCore/ecobjects/src/ECInstance.cpp
```diff
@@ -2687,6 +2698,12 @@ struct  InstanceXmlReader
                 return InstanceReadStatus::ECSchemaNotFound;
                 }
 
+            if (ECObjectsStatus::SchemaIsPruned == schemaStatus)
```
ColinKerr, line 2701: This check should be before the check for null schema ... I know the current behavior is for GetSchema to return the fallback schema but I think I'd even check for schema pruned and return directly when calling GetSchema.  

I assume the null check for the schema is old code because the current behavior makes that impossible.  If I were to clean things up more I would probably rewrite this code so that GetSchema does not return the fallback schema and we explicitly use it when necessary.  I don't think we should do that right now because we should just wrap this up ... but moving the pruned check to the call to GetSchema will make the code more resilient.

iModelCore/ecobjects/src/ECInstance.cpp
c-macdonald: That's what I pointed out in the very beginning...

iModelCore/ecobjects/src/ECInstance.cpp
akshay05n: It is improbable to hit this code if the schema is pruned because the code that finds the attributes in the relationship class is further down. We have a test called "SchemaDeserializationTest.PruneSchemas" that checks if the relationship works even when it uses constraints from pruned schema.

iModelCore/ecobjects/src/ECInstance.cpp
akshay05n: I moved this check to be before the null schema check.

iModelCore/ecobjects/src/ECInstance.cpp
akshay05n: The updated code does not return fallback schema in case the schema we're looking for is pruned

ColinKerr (approved):

PR #845, opened 2024-08-27
Make ECDb and ECObjects more resilient to handle schema evolutions

RohitPtnkr1996:
This PR contains only schema related changes to ECDb and ECObjects to make them more flexible when handling schemas that have unknowns possibly used with newer versions of ECDb.

A schema with a greater ECXml version than the one currently supported by ECDb can be deserialized, but it cannot be imported into the Db.

If a schema containing newer unknowns already exists in the Db opened by ECDb, the schema should be allowed to load.
However any classes or properties that contain unknowns will be skipped over and not be loaded.

This schema with unknowns cannot be then serialized.


iModelCore/ECDb/ECDb/SchemaReader.cpp
```diff
@@ -287,20 +297,30 @@ ECClassP SchemaReader::GetClass(Context& ctx, ECClassId ecClassId) const
...
+                if (schema.OriginalECXmlVersionGreaterThan(ECVersion::Latest))
+                    {
+                    CacheSchemaElementWithUnknowns(TABLE_Class, ecClassId); // Schema is of a newer version and the relationship class contains new unknown features. Mark this class to not be loaded.
+                    m_cache.SetClassEntryToNull(ecClassId);
+                    LOG.warningv("The ECRelationshipClass %s.%s has a StrengthType or ECRelatedInstanceDirection which is unsupported. The file might have been used with newer versions of the software. The class will not be loaded.",
                            schema.GetName().c_str(), className);
+                    }
+                else
+                    {
+                    LOG.errorv("Failed to load ECRelationshipClass %s.%s. Its StrengthType or ECRelatedInstanceDirection is unsupported. The file might have been used with newer versions of the software.",
```
rschili, line 315: I suggest changing the phrase "might have been used with newer versions", we actually know that it has been used with newer versions. I'd also avoid the term "file" since it could be read from memory. Something along the lines of "The schema uses a newer version of the standard than is known to the current software".

rschili (approved):

iModelCore/ECDb/ECDb/SchemaReader.cpp
RohitPtnkr1996: Done!

PR #849, opened 2024-08-29
Calculated property whose value is the related instance label

JonasDov:
Closes https://github.com/iTwin/itwinjs-core/issues/7050


iModelCore/ECPresentation/Source/Shared/ECExpressions/ECExpressionContextsProvider.cpp
```diff
@@ -263,14 +271,129 @@ struct ECInstanceMethodSymbolsProvider : IECSymbolProvider
...
+        return ExpressionStatus::Success;
+        }
+        
 protected:
     Utf8CP _GetName() const override {return "ECInstanceMethods";}
     void _PublishSymbols(SymbolExpressionContextR context, bvector<Utf8String> const& requestedSymbolSets) const override
         {
         context.AddSymbol(*MethodSymbol::Create("IsOfClass", NULL, &IsOfClass));
+        void* methodContext = const_cast<Context*>(&m_context);
+        context.AddSymbol(*MethodSymbol::Create("GetRelatedDisplayLabel", nullptr, &GetRelatedDisplayLabel, methodContext));
```
grigasp, line 393: Please add this function to [this table](https://github.com/iTwin/itwinjs-core/blob/efcafe30035c2469c3cace676e5ffcfd1e19ed5b/docs/presentation/advanced/ECExpressions.md?plain=1#L272) in associated `itwinjs-core` pr.

iModelCore/ECPresentation/Source/Shared/Queries/CustomFunctions.cpp
```diff
@@ -451,6 +451,13 @@ struct EvaluateECExpressionScalar : CachingScalarFunction<bmap<ECExpressionScala
             ECExpressionsCache noCache;
             if (ECExpressionsHelper(noCache).EvaluateECExpression(value, expression, *expressionContext) && value.IsPrimitive() && value.ConvertPrimitiveToString(expressionResult))
                 {
+                LabelDefinitionPtr labelDefinition = LabelDefinition::FromString(expressionResult.c_str());
+                if (labelDefinition->IsDefinitionValid())
+                    {
+                    Utf8StringCR labelRef = labelDefinition->GetDisplayValue();
+                    expressionResult = labelRef;
+                    }
```
grigasp, line 459: Why is this needed?

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/ContentIntegrationTests.cpp
```diff
@@ -6580,6 +6580,138 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, ContentRelatedInstancesSpec
...
+
+    // create the rule set
+    PresentationRuleSetPtr rules = PresentationRuleSet::CreateInstance(BeTest::GetNameOfCurrentTest());
+    m_locater->AddRuleSet(*rules);
+
+    ContentRuleP rule = new ContentRule("", 1, false);
+    rule->AddSpecification(*new ContentRelatedInstancesSpecification(1, 0, false, "", RequiredRelationDirection_Backward, relationshipAHasB->GetFullName(), classA->GetFullName()));
+    rules->AddPresentationRule(*rule);
+    rules->AddPresentationRule(*new LabelOverride(Utf8PrintfString("ThisNode.IsInstanceNode ANDALSO this.IsOfClass(\"%s\",\"%s\")", classB->GetName().c_str(), GetSchema()->GetName().c_str()),
+        1, "this.UserLabel", ""));
```
grigasp, line 6624: For simplicity, I think we should:
- use `ContentInstancesOfSpecificClassesSpecification` instead of `ContentRelatedInstancesSpecification`,
- use `InstanceLabelOverride` instead of `LabelOverride` (e.g. `rules->AddPresentationRule(*new InstanceLabelOverride(0, false, classB->GetFullName(), "UserLabel"))`).

iModelCore/ECPresentation/Source/Shared/ECExpressions/ECExpressionContextsProvider.cpp
JonasDov: https://github.com/iTwin/itwinjs-core/pull/7124

iModelCore/ECPresentation/Source/Shared/Queries/CustomFunctions.cpp
JonasDov: GetRelatedDisplayLabel returns stringified json object, that looks like this:
"{\"DisplayValue\":\"lc1\",\"TypeName\":\"string\",\"RawValue\":\"lc1\"}"

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/ContentIntegrationTests.cpp
JonasDov: Updated

iModelCore/ECPresentation/Source/Shared/ECExpressions/ECExpressionContextsProvider.cpp
grigasp: please make sure to complete the `itwinjs-core` PR after this one

iModelCore/ECPresentation/Source/Shared/Queries/CustomFunctions.cpp
grigasp: I think that's a problem - we may receive an ECExpression like `GetRelatedDisplayLabel("x", "y", "z") & "_ABC"` and it should concatenate label of `z` with string `_ABC`. With what we have now, `GetRelatedDisplayLabel` will return an stringified JSON object and add the `_ABC` suffix to it. This will result in an invalid JSON, so `LabelDefinition` won't parse it, so the end result will be something like `{ "DisplayValue": "", "RawValue": "", "TypeName": "string" }_ABC`.

IMO, `LabelDefinition` should be parsed in `ECInstanceMethodSymbolsProvider::GetRelatedDisplayLabel`. But first we need a test for this case.

JonasDov: Moved parsing to GetRelatedDisplayLabel and added `ContentInstancesOfSpecificClassesSpecification_GetsRelatedLabelWithAppendedValueThroughCalculatedProperty` test which checks the scenario you've mentioned.

iModelCore/ECPresentation/Source/Shared/ECExpressions/ECExpressionContextsProvider.cpp
```diff
@@ -373,7 +373,10 @@ struct ECInstanceMethodSymbolsProvider : IECSymbolProvider
 
             if (BE_SQLITE_ROW == QueryExecutorHelper::Step(stmt))
                 {
-                evalResult.InitECValue() = ECValue(stmt.GetValueText(0));
+                LabelDefinitionPtr labelDefinition = LabelDefinition::FromString(stmt.GetValueText(0));
+                Utf8String expressionResult = labelDefinition->GetDisplayValue();
```
grigasp, line 377: `GetDisplayValue()` returns a reference - we should avoid assigning it to `Utf8String`, since that makes an unnecessary copy. I don't think we need the `expressionResult` variable at all.

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/ContentIntegrationTests.cpp
```diff
@@ -6645,6 +6645,71 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, ContentInstancesOfSpecificC
...
+    ECClassCP classB = GetClass("B");
+    ECRelationshipClassCP relationshipAHasB = GetClass("A_Has_B")->GetRelationshipClassCP();
+
+    // insert some instances
+    IECInstancePtr instanceA = RulesEngineTestHelpers::InsertInstance(s_project->GetECDb(), *classA, [](IECInstanceR instance){instance.SetValue("UserLabel", ECValue("A_UserLabel")); });
+    IECInstancePtr instanceB = RulesEngineTestHelpers::InsertInstance(s_project->GetECDb(), *classB, [](IECInstanceR instance){instance.SetValue("UserLabel", ECValue("B_UserLabel")); });
+    RulesEngineTestHelpers::InsertRelationship(s_project->GetECDb(), *relationshipAHasB, *instanceA, *instanceB);
+
+    // set up input
+    KeySetPtr input = KeySet::Create(*instanceB);
```
grigasp, line 6679: Input is irrelevant when using `ContentInstancesOfSpecificClassesSpecification`. Just pass an empty `KeySet` when requesting content instead. Here and in the other two tests.

grigasp (approved):

PR #866, opened 2024-09-23
reread the extents from the database when getIModelProps is called after a pullChanges

nick4598:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7187

Fixes https://github.com/iTwin/itwinjs-backlog/issues/1240


iModelJsNodeAddon/JsInteropDgnDb.cpp
```diff
@@ -509,11 +509,20 @@ DgnDbStatus JsInterop::SimplifyElementGeometry(DgnDbR db, Napi::Object simplifyA
...
 +---------------+---------------+---------------+---------------+---------------+------*/
-void JsInterop::UpdateProjectExtents(DgnDbR dgndb, BeJsConst newExtents) {
+void JsInterop::UpdateProjectExtents(DgnDbR dgndb, BeJsConst newExtents, bool fromChangesetAppliedEvent) {
     auto& geolocation = dgndb.GeoLocation();
     AxisAlignedBox3d extents;
     extents.FromJson(newExtents);
-    geolocation.SetProjectExtents(extents);
+    if (fromChangesetAppliedEvent) {
+        // Check if our extents have actually changed. If not, return early.
+        if (geolocation.GetProjectExtents().IsEqual(extents))
```
bbastings, line 518: Couldn't we just always check if really changed and not bother with "fromChangesetAppliedEvent"?

iModelJsNodeAddon/JsInteropDgnDb.cpp
```diff
@@ -509,11 +509,20 @@ DgnDbStatus JsInterop::SimplifyElementGeometry(DgnDbR db, Napi::Object simplifyA
...
+            return;
+        // set isValid to false so that we recalculate the ecefLocation in SetProjectExtents with these new extents.
+        auto ecefLocation = geolocation.GetEcefLocation();
+        ecefLocation.m_isValid = false;
```
bbastings, line 522: This seems wrong unless there is a GCS?

iModelJsNodeAddon/JsInteropDgnDb.cpp
nick4598: Any opinions on what that tolerance should be? I pushed '.01' up just now.

Youre right, I probably dont need to check fromChangesetAppliedEvent before checking if itsEqual. I just figured this function has been working fine before checking for isEqual so might as well keep it to the fromChangesetAppliedEvent.

Changed it to check isEqual always

iModelJsNodeAddon/JsInteropDgnDb.cpp
nick4598: Does ecefLocation's `m_isValid` get marked as true anywhere except when the GCS is valid? Im guessing if there is no GCS, this code will just mark m_isValid as false when it likely already was false and have no effect?

Now based off of your other comment, I am also wondering if I should not be checking the fromChangesetAPpliedEvent in thsi function at all (i.e. also mark eceflocation as invalid when updateprojectExtents is called with any new extents).

bbastings: You can have a valid ecef location without a gcs. The reason you can mark the ecef location invalid when there is a gcs is because the ecef location can be computed from the gcs (it's effectively redundant). No gcs though and the ecef location is all you have to geo-locate the model.

iModelJsNodeAddon/JsInteropDgnDb.cpp
bbastings: I'd compare with something like DoubleOps::SmallMetricDistance(). 10 mm is significate and can't be attributed to fuzz...

iModelCore/iModelPlatform/DgnCore/DgnUnits.cpp
```diff
@@ -476,9 +476,12 @@ void DgnGeoLocation::SetProjectExtents(AxisAlignedBox3dCR newExtents)
     // JsonCpp and RapidJson differ slightly in precision of floating point numbers.
     // Tile content Ids include a hash of the project extents.
     // Differing precision => different content Ids => invalidate every cached tile in existence.
-    Json::Value jsonObj;
-    BeJsGeomUtils::DRange3dToJson(jsonObj, m_extent);
-    m_dgndb.SavePropertyString(DgnProjectProperty::Extents(), jsonObj.ToString());
+    if (!fromChangesetAppliedEvent)
```
bbastings, line 479: Don't see why this check is necessary either. You are expecting this to get called by JsInterop::UpdateProjectExtents only after it's verified that the value is really changed right? I don't like this "fromChangesetAppliedEvent" arg at all...I think it's hacky...

nick4598: The goal is to update the open iModel's props. If its from a changesetAppliedEvent that means the extents were already written to the be_prop table due to the changeset being applied so I figured I should skip that operation.

iModelJsNodeAddon/JsInteropDgnDb.cpp
nick4598: Okay so that means the ecefLocation can be independently and explicitly set? I switched it to check that the DgnGCS is not nullptr before invalidating the eceflocation

iModelCore/iModelPlatform/DgnCore/DgnUnits.cpp
bbastings: Doesn't make sense to me to call SetProjectExtents with an arg that says don't set the project extents. If you know the project extents changed and all it is that you want is to send out the onProjectExtentsChanged notification you should just do that and not be calling SetProjectExtents. I think Affan's comment on your itwinjs PR is the important one here...

iModelJsNodeAddon/JsInteropDgnDb.cpp
bbastings: Yes, it can be independently set and like I said is redundant when there is a GCS, Why do you think you need to invalidate it at all?

nick4598: I assumed that the ecefLocation was dependent on the project extents because it uses the center of the projectExtents

```        
        auto projectCenter = extents.GetCenter();
        if (SUCCESS == LatLongFromXyz(cartographicOrigin, projectCenter))
            {
            m_ecefLocation = EcefLocation(cartographicOrigin, projectCenter, dgnGCS);
            }
```

My test in core also asserts that the ecefLocation changed after pulling a changeset which passes. Of course that is after I  invalidated the ecefLocation but it did change. The iModel i chose "mirukuru.ibim" also appears to have a GCS as there is a test which asserts that `ecefLocation for iModels` in core.

bbastings: > I assumed that the ecefLocation was dependent on the project extents
It's not, no reason to invalidate it...

That code snippet is what sets up the ecef transform from the GCS when it's not defined yet. It just wants a "meaningful" world location point to use to establish the ecef origin. In the absence of any other information, the project center is a decent choice.

bbastings: > Am I understanding that code snippet correctly that it will set a new ECEFLocation ?

Yes, it derives the ecef transform from the GCS, since it's redundant when a GCS is present as I keep saying. The GCS is what you always want to use to convert between spatial and cartographic when available anyway. The ecef location becomes increasingly inaccurate as you get farther from the ecef origin due to a design model being drawn on a plane and the earth being curved.

nick4598: Is there any documentation you could point me to so that I can understand this better?

Is setting the ecef transform just a formality when there is a GCS then? Why do we do it? I suppose so that any code which converts between spatial and cartographic has the option of using the ECEF transform? even though its not recommended to do so?

 What code is converting between spatial and cartographic? I had assumed maybe the GetGeoCoordsFromIModelCoords and GetIModelCoordsFromGeoCoords function but Im not sure thats right.

bbastings: The measure tools use IModelConnection.spatialToCartographic...

iModelCore/iModelPlatform/DgnCore/DgnUnits.cpp
nick4598: switched to affan's suggestion

iModelJsNodeAddon/JsInteropDgnDb.cpp
nick4598: Thanks, also clarified some things on call. Much appreciated. Feel free to resolve if comfortable

aruniverse:
@khanaffan plz review

khanaffan (approved):

pmconne (approved):

aruniverse:

PR #872, opened 2024-09-27
Add db to OpenGroup to resolve schema changes

mattbjordan:
Opening changesets as a group requires the context of an iModel db to resolve any schema changes that may have happened in those changesets.

Fixes a bug where OpenGroup failed when the changesets included schema changes.

itwinjs-core:


pmconne (commented):
Looks reasonable to me, @khanaffan can you review?

PR #874, opened 2024-09-30
Added faster alternative to InVirtualSet()

soham-bentley:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7260
Fixes https://github.com/iTwin/itwinjs-backlog/issues/754


iModelCore/ECDb/ECDb/ECSql/ValueExp.cpp
```diff
@@ -817,47 +817,92 @@ Exp::FinalizeParseStatus MemberFunctionCallExp::_FinalizeParsing(ECSqlParseConte
     if (mode == Exp::FinalizeParseMode::AfterFinalizingChildren)
         {
         if (this->m_tableValuedFunc) {
```
rschili, line 819: You created a IsTableValuedFunc() method for this check. Either stick with it and use it everywhere or do this everywhere.

iModelCore/ECDb/ECDb/ECSql/ValueExp.cpp
```diff
@@ -915,6 +960,11 @@ Utf8String MemberFunctionCallExp::_ToString() const
 //+---------------+---------------+---------------+---------------+---------------+------
 bool MemberFunctionCallExp::_TryDetermineParameterExpType(ECSqlParseContext& ctx, ParameterExp& parameterExp) const
     {
+    if(this->m_tableValuedFunc && m_functionName.EqualsIAscii("IdSet"))
```
rschili, line 963: Same, use IsTableValuedFunc() here, also, we should put "IdSet" into a const/define somewhere instead of hard typing it in multiple spots

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -1899,8 +1899,35 @@ bool ECSqlParams::TryBindTo(ECSqlStatement& stmt, std::string& err) const {
...
+                    st = stmt.BindVirtualSet(index, idSet);
+                }
+                else if(binderInfo.GetBinderType() == BinderInfo::BinderType::ArrayECSqlBinderType && binderInfo.CheckIfBinderIsForIdSetVirtualTable())
+                {
+                    bool allElementsAdded = true;
+                    IECSqlBinder& binder = stmt.GetBinder(index);
+                    IdSet<BeInt64Id> set(param.GetValueIdSet());
+                    for(auto& ids: set)
+                    {
+                        st = ids.IsValid() ? binder.AddArrayElement().BindInt64((int64_t) ids.GetValue()) : binder.AddArrayElement().BindNull();
```
rschili, line 1915: Should we really Bind Null for invalid IDs? That seems dangerous to me. Maybe better skip those and log an error? Or error-out altogether? @khanaffan

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -1899,8 +1899,35 @@ bool ECSqlParams::TryBindTo(ECSqlStatement& stmt, std::string& err) const {
...
+                        }
+                    }
+                    if(allElementsAdded) // If even one array element has failed to be added we set the status for the entire operation as ECSqlStatus::Error although for the time being we don't do anything with status even if it fails
+                        st = ECSqlStatus::Success;
+                    else
+                        st = ECSqlStatus::Error;
+                }
+                else
+                {
+                    st = ECSqlStatus::Error;
```
rschili, line 1928: There should be a log message for this, otherwise troubleshooting will be very hard

iModelCore/ECDb/PublicAPI/ECDb/IECSqlBinder.h
```diff
@@ -10,6 +10,40 @@
 
 BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
 
+//=======================================================================================
+// @bsiclass
+//+===============+===============+===============+===============+===============+======
+struct BinderInfo final
```
rschili, line 16: I've thought about this one for a while. I understand you want to encapsulate the two pieces of information here.
But overall, it seems we'd be better off just putting this into ECSqlBinder directly, or is there some reason not to?
Requires less code and maintenance, and BinderInfo is not a very good term, it doesn't provide context on what info is encapsulated, while other info sits on the Binder directly.
I'd suggest to get rid of this struct.

iModelCore/ECDb/PublicAPI/ECDb/ECSqlStatement.h
```diff
@@ -249,6 +249,11 @@ struct EXPORT_VTABLE_ATTRIBUTE ECSqlStatement
         //! @return Parameter binder
         ECDB_EXPORT IECSqlBinder& GetBinder(int parameterIndex);
 
+        //! Gets BinderInfo for the binder associated with the parameter at the specified index.
+        //! @param[in] parameterIndex Parameter index.
+        //! @return Parameter BinderInfo. If there is no valid binder on that index the it will still return BinderInfo with Binder type NoopECSqlBinderType
+        ECDB_EXPORT BinderInfo const& GetBinderInfo(int parameterIndex) { return GetBinder(parameterIndex).GetBinderInfo(); }
```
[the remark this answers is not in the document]
soham-bentley: From the statement level we get a const reference so that we can just read values and modify values but from the binder level I have exposed normal reference so that we can also modify values if required and I have also done so in ECSqlBinder.cpp

iModelCore/ECDb/ECDb/ECSql/ValueExp.cpp
soham-bentley: done

iModelCore/ECDb/ECDb/ECSql/ValueExp.cpp
soham-bentley: Done

iModelCore/ECDb/ECDb/ECSql/ValueExp.cpp
soham-bentley: removed hard coded string as well

iModelCore/ECDb/PublicAPI/ECDb/ECSqlStatement.h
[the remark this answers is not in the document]
rschili: oh, I see. But that seems like it could be problematic, too. You could just make the fields on binder mutable, that BinderInfo object is bypassing the const mechanism anyways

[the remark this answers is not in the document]
soham-bentley: yes I have made it const everywhere

soham-bentley:
/ azp run imodel-native

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
```diff
@@ -434,6 +434,19 @@ enum class DbDeserializeOptions {
 };
 ENUM_IS_FLAGS(DbDeserializeOptions)
 
+//=======================================================================================
+// @bsiclass
+//=======================================================================================
+enum StatementState
+{
+    // The first four values are in accordance with sqlite3.c
```
soham-bentley, line 442: Is this enum structure and these associated comments okay?

iModelCore/ECDb/Tests/NonPublished/ECDbIdSetVirtualTableTests.cpp
```diff
@@ -0,0 +1,622 @@
...
+    }
+    {
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, "SELECT ECInstanceId FROM ECVLib.IdSet('[1,2,3,4,5]'), meta.ECClassDef where ECInstanceId = id group by ECInstanceId"));
+        ASSERT_EQ(true, m_ecdb.ExplainQuery(stmt.GetNativeSql(), true).Contains("SCAN IdSet VIRTUAL TABLE INDEX 1"));
+        ASSERT_EQ(true, m_ecdb.ExplainQuery(stmt.GetNativeSql(), true).Contains("SEARCH main.ec_Class USING INTEGER PRIMARY KEY (rowid=?)"));
+    }
+    {
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, "SELECT ECClassId FROM ECVLib.IdSet('[1,2,3,4,5]'), meta.ECClassDef where ECClassId = id group by ECClassId"));
```
soham-bentley, line 334: These are the query plan tests. How does these query plans look? @khanaffan

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
rschili: Phew, this is uncharted terrain for me, but I see the native values for VDBE*State are only the first four of these. NOT_PREPARED is not part of the values.
Given that they may add additional values, and the next one would be assigned = 4, this seems dangerous.
It seems much more robust to me to only expose the states which they use and always call "IsPrepared()" before that. That seems to be the intended design.

Your abstraction makes it more convenient to use, but in this case, modifying the API as little as possible seems better.

rschili (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
```diff
@@ -684,7 +684,7 @@ void ECSqlExpPreparer::RemovePropertyRefs(ECSqlPrepareContext& ctx, ClassRefExp
         if (propertyNameExp->IsPropertyRef())
             continue;
         if (propertyNameExp->IsVirtualProperty())
-            break;
```
soham-bentley, line 687: This "break" statement was stopping properties to be added and causing a malfunction in case of JOINS with virtual tables
Is there any specific reason why this was "break" and not "continue" ? @khanaffan  @ColinKerr  @rschili

rschili (approved):

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
ColinKerr: I agree with Rob

iModelCore/ECDb/ECDb/BuiltInVTabs.cpp
```diff
@@ -151,8 +151,192 @@ DbResult ClassPropsModule::Connect(DbVirtualTable*& out, Config& conf, int argc,
...
+        if(id == 0)
+            return BE_SQLITE_ERROR;
+        m_idSet.insert(id);
+    }
+    else
+        return BE_SQLITE_ERROR;
+    return BE_SQLITE_OK;
+}
+
+//---------------------------------------------------------------------------------------
```
ColinKerr, line 238: This PR has inconsistent bracket style.  Even within one method you are switching style.  Please use a consistent style, the style used by the rest of the C++ code is preferred

ColinKerr:
Have we gotten feedback on this new syntax?  If not I suggest that we put this behind the experimental flag so we can release and get feedback before we lock in on it

khanaffan (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparer.cpp
khanaffan: i do not think its a feature, its a bug :)

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
soham-bentley: What should be the behavior in case someone calls GetStatementState() before the statement is prepared?

iModelCore/ECDb/ECDb/BuiltInVTabs.cpp
soham-bentley: All bracket styles fixed

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
ColinKerr: The problem is that the underlying flag is a u8 so if Sqlite adds another state it will be 4.  Unless we notice this we will now mistake that state for unprepared.


You could use a larger number, something beyond 255 would ensure there would be no conflict but picking something still within the u8 range but higher like 100 should be just as safe in practice.

khanaffan:
> Have we gotten feedback on this new syntax? If not I suggest that we put this behind the experimental flag so we can release and get feedback before we lock in on it

Technically, there is no syntax change to ECSQL. We already expose virtual tables `json1.json_each(json)` and `json1.json_tree(json)`. This made it straightforward to implement `IdSet` since we already support the syntax for virtual tables. The only new aspect is the virtual table/class itself, which is not experimental as it not a work in progress.

In my opinion, an experimental tag should be applied to any new feature that introduces new syntax (specific to ECSQL and not standard SQL) or any functionality, such as Functions, VirtualTables, or Pragmas, that are still in development and may change over time until they become stable.

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
soham-bentley: I have implemented the TryGetStatementState approach where we pass the StatementState as a reference arg and modify it's value and returns true if everything is successful or returns false

ColinKerr:
> > Have we gotten feedback on this new syntax? If not I suggest that we put this behind the experimental flag so we can release and get feedback before we lock in on it
> 
> Technically, there is no syntax change to ECSQL. We already expose virtual tables `json1.json_each(json)` and `json1.json_tree(json)`. This made it straightforward to implement `IdSet` since we already support the syntax for virtual tables. The only new aspect is the virtual table/class itself, which is not experimental as it not a work in progress.
> 
> In my opinion, an experimental tag should be applied to any new feature that introduces new syntax (specific to ECSQL and not standard SQL) or any functionality, such as Functions, VirtualTables, or Pragmas, that are still in development and may change over time until they become stable.

Maybe syntax was the wrong word to sue here.  IdSet is new ... have we had anyone try it and give us feedback?

iModelCore/ECDb/ECDb/BuiltInVTabs.cpp
```diff
@@ -151,8 +151,186 @@ DbResult ClassPropsModule::Connect(DbVirtualTable*& out, Config& conf, int argc,
...
+    }
+    bool flag = doc.ForEachArrayMember([&](BeJsValue::ArrayIndex a, BeJsConst k1)
+                                            {
+                                            if(BE_SQLITE_OK != FilterJSONBasedOnType(k1))
+                                                {
+                                                GetTable().SetError(SqlPrintfString("IdSet vtab: The element with index %u is invalid", a));
+                                                return true;
+                                                }
+                                            return false; 
+                                            });
```
ColinKerr, line 200: inconsistent bracket formatting

soham-bentley: done

iModelCore/ECDb/ECDb/ECSql/PragmaECSqlPreparedStatement.cpp
```diff
@@ -634,10 +636,17 @@ DbResult PragmaECSqlPreparedStatement::DoStep() {
...
-    return m_resultSet->Step();
+    if(m_isFirstStep)
+        {
+        if (!m_parameterMap.OnBeforeFirstStep().IsSuccess())
+            return BE_SQLITE_ERROR;
+        }
+    
+    DbResult res = m_resultSet->Step();
+    // if step actually succeeded and returned BE_SQLITE_DONE or BE_SQLITE_ROW on the sqlite side then we set this flag to false if flag is true, if the returned value is something else like BE_SQLITE_SCHEMA or anything else we don't set the flag to false
+    if((res == BE_SQLITE_DONE || res == BE_SQLITE_ROW) && m_isFirstStep)
```
soham-bentley, line 647: Here, If the step returned anything other than BE_SQLITE_DONE or BE_SQLITE_ROW I am not setting the m_isFirstStep flag to false, because according to me that means the first step is not completed. If the returned value is BE_SQLITE_SCHEMA then also I am not setting the m_isFirstStep flag to false, Is that the correct assumption and behavior from my side?

soham-bentley: As per discussion with @khanaffan over a call BE_SQLITE_DONE and BE_SQLITE_ROW should be the only two values we should be concerned about and for these two values we should be setting the m_isFirstStep flag to false. For other values like BE_SQLITE_SCHEMA or any other value m_isFirstStep flag should not be set to false.

khanaffan (approved):

soham-bentley:

PR #883, opened 2024-10-14
Make classIdsToClassNames explicit parameter independent of UseJsPropertyName

khanaffan:
**Previous behavior**

`UseJsPropertyName` also implicitly mean `ClassIdToClassName` was `true`. But if explicitly specified `ClassIdToClassName: false` we did not return classId(s) instead it will still return class names.

**New behavior**

`UseJsPropertyName` still default to `ClassIdToClassName: true` unless `ClassIdToClassName: false` is specified. If false was specified, then classId(s) are returned.

Instance query when use with option `USE_JS_NAMES` continue to return `UseJsPropertyName: true, ClassIdToClassName: true`.

@soham-bentley added  new test for various permutations of `ECSqlRowFormat`, `classIdToClassNames` & `abbrivateBlob`.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7264


rschili (approved):

khanaffan:
closing is not required. As the mention property has been marked deprecated in anther PR

PR #878, opened 2024-10-04
Fix regression caused by ClassId conversion to class name vs hexid

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7238

PR #885, opened 2024-10-16
Fixed iModel Evolution Tests

naveedkhan8067:
In response to the following changeset: https://github.com/iTwin/imodel-native/pull/878, the behavior of tests is updated.

With the new behavior, if alias or not, the system property will be true.


khanaffan (approved):

nick4598 (approved):

nick4598:
Why didn't this test have to be changed in the previous PR?

naveedkhan8067:
> Why didn't this test have to be changed in the previous PR?

Ideally, it should be fixed with the other PR but these tests are not part of the automated validation build in `imodel-native`. We do execute them as part of our validation in bis-schemas and there we reproduced the issue.

PR #889, opened 2024-10-18
Presentation: Fix calculated property display values

JonasDov:
Closes https://github.com/iTwin/presentation/issues/732


saskliutas (changes requested):

iModelCore/ECPresentation/Source/Content/ContentItemBuilder.cpp
```diff
@@ -19,14 +19,18 @@ USING_NAMESPACE_BENTLEY_ECPRESENTATION
...
     NULL_FORMATTED_PRIMITIVE_VALUE_PRECONDITION(value, type);
     rapidjson::Document json(allocator);
     ECValue v = ValueHelpers::GetECValueFromSqlValue(type, extendedType, value);
     Utf8String stringValue;
     if (v.IsNull())
         json.SetString("");
+    else if (type == PRIMITIVETYPE_DateTime)
+        {
+        json.SetString(v.GetDateTime().ToString().c_str(), json.GetAllocator()); ;
+        }
```
grigasp, line 33: In our C++ code we're not adding curly braces for one-liners. Also removed the extra semi-colon.

```suggestion
    else if (type == PRIMITIVETYPE_DateTime)
        json.SetString(v.GetDateTime().ToString().c_str(), json.GetAllocator());
```

iModelCore/ECPresentation/Source/Content/ContentItemBuilder.cpp
```diff
@@ -149,13 +154,22 @@ rapidjson::Document ContentValuesFormatter::GetFormattedValue(ECPropertyCR prop,
...
     DIAGNOSTICS_HANDLE_FAILURE(DiagnosticsCategory::Content, "Unexpected ECProperty type");
     }
+
+/*---------------------------------------------------------------------------------**//**
+// @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+rapidjson::Document ContentValuesFormatter::GetFormattedValue(PrimitiveType type, Utf8StringCR extendedType, IECSqlValue const& value, rapidjson::MemoryPoolAllocator<>* allocator) const
+    {
+    return GetFallbackPrimitiveValue(type, extendedType, value, allocator);
+    }
```
grigasp, line 164: I don't understand this. It's an overload of the other `GetFormatterValue` function, so I'd expect the behavior to be similar except for different arguments, but it's not. To me it looks like this method is not needed at all, since we should call `GetFormattedPrimitiveValue` instead.

grigasp (changes requested):

iModelCore/ECPresentation/Tests/NonPublished/Integration/Content/ContentIntegrationTests.cpp
```diff
@@ -3270,20 +3270,45 @@ TEST_F(RulesDrivenECPresentationManagerContentTests, ContentInstancesOfSpecificC
...
     EXPECT_FALSE(jsonValues["CalculatedProperty_7"].GetBool());
+    EXPECT_STREQ("False", jsonDisplayValues["CalculatedProperty_7"].GetString());
+
     EXPECT_EQ(0.25, jsonValues["CalculatedProperty_8"].GetDouble());
+    EXPECT_STREQ("0.25", jsonDisplayValues["CalculatedProperty_8"].GetString());
+
     EXPECT_STREQ("False", jsonValues["CalculatedProperty_9"].GetString());
+    EXPECT_STREQ("False", jsonDisplayValues["CalculatedProperty_9"].GetString());
+
     EXPECT_STREQ("2017-05-30T00:00:00.000", jsonValues["CalculatedProperty_10"].GetString());
```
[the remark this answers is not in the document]
JonasDov: Changed, raw value now returns utc string

iModelCore/ECPresentation/Source/Content/ContentItemBuilder.cpp
JonasDov: Removed `GetFormattedValue`

grigasp: I see you changed the call to `GetFallbackPrimitiveValue`, but is that correct, though? I believe we should be calling `GetFormattedPrimitiveValue` instead.

JonasDov: `GetFormattedPrimitiveValue` could not be used, because it requires `ECPropertyCR` parameter. This AddValue method is only used for calculated properties, renamed it to: `AddCalculatedPropertyValue`.

grigasp (approved):

grigasp:
@mergifyio backport release/4.9.x

PR #891, opened 2024-10-18
Regression 4.9.2 where aliased ECClassId switch from returning classId to className

khanaffan:
1. CTE query returning classid switch from className to classId in `4.9.x`. This broke a user. 
2. [PR 7235](https://github.com/iTwin/itwinjs-core/pull/7238) Fix CTE, but this requires properly handling extendType information. This result in another break in behavior where aliased classid property was return as className. In 4.8.x/ 4.9.0 it used to return Id not class name. This break only happens for Concurrent Query for ECSqlStatement was fine as it set a flag called `DoNotConvertClassIdsToClassNamesWhenAliased` to reproduce previous behavior.
3. This PR also set `DoNotConvertClassIdsToClassNamesWhenAliased` for concurrent query to reproduce behavior of classId returning class name when aliased to be compilable with 4.8.x

First query that broke user is following 
```sql
SELECT t(aClassId) AS (SELECT ECClassId FROM Bis.Element) SELECT * FROM t
-- 4.8.x return className
-- 4.9.0 return classId
```
```sql
SELECT ECClassId AS aClassId FROM Bis.Element
--4.9.0 returned classId
--4.9.2 return className
```

After current fix

```sql
SELECT t(aClassId) AS (SELECT ECClassId FROM Bis.Element) SELECT * FROM t
-- return className as in 4.8.x
```
```sql
SELECT ECClassId AS aClassId FROM Bis.Element
-- returned classId  as in 4.8.x
```

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7279

### DIFF
--- a/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
+++ b/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
@@ -1006,6 +1006,7 @@ void QueryHelper::Execute(CachedQueryAdaptor& cachedAdaptor, RunnableRequestBase
     const auto abbreviateBlobs = request.GetAbbreviateBlobs();
     const auto includeMetaData= request.GetIncludeMetaData();
     const auto classIdToClassNames = request.GetConvertClassIdsToClassNames();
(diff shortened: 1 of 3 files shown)


pmconne (approved):

khanaffan:
@Mergifyio backport release/4.9.x

PR #899, opened 2024-10-23
Fix addon declarations

grigasp:
https://github.com/iTwin/itwinjs-core/pull/6828 removed internal `StatusCodeWithMessage` interface, which was being used in addon declarations in a number of places. For consumers this resulted in the following error:

```
../common/temp/node_modules/.pnpm/@bentley+imodeljs-native@4.9.31/node_modules/@bentley/imodeljs-native/NativeLibrary.d.ts(7,132): error TS2305: Module '"../../../../../../../../platform/node_modules/@itwin/core-bentley/lib/cjs/core-bentley"' has no exported member 'StatusCodeWithMessage'.
```

Here I just add the same type of interface to the addon itself.

This reveals that our native addon package setup is flawed - it has no real dependencies on `core-bentley`, `core-common`, etc., but is using their types, relying on `core-backend` delivering those dependencies. This makes it very easy to have dependencies' types removed or changed without noticing they're being used in the addon until after the release. I think these `@itwin` packages should be regular dependencies, but not making this change in this PR as that probably requires a wider discussion.

### DIFF
--- a/iModelJsNodeAddon/api_package/ts/package-lock.json
+++ b/iModelJsNodeAddon/api_package/ts/package-lock.json
@@ -8,10 +8,10 @@
       "name": "imodeljs-native-typescript-api",
       "version": "0.0.1",
       "devDependencies": {
-        "@itwin/build-tools": "^4.5.0-dev.29",
-        "@itwin/core-bentley": "^4.5.0-dev.29",
-        "@itwin/core-common": "^4.5.0-dev.29",
-        "@itwin/core-geometry": "^4.5.0-dev.29",
+        "@itwin/build-tools": "^4.9.5",
+        "@itwin/core-bentley": "^4.9.5",
+        "@itwin/core-common": "^4.9.5",
+        "@itwin/core-geometry": "^4.9.5",
         "@itwin/eslint-plugin": "^3.7.0-dev.8",
         "@types/chai": "^4.3.6",
         "@types/chai-as-promised": "^7.1.6",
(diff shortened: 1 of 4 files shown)


pmconne (approved):

aruniverse:
fyi @nick4598 @DanRod1999

aruniverse:
> has no real dependencies on core-bentley, core-common, etc., but is using their types, relying on core-backend delivering those dependencies

If its just using these pkgs for types then dev dependency is correct. I think the issue here is the pkg manager being too loose and allowing to use dependencies outside of direct deps. Wonder if we switch to using pnpm here

grigasp:
> > has no real dependencies on core-bentley, core-common, etc., but is using their types, relying on core-backend delivering those dependencies
> 
> If its just using these pkgs for types then dev dependency is correct. I think the issue here is the pkg manager being too loose and allowing to use dependencies outside of direct deps. Wonder if we switch to using pnpm here

If the types from dependencies are exposed through exported APIs (you see them in output declaration files), then these dependencies should be regular ones rather than "dev". Otherwise, we end up with declarations exposing types unknown to consumer code.

On the other hand, I think in this PR the problem wasn't related to regular vs dev dependencies. Instead, it was probably because the dependencies are specified with `^`, which says it's okay to use higher dependency versions. The way we have everything set up, that's not really the case - `@internal` APIs may be removed at any time, even in patch releases, so packages using `@internal` APIs should have dependencies specified at exact version, without `^` or even `~`. This is a similar problem to https://github.com/iTwin/itwinjs-core/issues/7373.

aruniverse:
> packages using @internal APIs should have dependencies specified at exact version, without ^ or even ~

if we're ok with using `~`, then i suggest maybe we parse the [major].[minor] the addon is for in core from `iModelJsNodeAddon/package_version.txt` and update the dependencies in package.json

this can be a step that runs on our version bump jobs after we release a new minor

PR #902, opened 2024-10-28
Use `USE_IPv6` instead of  `ENABLE_IPV6`

nick4598:
From curl patchnotes: 

- build: prefer `USE_IPV6` macro internally (was: `ENABLE_IPV6`)

  Before this patch, two macros were used to guard IPv6 features in curl
  sources: `ENABLE_IPV6` and `USE_IPV6`. This patch makes the source use
  the latter for consistency with other similar switches.

  `-DENABLE_IPV6` remains accepted for compatibility as a synonym for
  `-DUSE_IPV6`, when passed to the compiler.

  `ENABLE_IPV6` also remains the name of the CMake and `Makefile.vc`
  options to control this feature.

I believe this is the issue affecting some connectors ability to grab a token.

2024-10-17 12:12:10,312 ERROR [0x000070d0] Bentley.Http - * HTTP #4 Connection error: Could not resolve hostname 'Could not resolve host: ::1' http://[::1]:50641/


aruniverse (approved):

jffmarker:
@nick4598, I have confirmed this works in the Connector with iTwin Synchronizer.

chuckkir (approved):

PR #903, opened 2024-10-28
Fix issue with upgrading from ECDb profile version 4001 to later

khanaffan:
There were two issues found
1. **BE_PROP is treated as a schema table**. When applying the changeset, we mistakenly treated `be_Prop` as a data table, although it stores profile versions. After applying the schema changeset, we update the table and validate the mapping, requiring the latest profile version to avoid validation failures. In this instance, we failed to validate `ClassView` because we couldn't read `NamedEnumerators`, which is supported in version 4002. However, since the `be_prop` changes weren't applied, the schema reader used version 4001, which doesn't load `NamedEnumerator`, leading to the map validation failure.
2. **SchemaSync uses diff to patch the container or file for the underlying SQL schema**. This is necessary because the 4002 profile adds new tables. While SchemaSync can synchronize data tables using data in `ec_*` tables, it couldn't create meta tables if they were missing. To address this, I added the SchemaDiff tool based on sqldiff. It generates patch SQL to sync the underlying SQL database. Now, SchemaSync can push/pull changes to profile tables as well. If new `ec_*`, `dgn_*`, or `be_*` tables are added or existing ones are modified, it will be able to sync them automatically.

Test added to verify profile upgrade from 4001 to latest to verify above. 

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7314


pmconne:
@khanaffan any evidence that the integration tests added in the linked itwinjs-core PR pass with these native changes. PR validation apparently doesn't run integration tests. Don't want to merge native PR with unverified changes.

@DanRod1999 why doesn't PR validation validate *everything*?

DanRod1999:
Is this not "everything". I thought after adding the integration tests we were covering it all, but If there's more I will add them.
https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3343303&view=logs&s=f2fc6aea-a9fb-5cd6-442b-07cc52f13929&j=021c8522-efef-5a85-8902-57b1cdfbb0c2

pmconne:
> I thought after adding the integration tests we were covering it all

Totally overlooked that, sorry/thanks!

iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
```diff
@@ -1551,16 +1551,31 @@ BentleyStatus TxnManager::PatchSlowDdlChanges(Utf8StringR patchedDDL, Utf8String
...
+
+    DbResult result = BE_SQLITE_OK;
     if (!info.IsEmpty()) {
-        LOG.infov("Skipping DDL Changes as IModel has schema sync enabled. Sync-Id {%s}.", info.GetSyncId().c_str());
+        // In SchemaSync, we still need to apply schema changes to ec_*, dgn_*, and be_* tables.
+        // We cannot determine which DDL is for profile tables, so we try applying all of them.
+        // If it fails, it's fine because SchemaSync::pull() will patch/update all tables as necessary after applying the changeset.
+        // Not applying the DDL can cause the current changeset to fail if a column in the profile table is missing.
+        // This issue is detected when applying a changeset that upgrades the ECDb profile from version 4.0.0.1 to a newer version.
+        // Version 4.0.0.2 adds new tables and columns to ec_* tables. Since SchemaSync::pull() is called,
```
pmconne, line 1565: > Since SchemaSync::pull() is called,

Did you have more to say here?

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -823,6 +823,505 @@ void     SnappyFromBlob::Finish() {m_blobIO.Close();}
...
+    DbResult rc;
+    auto lhsTables = MetaData::QueryTableNames(lhsDb, "main", rc);
+    if (rc != BE_SQLITE_OK)
+        return rc;
+
+    auto rhsTables = MetaData::QueryTableNames(rhsDb, "main", rc);
+    if (rc != BE_SQLITE_OK)
+        return rc;
+
+    // if (excludeFilter == nullptr) {
```
pmconne, line 1066: Did you comment this out intentionally? I ask because later, you use `excludeFilter` without checking if it's empty. If it is empty, it will throw `bad_function_call`.

iModelCore/ECDb/ECDb/SchemaSync.cpp
```diff
@@ -91,19 +76,31 @@ ProfileVersion SchemaSyncHelper::QueryProfileVersion(DbR db, ProfileKind kind) {
...
+    if (conn.IsReadonly()) {
+        LOG.error("Failed to save profile version. Database is readonly.");
+        return BE_SQLITE_ERROR;
+    }
     Utf8String profileVersionStr = ver.ToJson();
-    rc = conn.SavePropertyString(profileSpec, profileVersionStr);
-    if (rc != BE_SQLITE_OK)
+    auto rc = conn.SavePropertyString(profileSpec, profileVersionStr);
+    if (rc != BE_SQLITE_OK) {
+        LOG.error("Failed to save profile version.");
```
pmconne, line 101: Don't want to log `rc`?

pmconne (approved):
see comments.

khanaffan:
> @khanaffan any evidence that the integration tests added in the linked itwinjs-core PR pass with these native changes. PR validation apparently doesn't run integration tests. Don't want to merge native PR with unverified changes.
> 
> @DanRod1999 why doesn't PR validation validate _everything_?

This is not even a integration test as such. It uses azurite and local hub. We might need to look into backend integration tests and may be move them out as standalone backend test. Azurite is setup with integration test so that is why i had put the test in there with other cloudsqlite test.

khanaffan:
@DanRod1999 can we separate true integration test that actually connect to deployed services from one that use azurite and never access any deployed service? e.g. we can then call it `test:integration:with_mocked_serv` and  `test:integration:with_deployed_serv`. The first one will run like any normal standalone test.

iModelCore/ECDb/ECDb/SchemaSync.cpp
khanaffan: I added log error messages where i was just return rc.

iModelCore/BeSQLite/BeSQLite.cpp
khanaffan: fixed. It should not call function if it was not provided.

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -823,6 +823,514 @@ void     SnappyFromBlob::Finish() {m_blobIO.Close();}
...
+    }
+
+    auto filterOutUnsupportedTables = [&](std::vector<MetaData::TableInfo>& list) {
+        list.erase(std::remove_if(list.begin(), list.end(), [&](auto const& table) {
+            return excludeFilter != nullptr ? excludeFilter(table) : false;
+        }), list.end());
+    };
+
+    filterOutUnsupportedTables(lhsTables);
+    filterOutUnsupportedTables(rhsTables);
```
pmconne, line 1093:
```suggestion
    if (nullptr != excludeFilter) {
        auto filterOutUnsupportedTables = [&](std::vector<MetaData::TableInfo>& list) {
            list.erase(std::remove_if(list.begin(), list.end(), [&](auto const& table) {
                return excludeFilter(table);
            }), list.end());
        };

        filterOutUnsupportedTables(lhsTables);
        filterOutUnsupportedTables(rhsTables);
    }
```

PR #904, opened 2024-10-30
Support for running against libnode.dll

swbsi:
The convention of GetModuleHandle(NULL) in addon delay load handlers (https://github.com/nodejs/node-gyp/blob/e6f4ede10cca28e9edeaa85d7830914c5d1499c7/src/win_delay_load_hook.cc#L31) does not work when node is built as a shared library on windows in an embedding scenario, since the napi_* APIs are then defined in a dll, not the host exe.


jffmarker:
@swbsi do we still need this chnage?

PR #905, opened 2024-10-30
Improve Error Message for throwSqlError()

DanRod1999:
Resolves [Issue](https://github.com/iTwin/itwinjs-core/issues/7299)
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7308

Currently the throwSqlError() function only gives "SQL Error", and only after enabling sql logs can we see a more useful error message. See issue for more details.

instead of using threSqlError() I used throwSqlResult() which when given a DBResult we can gather the error message from SQL. Also expand throwSqlResult() to provide the full sql log message.


iModelJsNodeAddon/JsInteropDgnDb.cpp
```diff
@@ -871,7 +871,7 @@ Napi::String JsInterop::InsertLinkTableRelationship(DgnDbR dgndb, Napi::Object o
     BeSQLite::EC::ECInstanceKey relKey;
     auto rc = dgndb.InsertLinkTableRelationship(relKey, *relClass, sourceId, targetId, props.get()); // nullptr is okay if there are no props
     if (BE_SQLITE_OK != rc)
-        throwSqlError();
+        throwSqlError(rc);
```
khanaffan, line 874: Could be something like following, may be with source/target id.

```cpp
BeNapi::ThrowJsException(
    obj.Env(), 
    SqlPrintfString("Failed to insert relationship (%s)", relClass->GetFullName()).GetUtf8CP(), 
    (int) rc
);
```

DanRod1999: Proposing to instead use:
```
       JsInterop::throwSqlResult(
          "Failed to insert relationship",
          dgndb.GetDbFileName(),
          rc
        );
```
I also modified throwSqlResult to return the entire log error string, so it would look like this:
Error: Failed to insert relationship [D:\itwinjs-core\core\backend\lib\cjs\test\output\IModel\insertDuplicateInstance.bim]: rc=2067, constraint failed (BE_SQLITE_CONSTRAINT_UNIQUE)

khanaffan (approved):

DanRod1999:

PR #916, opened 2024-11-08
SQLite warning log spam when creating new Standalone iModels

naronchen:
Fixes #6066

changed impl of tableExists/ColumnExists/GetColumns in BeSQLite.cpp to lower number of warnings by querying against sqlite_master or pragma_table_info

added tests for warning with custom logger in imodelPlatform


naronchen:
Now there are 20+ warnings still
image
I wonder if they are avoidable. 

those warnings somehow only appear after tests complete, so I cant find where they are actually generated

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -3167,23 +3188,22 @@ void Db::FlushPageCache()
...
+
+    GetColumns(columns, tableName);
+    return columns.end() != std::find_if(columns.begin(), columns.end(), [columnName](Utf8StringCR col) { return col.EqualsIAscii(columnName); });
     }
 
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
 bool Db::GetColumns(bvector<Utf8String>& columns, Utf8CP tableName) const
     {
```
khanaffan, line 3201: You might also want to parse tableName and look for dot, as table name could be qualified by tablespace.

In this case, `PRAGMA table_info(?)` if table space as not given. Otherwise, use  `PRAGMA %s.table_info(?)`

and `columns.push_back(statement->GetValueText(0));` need to change to `columns.push_back(statement->GetValueText(`));`

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -3148,10 +3148,31 @@ DbResult Db::TruncateTable(Utf8CP tableName) const
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-bool Db::TableExists(Utf8CP tableName) const
+bool Db::TableExists(Utf8CP tableName, Utf8CP tableSpace) const
```
khanaffan, line 3151: Since you just parse tablespace from tableName and the function previously expected tablename to be optionally qualified by tablespace. So we can avoid adding this argument and just depend on caller to give us e.g. `TableExists('TEMP.table1')` instead of `TableExists('table1', 'TEMP')`

iModelCore/BeSQLite/BeSQLite.cpp
naronchen: addressed

khanaffan (approved):

pmconne (approved):

PR #919, opened 2024-11-14
Reenable disabled font test

tcobbs-bentley:
This will almost certainly cause the Mac pipelines to fail until we can figure out what is causing the failure, but these tests do need to run on all platforms.

### DIFF
--- a/iModelJsNodeAddon/api_package/ts/src/test/EmbedFont.test.ts
+++ b/iModelJsNodeAddon/api_package/ts/src/test/EmbedFont.test.ts
@@ -37,10 +37,7 @@ describe("embed fonts", () => {
       fs.removeSync(tempSQLiteDbName);
     tempSQLiteDb.createDb(tempSQLiteDbName);
 
-    if (process.platform !== "darwin") {
-      // This is failing on CI Macs, so disabling it for now.
-      tempSQLiteDb.embedFont(dummyRsc);
-    }
+    tempSQLiteDb.embedFont(dummyRsc);
     tempSQLiteDb.embedFont({ fileName: shxFileName });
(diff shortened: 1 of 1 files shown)


iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -294,7 +294,8 @@ template<typename T_Db> struct SQLiteOps {
                 BeNapi::ThrowJsException(info.Env(), "invalid face");
             faces.emplace_back(face);
 
-            auto napiData = argJson[JsInterop::json_data()].AsNapiValueRef();
+            auto jsonData = argJson[JsInterop::json_data()];
+            auto napiData = jsonData.AsNapiValueRef();
```
tcobbs-bentley, line 298: @pmconne The fact that this change fixes the problem strongly implies something else is wrong (like uninitialize variables). Can you think of any legitimate reason why this change would fix the problem? If not, do you know who can investigate the underlying code to try to determine if there is a bug there? Note that the failure only happened on the Mac CI build machines, and could not be reproduced anywhere else.

pmconne: Yes, `operator[]` produces a temporary object whose lifetime ends at the end of that statement, leaving the return value of `AsNapiValueRef` pointing to freed memory. Your change extends the lifetime. It's a good fix; the BeJsConst API is easy to trip over.

pmconne (approved):

iModelJsNodeAddon/IModelJsNative.cpp
tcobbs-bentley: OK, thanks. I added a comment to indicate that. Maybe that's why it took C++ so long to create a working shared_ptr.

PR #929, opened 2024-11-21
Remove special case for Mac ARM to be cross-compiled on x64 macs. If …

chuckkir:
…anything we will push the other direction and compile x64 on ARM.


khanaffan (approved):

aruniverse:
@chuckkir do we still need this?

PR #934, opened 2024-12-09
Chuck/merge libxml2

chuckkir:
Updates to libxml2-2.13.5. There is a corresponding change in BeXML. Libxml now enforces that the buffer length does not include the null terminating character by checking that all characters are used. The BeXML code passes in the length including the null terminator.  Perhaps that was needed in an earlier version, perhaps it was always wrong. I verified that running without the null character in the length works in the previous version.


iModelCore/Bentley/BeXml/BeXml.cpp
```diff
@@ -420,15 +420,16 @@ BeXmlDomPtr BeXmlDom::CreateAndReadFromString (BeXmlStatus& xmlStatus, Utf16CP s
...
-    // It's ok if characterCount includes the zero terminator.
+    // It's ok if characterCount includes the zero terminator although the call to xmlReadMemory must not.
 #if defined (_WIN32)
     // This uses wcslen, which is invalid on UTF-16 strings on non-Win32.
     BeAssert (charCountCloseEnough (characterCount, (WCharCP) source));
 #endif
 
-    size_t numCharsToCopy = ((0 == characterCount) || (source[characterCount - 1])) ? characterCount + 1 : characterCount; // TFS 15586
+    // Subtract if characterCount includes the null terminator.
+    size_t numCharsToCopy = (0 == characterCount || source[characterCount] != '\0') ? characterCount : characterCount - 1; // TFS 15586
```
chuckkir, line 430: @ColinKerr please review this file for the XML changes caused by the new libxml requiring that the character count not include the nul terminator.

pmconne (approved):

PR #936, opened 2024-12-11
imodeljs-native ts maintenance

aruniverse:
update deps for native pkg
- drop node18, 
- add support for node22, 
- update eslint & fix lint errors
- point to 5.0 pre-releases to catch any breaking changes we make in core


iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -1363,7 +1363,7 @@ export declare namespace IModelJsNative {
 
   class ECSchemaXmlContext {
     constructor();
-    public addSchemaPath(path: string): void;
+    public addSchemaPath(schemaPath: string): void;
```
calebmshafer, line 1366: why change this?

aruniverse: i did, just to be more explicit, its just a var rename in the types so im pretty sure it wouldnt break anything, but i can revert

calebmshafer: should be fine. I'll approve

calebmshafer (approved):

PR #939, opened 2024-12-16
Crash while stepping through results of ExtractPropertyExp for ArrayProperties

naronchen:
issue#1325 on itwinjs-backlog
https://github.com/iTwin/itwinjs-backlog/issues/1325


iModelCore/ECDb/ECDb/InstanceReaderImpl.cpp
```diff
@@ -407,7 +407,10 @@ bool Class::Seek(ECInstanceId rowId, ECN::ECClassId& rowClassId) const {
 // @bsimethod
 //+---------------+---------------+---------------+---------------+---------------+------
 bool Property::Seek(ECInstanceId rowId, ECN::ECClassId& rowClassId) const {
-    return m_table->Seek(rowId, &rowClassId);
+    OnAfterReset();
+    bool result = m_table->Seek(rowId, &rowClassId);
```
rschili, line 411: TableView::Seek is implemented below in this file and public. Callers can just call ->GetTable()->Seek() instead, and we would be back to the crash, no?
@khanaffan please have a look at this, too. Thanks

khanaffan: What you referring to is private impl of `InstanceReader` (pimpl). All it set behind `InstanceReader` api. None of the nested classes is accessible to caller. Only thing they can access is `InstanceReader`. which only expose following and what you referring to set behind `Impl* m_pImpl` which is forward declaration. We do same with ECSqlStatement the impl has many public method with no intention to be called directly. This is to avoid `friend` semantic and including privately held data structures from being part of public header file.  pImpl allow you to hide all your implementation from external user so you can break and change it without breaking external interface. 

```cpp
struct InstanceReader final {
 ...
 struct Impl;
    private:
        Impl* m_pImpl;
    public:
        InstanceReader(InstanceReader const&) = delete;
        InstanceReader& operator = (InstanceReader&) = delete;
        ECDB_EXPORT explicit InstanceReader(ECDbCR);
        ECDB_EXPORT ~InstanceReader();
        ECDB_EXPORT bool Seek(Position const&, RowCallback, Options const& = Options()) const;
        ECDB_EXPORT void Reset();
};
```

rschili (approved):

PR #946, opened 2024-12-30
Attempt to load/create WW15MGH._96 even if writable directory cannot be resolved

EimantasMorkunas:
Should fix https://github.com/iTwin/itwinjs-backlog/issues/1118.

The binary shadow file should exist in cloud cache workspace (`bTime = CS_fileModTime (__This->binaryPath)` below the changed code should return non-zero value) so it should be irrelevant if writable directory is available. The `CS_rwDictDir()` function copies source path to the result so the binary path should be valid even in case of an error.

If cloud cache is not used or binary file does not exist and a writable directory is not available, the code should still error out when attempting to create the binary file.


iModelCore/libsrc/csmap/Source/CS_egm96.c
```diff
@@ -529,12 +529,14 @@ int CSmkBinaryEgm96 (struct cs_Egm96_ *__This)
 	double deltaLng = cs_Zero;
 
 	st = CS_rwDictDir (__This->binaryPath,sizeof (__This->binaryPath),__This->filePath);
+#if !defined(GEOCOORD_ENHANCEMENT)	/* Even if writable dir cannot be resolved, the binary file might already exist. */
```
tcobbs-bentley, line 532: Given the extremely weird circumstances, I don't feel like disabling the error checking code on all platforms is appropriate. I think it would be better to have something like:
```c++
#if defined(GEOCOORD_ENHANCEMENT) && defined(TARGET_OS_IPHONE)
    st = 0; // Appropriate comment
#endif
```

EimantasMorkunas: IMO not handling `CS_rwDictDir()` errors is fine.

The function returns non-zero if txt path is not writable and no fallback is provided. Looking at the rest of the function, it uses the path returned by function (which, if no fallback is used, is the same as dir as txt path, even on errors) to construct the path for binary file and checks if it exist. If the file does exist, the function does not do anything else. If it does not exist, it attempts to create the binary file from txt which would fail when opening a file for write if the dir is not writable, triggering the same error handling. So the only thing the disabled error handling accomplishes is failing earlier which also happens to fail in cases where it could continue without issues.

To add, when iTwin.js is using the cloud workspace, it does not cache the text version, only the binary file. So there's no way for the code to create the binary file anyways, it would error out on attempting to open the text file.

Finally, if you search for `CS_rwDictDir()` usages in the code base, you'll see that this is the only place that checks its return value, even though its used in similar create-a-binary-from-txt scenarios ([example](https://github.com/iTwin/imodel-native/blob/d251128b182048a212004bc2c13fb937c938f9d3/iModelCore/libsrc/csmap/Source/CS_osgm91.c#L424)).

tcobbs-bentley (approved):

aruniverse:
@EimantasMorkunas does this PR replace https://github.com/iTwin/imodel-native/pull/761?

EimantasMorkunas:
> @EimantasMorkunas does this PR replace #761?

Yes.

tcobbs-bentley:
@AlainRobertAtBentley @iTwin/itwinjs-core-admins Can we get a review of this PR?

aruniverse (approved):

aruniverse:

PR #954, opened 2025-01-08
static link with parasolid (winx64)

jffmarker:
To prevent frustrum collisions when integrating with other Apps. By statically linking on Windows, the addon will gets its own frustrum and copy of Parasolid functions to isolate itself from other Apps (when loaded in-process). This allows for greater interoperability when writing Connectors. As it stood today, Parasolid was always loaded when the addon initialized, so static linking should not have notable impact on loading and startup.

https://github.com/iTwin/imodel-native-internal/pull/721


bbastings (approved):

khanaffan (approved):

DanRod1999:

PR #956, opened 2025-01-08
Allow attach/detach db to IModelDb connection

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7530

Allow attach/detach db to active connection. This allow to combine data from more then on db using ECSQL.

Note: ConcurrentQuery has been updated to support attach detach as well.

https://github.com/iTwin/itwinjs-core/issues/7532


rschili:
Had a teams chat with Affan, I asked what happens if the attached ECDb is incompatible, like a different profile or incompatible schemas. It seems the code is not robust in handling that. We should add some more tests and safeguards.

iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -2786,6 +2786,36 @@ DbResult Db::DetachDb(Utf8CP alias) const
...
+        AttachFileInfo info;
+        info.m_alias = stmt.GetValueText(1);
+        info.m_fileName = stmt.GetValueText(2);
+        if (info.m_alias.EqualsIAscii("main")) {
+            info.m_type = AttachFileTypes::Main;
+        } else if (info.m_alias.EqualsIAscii("schema_sync_db")){
+            info.m_type = AttachFileTypes::SchemaSync;
+        } else if (info.m_alias.EqualsIAscii("ecchange")){
+            info.m_type = AttachFileTypes::ECChangeCache;
+        } else if (info.m_alias.EqualsIAscii("temp")){
```
tcobbs-bentley, line 2809: It looks like these specific aliases have special meaning. If I'm not misunderstanding, there should probably be documentation and code on the TS side forbidding their use as aliases in the calls to `attachDb`. It seems like if someone calls `attachDb("somefile", "main")`, the call will succeed, but that it will cause problems. I could be wrong about all of this, since I don't fully understand the native logic.

iModelCore/BeSQLite/PublicAPI/BeSQLite/BeSQLite.h
```diff
@@ -393,6 +393,21 @@ enum class WalCheckpointMode {
     Truncate=3, /* Like RESTART but also truncate WAL */
 };
 
+enum class AttachFileTypes {
```
tcobbs-bentley, line 396: Looking at other `enum class` types in BeSqlite.h, almost all are singular. Furthermore, all the other "...Type" ones are singular. I feel this should be `AttachFileType`.

rschili (approved):

iModelCore/BeSQLite/BeSQLite.cpp
khanaffan: No the call will fail. It down to sqlite which should fail the attach and error would be returned. This enum was for query reader api which must attach each of these over to a separate sqlite connection held by separate thread. I like to avoid attaching main/temp/sync_db to concurrent query secondary connection.

tcobbs-bentley: I'm saying that these special aliases should be documented in the two `attachDb` functions in itwinjs-core. Otherwise people will get an exception thrown at run-time and not understand why.

pmconne (approved):

iModelCore/BeSQLite/BeSQLite.cpp
khanaffan: ah ok will do

khanaffan: so i added docs to itwin.js side

PR #970, opened 2025-01-17
Wrong property types issue

Abhijeet-Bentley:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7599

- Added tests to check for code's behaviour if wrong property tags are encountered
- Expected behavior on encountering wrong property tag: 

1. Older version: Default to string
2. Current/Latest version: Throws ERROR
3. Future version: Default to string

- Added code fix which takes care of all above scenarios.


iModelCore/ecobjects/src/ECProperty.cpp
```diff
@@ -44,6 +44,11 @@ SchemaReadStatus ReadTypeNameWithPruning(pugi::xml_node node, ECPropertyP prop,
     // must come after prune check to make sure we don't default to string properties that should be pruned
     if (setTypeStatus == ECObjectsStatus::ParseError && ignoreParseErrors)
         {
+        // Unknown type encounter in the schema for the current version returning ERROR status
+        if (!(prop->GetClass().GetSchema().OriginalECXmlVersionGreaterThan(ECVersion::Latest)) && !(prop->GetClass().GetSchema().OriginalECXmlVersionLessThan(ECVersion::Latest)))
```
RohitPtnkr1996, line 48: Why not just check for ECVersion::Latest ?

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -1535,6 +1535,120 @@ TEST_F(ECSqlStatementTestFixture, IsNull)
...
+                                            <ECProperty propertyName="p2d" typeName="Point2d" />
+                                            <ECProperty propertyName="p3d" typeName="Point3d" />
+                                        </ECStructClass>
+                                        <ECEntityClass typeName="UseOfWrongPropertyTags">
+                                            <ECProperty propertyName="Struct" typeName="PrimStruct" />
+                                            <ECStructArrayProperty propertyName="Struct_Array" typeName="PrimStruct" />
+                                        </ECEntityClass>
+                                </ECSchema>)xml";
+
+        if (testCaseNumber == 1)
```
RohitPtnkr1996, line 1573: The for loop does effectively nothing.
Each iteration executes it's own separate code block.

So why don't you either remove the for loop and have separate blocks for each version ?
Or else you can optimize the for loop better to handle the different versions.

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -1535,6 +1535,120 @@ TEST_F(ECSqlStatementTestFixture, IsNull)
...
+                                        </ECEntityClass>
+                              </ECSchema>)xml";
+            }
+        
+        // Schema should always be deserialized successfully irrespective of the ECXml version
+        ASSERT_EQ(deserializationStatus, SchemaReadStatus::Success == ECSchema::ReadFromXmlString(schema, Utf8PrintfString(xmlSchema, majorVersion, minorVersion).c_str(), *context)) << "Test case number " << testCaseNumber << " failed at deserializing.";
+        ASSERT_EQ(deserializationStatus, schema.IsValid()) << "Test case number: " << testCaseNumber << " failed due to invalid schemas.";
+            
+        // Checking if the wrong property tags are defaulted to string type
+        if (testCaseNumber != 4)
```
RohitPtnkr1996, line 1627: It would be cleaner and more readable if you check deserializationStatus instead of the test case number.

iModelCore/ecobjects/test/NonPublished/SchemaComparerTests.cpp
```diff
@@ -1121,6 +1121,54 @@ TEST_F(SchemaCompareTest, CompareECClassIdentical)
...
+    LoadSchemasIntoContext(m_secondContext, secondSchemasXml);
+    
+    SchemaComparer comparer;
+    SchemaDiff changes;
+    comparer.Compare(changes, m_firstContext->GetCache().GetSchemas(),  m_secondContext->GetCache().GetSchemas());
+
+    ASSERT_EQ(1, changes.Changes().Count());
+
+    // We default to string when the property tag is wrong therefore, property change should be observed.
+    ASSERT_TRUE(changes.Changes()[0].Classes()[0].Properties()[0].IsChanged());
```
RohitPtnkr1996, line 1167: Here we are just checking that the property has changed.
Can you look a bit deeper to check if we can test exactly what is being changed ?

RohitPtnkr1996 (approved):

pmconne:
> TS tests' PR Link: https://github.com/iTwin/itwinjs-core/pull/7599

Does the itwinjs-core PR depend upon the changes made in this imodel-native PR?
If so, please link your PRs [correctly](https://dev.azure.com/bentleycs/iModelTechnologies/_wiki/wikis/iModelTechnologies.wiki/28868/imodel-native-PR-Validation).
Otherwise please explain the relationship between these two PRs.

Abhijeet-Bentley:
> > TS tests' PR Link: [iTwin/itwinjs-core#7599](https://github.com/iTwin/itwinjs-core/pull/7599)
> 
> Does the itwinjs-core PR depend upon the changes made in this imodel-native PR? If so, please link your PRs [correctly](https://dev.azure.com/bentleycs/iModelTechnologies/_wiki/wikis/iModelTechnologies.wiki/28868/imodel-native-PR-Validation). Otherwise please explain the relationship between these two PRs.

> These two PRs are not related but, are for the same work item. I have removed the link from the descriptions of both.

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -1535,6 +1535,94 @@ TEST_F(ECSqlStatementTestFixture, IsNull)
...
+                                            <ECProperty propertyName="p2d" typeName="Point2d" />
+                                            <ECArrayProperty propertyName="p3d" typeName="Point3d" />
+                                        </ECStructClass>
+                                        <ECEntityClass typeName="UseOfWrongPropertyTags">
+                                            <ECProperty propertyName="Struct" typeName="PrimStruct" />
+                                            <ECStructArrayProperty propertyName="Struct_Array" typeName="PrimStruct" />
+                                        </ECEntityClass>
+                                    </ECSchema>)xml";
+
+    for (const auto& [testCaseNumber, majorVersion, minorVersion, xmlSchemaVar, deserializationStatus, importStatus] : std::vector<std::tuple<unsigned int, unsigned int, unsigned int, Utf8CP, bool, bool>>
```
ColinKerr, line 1585: I suggest taking SchemaReadStatus and SchemaImportResult as inputs as it would make it easier to read the test cases and the error outputs.

The final use of importStatus would be bit more work but I think it's worth the improved readability on the rest

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -10527,8 +10615,52 @@ TEST_F(ECSqlStatementTestFixture, AliasedEnumProps)
...
+
+    stmt.Finalize();
+    CloseECDb();
+        
+    }
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+TEST_F(ECSqlStatementTestFixture, AliasedEnumPropsFor_V3_2)
```
ColinKerr, line 10661: Why does this schema fail to import?

iModelCore/ecobjects/src/ECProperty.cpp
```diff
@@ -44,6 +44,13 @@ SchemaReadStatus ReadTypeNameWithPruning(pugi::xml_node node, ECPropertyP prop,
     // must come after prune check to make sure we don't default to string properties that should be pruned
     if (setTypeStatus == ECObjectsStatus::ParseError && ignoreParseErrors)
         {
+        // Unknown type encounter in the schema for the current version returning ERROR status
+        if (prop->GetClass().GetSchema().GetOriginalECXmlVersionAsString() == ECSchema::GetECVersionString(ECVersion::Latest))
```
ColinKerr, line 48: This is wrong.

Today it's ok because we want to allow older (EC 2, 3.0 and 3.1) and newer (EC 3.3 and 3.4) to pass through but EC 3.2 to fail.  ECVersion::Latest is 3.2.

When ECVersion::Latest is 3.3 we will allow 3.2 to pass through.  I suggest using the ECSchema methods for comparing ECVersion and original ECXml version:
```cpp
    //! Returns true if the original xml version is greater or equal to the input ECVersion
    bool OriginalECXmlVersionAtLeast(ECVersion version);
    //! Returns true if the original xml version is less than the input ECVersion
    bool OriginalECXmlVersionLessThan(ECVersion version);
    //! Returns true if the original xml version is greater than the input ECVersion
    bool OriginalECXmlVersionGreaterThan(ECVersion version);
```

iModelCore/ecobjects/test/NonPublished/SchemaComparerTests.cpp
```diff
@@ -1121,6 +1121,54 @@ TEST_F(SchemaCompareTest, CompareECClassIdentical)
 //----------------------------------------------------------------------------------------
 // @bsimethod
 //---------------+---------------+---------------+---------------+---------------+--------
+TEST_F(SchemaComparerXmlTests, CompareSchemasWithWrongPropertyTags)
+    {
+    //Test created for 3.1 version of schema specifically 
+    //to observe defaulting behavior for and when the property tags are wrong
```
ColinKerr, line 1127: I've seen this in a few other places but you use 'tag' when I think you mean property type: https://www.itwinjs.org/bis/ec/ec-property/#ecproperty

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
Abhijeet-Bentley: I have incorporated these changes and now, added better error outputs.

iModelCore/ecobjects/src/ECProperty.cpp
Abhijeet-Bentley: Now we check for versions they belong to [3.2, Latest] inclusive or not and, if they do we return "Error" status.

rschili (approved):

ColinKerr (approved):

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
ColinKerr: I see the typeName is Domain not Domains ... I missed that when I first read the schema.  I wonder if this was an unintentional mistake when the original test case was written.

Maybe it just so happened that the enum they misspelled was of string type so the writer never noticed.  Looking at the test name and the contents of the test it looks like they were just testing that you could use an alias on an enum property of each backing type.  If so then the misspelling actually makes the test less good.

@rschili do you agree?  If so I think the right move is to fix the original test so the correct enum type name is used.  I don't think there is anything special about a property type name that is close to an enum vs any other unknown type.

@RohitPtnkr1996 do we have a test case for an enum which is of an unknown type?  e.g. not a string or int.

rschili: Yeah I agree, that's definitely 95% sure a misspell.

RohitPtnkr1996: We have tests for it on both native and itwinjs, however, I only covered future newer ecxml versions when I added them.

I don't think we had any that explicitly tested the enum's defaulting behavior for the versions <= 3.2, or they would've failed after Abhijeet's changes.

RohitPtnkr1996: Yeah we have tests on both native and itwinjs for enums with an unknown type.
It is treated as a hard error on versions <= 3.2.
On newer, we default the enum's backing type to a string.

But for this, I think it was a typo and quite unfortunate that the ECProperty simply defaulted to a string and the true intention of the schema was missed for so long.

Abhijeet-Bentley: I have now fixed the test case.

PR #972, opened 2025-01-22
Fix text not appearing when requested font is not in font table

DanRod1999:
[Resolves](https://github.com/iTwin/itwinjs-backlog/issues/1304)
When a font is requested for text but it is not in the dgn_font table the text does not get displayed.

I believe it is due to the fontId being set to zero in itwinjs-core when the requested font is not found. [Code here](https://github.com/iTwin/itwinjs-core/blob/master/core/backend/src/TextAnnotationLayout.ts#L75)

Because of this when generating the opcodes for the geom stream, the invalid id is caught and the opcode for drawing a text view is not added to the Geom Stream io collection.

I do not know if this is the best solution, or if core should not use 0 for this fontId.


markschlosseratbentley:
What does font id 0 mean? If we do not throw an error, and instead actually use font id 0 alongside the text opcode, have you seen what font it will use?

markschlosseratbentley:
> What does font id 0 mean? If we do not throw an error, and instead actually use font id 0 alongside the text opcode, have you seen what font it will use?

Never mind - I saw your other comment elsewhere about the fallback font it will use.  Can we rely on this being the case always?  (I hope so).

pmconne (changes requested):
See my comments on the [issue](https://github.com/iTwin/itwinjs-backlog/issues/1304). We don't want to allow people to insert text strings with font Id 0.

PR #973, opened 2025-01-22
Fix issue where query get interrupted during prepare.

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7592

Fixes: https://github.com/iTwin/itwinjs-core/issues/7549
- Do not interrupt the query during preparation. The query can only be interrupted during the `step()` function.
- Add an option to prepare ECSQL using a thread database connection. Experimental.
  - Default to false. Not enabled.
  - Thread does not wait to acquire lock on primary connection to prepare statement. Enhance faster.
  - Side effect more memory as each connection load its on EC Schema when preparing ECSQL.
- Query statistics now also return preparation time in nanoseconds.
- Introduce a feature to automatically shut down concurrent queries when idle for a set time:
  - Defaults to 30 minutes.
  - Closes all threads and releases resources used by concurrent queries.
  - Automatically wakes up when a new request comes in.
  - New error code is added `ShuttingDown` which cause frontend to retry.
- Allow setting the memory map I/O size for each connection, defaulting to 0 (not used):
  - This is an experimental option. It improves query performance on regular files but is untested with cloud SQLite.
- The statement cache size per thread can now be configured, with the default remaining at 40 (unchanged from previous).
- Fix a issue where `delay` use for testing cause every query to delay at least once even when not testing.
- Added `allowTestingArgs` help simulate a condition for testing. 
- Found another issue which cause thread dead lock during closing connection. `ConcurrentQuery` config is not stored on db connection anymore. 
The default config look as following. Only new change is `autoShutdowWhenIdealForSeconds: 1800`

```ts
{
  workerThreads: 4,
  requestQueueSize: 2000,
  ignorePriority: false,
  ignoreDelay: true,
  doNotUsePrimaryConnToPrepare: false,
  autoShutdowWhenIdlelForSeconds: 1800,
  statementCacheSizePerWorker: 40,
  monitorPollInterval: 1000,
  memoryMapFileSize: 0,
  allowTestingArgs: false,
  globalQuota: { time: 60, memory: 8388608 }
}
```


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -80,9 +80,13 @@ std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool
     ErrorListenerScope err_scope(const_cast<ECDb&>(m_conn.GetPrimaryDb()));
     if (usePrimaryConn)
         status = newCachedAdaptor->GetStatement().Prepare(m_conn.GetPrimaryDb(), ecsql, !suppressLogError);
-    else
-        status = newCachedAdaptor->GetStatement().Prepare(m_conn.GetPrimaryDb().Schemas(), m_conn.GetDb(), ecsql, !suppressLogError);
-
+    else {
+        if (m_doNotUsePrimaryConnToPrepare) {
```
rschili, line 84: Just realized we already have the usePrimaryConn flag.
Putting a second m_doNotUsePrimaryConnToPrepare flag alongside seems really confusing. Especially since one is named positive and the other negated.
I suggest we call this m_allowPrepareOnSeparateThread.
I also wondered why one is given to this method while the other is pulled from config, but I guess one is global while the other can be set on a per-query basis.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -80,9 +80,13 @@ std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool
     ErrorListenerScope err_scope(const_cast<ECDb&>(m_conn.GetPrimaryDb()));
     if (usePrimaryConn)
         status = newCachedAdaptor->GetStatement().Prepare(m_conn.GetPrimaryDb(), ecsql, !suppressLogError);
-    else
-        status = newCachedAdaptor->GetStatement().Prepare(m_conn.GetPrimaryDb().Schemas(), m_conn.GetDb(), ecsql, !suppressLogError);
-
+    else {
+        if (m_doNotUsePrimaryConnToPrepare) {
+            status = newCachedAdaptor->GetStatement().Prepare(m_conn.GetDb(), ecsql, !suppressLogError);
```
rschili, line 85: Should this use the overload taking a schemaManager? Its documentation says it's for multi-threading scenarios. If not, can you explain?

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
khanaffan: I agree that the names can be confusing. The `usePrimaryConn` flag indicates whether to run the query synchronously. The other flag controls how to prepare the query if it runs asynchronously, using either the primary schema cache or the thread connection cache. By default, this feature is off. During testing, I found an issue where, despite each thread having its own connection, there are moments when they serialize, causing longer result production times instead of being faster. I will log an issue, as it seems there's something causing the simultaneous use of `ECSqlStatement` to not be entirely concurrent. My guess is that it might be related to the functions we register, but this is beyond the scope of this PR.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
khanaffan: No the whole point is not to use primary connection and see if we can prepare more in less by not competing for primary statement.

ColinKerr (approved):

calebmshafer (approved):

khanaffan:
backport-to-release/4.11.x

khanaffan:
@Mergifyio backport release/4.11.x release/4.10.x

khanaffan:
@Mergifyio backport release/4.10.x  release/4.11.x

PR #974, opened 2025-01-23
Presentation: Fix incorrect class grouping nodes' grouped instances count

grigasp:
When a hierarchy specification selects from two base classes of the same instance (e.g. through the use of a mixin), `groupedInstancesCount` of class grouping nodes was being duplicated.

1. Improved hierarchies validation in tests to compare expected instances' count with the actual count set on grouping nodes.
2. Fixed navigation query builder to filter-out duplicate select classes, when the duplication happens due to class grouping.
3. Tests' improvement detected a bug in hierarchies update, where we weren't updating grouping nodes after deleting grouped instances. Fixed that as well.


grigasp:

grigasp:
@mergifyio backport release/4.10.x

JonasDov (approved):

PR #977, opened 2025-01-27
Handle unknown mapping strategies in newer imodels

RohitPtnkr1996:
If we have a newer iModel which has a class with a mapping strategy unknown to the current ECDb runtime, the class will be loaded but, however, if we try to query it, we will end up with an "...unsupported map strategy...class wasn't loaded..." error.

Unlike the class modifier or relationship strength fields, a class's mapping strategy does not get checked when the class gets loaded.

This PR adds a new value to the MapStrategy Enum to handle an unknown strategy.
If we try to query this class, we should get a null view even if we are successful in loading the property map (due to the general uncertainty with an unknown mapping strat).


RohitPtnkr1996:
@khanaffan will you please take a look ?

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
```diff
@@ -250,6 +250,62 @@ BentleyStatus DbMapSaveContext::TryGetPropertyPathId(PropertyPathId& id, ECN::EC
     return SUCCESS;
     }
 
+namespace
+    {
+    static std::vector<Utf8String> classIdList;
+    Utf8String GetClassIdListAsString() { return BeStringUtilities::Join(classIdList, ","); }
+
+    void CollectAllBaseClassIds(ECClassCR ecClass)
```
rschili, line 258: There is a better way to do this running a sql statement on ec_cache_ClassHierarchy. There may be a public method for this somewhere, or at least there should be examples in the ecdb code.

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
```diff
@@ -283,12 +339,64 @@ BentleyStatus DbClassMapLoadContext::Load(DbClassMapLoadContext& loadContext, Cl
                 return ERROR;
         }
 
+    auto isMappingUnknown = false;
+    Utf8String logMessage;
```
rschili, line 343: Would it be easier instead of using logMessage, to just directly log in the spots where it is assigned? It's not used for anything other than eventually calling log.warning on it

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
```diff
@@ -283,12 +339,64 @@ BentleyStatus DbClassMapLoadContext::Load(DbClassMapLoadContext& loadContext, Cl
...
+        if (!relConstraintCheckStmt.IsValid())
+            return ERROR;
+
+        while (relConstraintCheckStmt->Step() == BE_SQLITE_ROW)
+            {            
+            if (const auto constraintClass = schemaManager.GetClass(relConstraintCheckStmt->GetValueId<ECClassId>(0)); constraintClass != nullptr)
+                {
+                if (IsMappingUnknownInBaseOrConstraintClasses(schemaManager, *constraintClass, logMessage))
+                    {
+                    LOG.warningv(logMessage.c_str());
```
rschili, line 395:
```suggestion
                    LOG.warning(logMessage.c_str());

```

iModelCore/ECDb/ECDb/MapStrategy.h
```diff
@@ -18,7 +18,8 @@ enum class MapStrategy
     TablePerHierarchy = 2,
     ExistingTable = 3,
     ForeignKeyRelationshipInTargetTable = 10,
-    ForeignKeyRelationshipInSourceTable = 11
+    ForeignKeyRelationshipInSourceTable = 11,
+    UnknownMapping = 99
```
rschili, line 22: I'm wondering if there is a better name for this. At least we should put a comment to indicate what it means. "Unknown" could be anything, but this is only used if the ECVersion is higher than latest and the mapping is not known. I'm thinking "UnrecognizedECVersion", "ExceedsLatestECVersion", "NewerThanSupported"... something like that.

iModelCore/ECDb/ECDb/ViewGenerator.cpp
```diff
@@ -616,31 +616,34 @@ BentleyStatus ViewGenerator::RenderEntityClassMap(NativeSqlBuilder& viewSql, Con
...
-                        view.Append(" ").Append(filterSQL.c_str());
+                    const auto polymorphicInfo = isSelectFromView ? ctx.GetAs<SelectFromViewContext>().GetPolymorphicInfo() : PolymorphicInfo::All();
+                    Utf8String filterSQL;
+                    if (SUCCESS != GenerateECClassIdFilter(filterSQL, classMap, partition->GetTable(), classIdPropertyMap->GetColumn(), polymorphicInfo))
+                        return ERROR;
+
+                    if (!filterSQL.empty())
+                        {
+                        if (polymorphicInfo.IsOnly())
+                            view.Append(" WHERE ").Append(filterSQL.c_str());
```
rschili, line 635: Not criticising your code, it was already this deeply nested before. We are 6 layer deep in nesting here in addition to this method being over 100 lines. I feel the strong urge to have this cleaned out...

rschili (approved):

iModelCore/ECDb/ECDb/ViewGenerator.cpp
RohitPtnkr1996: Dunno why all these lines are being picked up.
I've only updated the if conditions below !

Yeah I get your point, I'll take a look at getting this cleaned up and a bit clearer to read.

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
RohitPtnkr1996: Done !

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
RohitPtnkr1996: Done !

iModelCore/ECDb/ECDb/ClassMapPersistenceManager.cpp
RohitPtnkr1996: Done !

iModelCore/ECDb/ECDb/MapStrategy.h
RohitPtnkr1996: I've changed it to UnsupportedByECVersion.
MapStrategy::UnsupportedByECVersion will make it clear when reading.

rschili (approved):

PR #980, opened 2025-01-29
Updated test code for the upcoming BisCore 01.00.17

RohitPtnkr1996:
The upcoming BisCore.01.00.17 has the “MemberPriority” property moved from “ElementGroupsMembers” to the abstract class “ElementRefersToElements”.
The updated test cases in the PR have the “schemaLockHeld” import option set explicitly to false and remapping the data to a new column is a data transform, which is causing the tests to fail.

Any schema that references BisCore.1.0.17 will now need the schema lock in hand in order to import it.

@khanaffan, will you please take a look ?


rschili (approved):

rschili:
I've been thinking.... instead of just allowing this in general, we may narrow it down and just allow it for the particular case, so we still catch other schema updates which do data transform without us knowing.
But it would be hard to come up with a reliable mechanism across all tests, plus I'm not sure we want to be notified and have to adjust tests every time this happens in the future. So I'm fine with this fix.

khanaffan (approved):

Issue #897, opened 2024-10-22
ECSQL with VALUES(...),(...) has a limit


JonasDov:
The following ECSQL statement:
```sql
SELECT * FROM (VALUES(1), (2), (3))
```
Gets converted to:
```sql
SELECT [K0] FROM (SELECT 1 [K0] UNION ALL SELECT 2 UNION ALL SELECT 3)
```
This works fine when there are 500 or fewer statements. But with more than 500 statements, the following error is returned:
`SQLite statement failed to prepare: BE_SQLITE_ERROR too many terms in compound SELECT (BE_SQLITE_ERROR)`.
For example passing the following ECSQL:
```sql
SELECT * FROM (VALUES(1), (2), (3), ..., (499), (500), (501))
```
Results in the error above.

This is because `SELECT ... UNION ALL ...` has a limit of 500 set, see: https://www.sqlite.org/limits.html#:~:text=Maximum%20Number%20Of%20Terms%20In%20A%20Compound%20SELECT%20Statement
And VALUES(...),(...) should not have a limit, see: https://www.sqlite.org/lang_select.html#values:~:text=6.-,The%20VALUES%20clause,-The%20phrase%20%22VALUES

aruniverse:
@JonasDov is this still relevant?

JonasDov:
Personally, not using `VALUES` keyword, but the bug still exists.

grigasp:
IMO, at the very least we should document this limit somewhere. But I don't see a reason why we should have this limit in the first place.

PR #1000, opened 2025-02-11
support for Values keyword

naronchen:
#897


iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -2788,22 +2787,24 @@ BentleyStatus ECSqlParser::ParseValuesCommalist(std::unique_ptr<SelectStatementE
...
-            prevSelectExp = std::make_unique<SelectStatementExp>(std::move(singleSelect));
-        else
+        // return error if each valueExpList dont have the same number of elements
+        if (i > 0 && valueExpList.size() != rowValuesExpList[0]->GetChildrenCount())
             {
-            std::unique_ptr<SelectStatementExp> selectExp = nullptr;
-            selectExp = std::make_unique<SelectStatementExp>(std::move(singleSelect), SelectStatementExp::CompoundOperator::Union, true /*isAll*/, std::move(prevSelectExp));
-            prevSelectExp = std::move(selectExp);
+            // BeAssert(false && "All rows in VALUES clause must have the same number of elements");
+            Issues().Report(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSQL, ECDbIssueId::ECDb_0492, "All rows in VALUES clause must have the same number of elements");
```
naronchen, line 2798: In this case, should it be `BeAssert` or `Issues()`?

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -3437,7 +3438,7 @@ BentleyStatus ECSqlParser::ParseValueExp(std::unique_ptr<ValueExp>& valueExp, OS
                 case OSQLParseNode::value_creation_fct:
                     return ParseValueCreationFuncExp(valueExp, parseNode);
                 default:
-                    Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSQL, ECDbIssueId::ECDb_0493,
+                    Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSQL, ECDbIssueId::ECDb_0737, // @todo - naron: what issue id should be here
```
naronchen, line 3441: where can I find what each issueId means?

iModelCore/ECDb/ECDb/ECSql/ECSqlPropertyNameExpPreparer.cpp
```diff
@@ -388,6 +388,17 @@ ECSqlStatus ECSqlPropertyNameExpPreparer::PrepareInSubqueryRef(NativeSqlBuilder:
                     ECSqlFieldFactory::CreateField(ctx, exp.GetParent()->GetAsCP<DerivedPropertyExp>(), ctx.GetCurrentScope().GetNativeSqlSelectClauseColumnCount());
                 break;
                 }
+            case Exp::Type::SqlColumnName:
+                {
+                    NativeSqlBuilder sqlSnippet;
```
naronchen, line 393: i  can also just append the resolved path here?

iModelCore/ECDb/ECDb/ECSql/SelectStatementExp.cpp
```diff
@@ -781,6 +781,10 @@ Exp::FinalizeParseStatus SelectClauseExp::_FinalizeParsing(ECSqlParseContext& ct
     {
     if (mode == Exp::FinalizeParseMode::BeforeFinalizingChildren)
         {
+        // if child is a sqlColumnNameExp, nothing needed to be done
+        if (Contains(Exp::Type::SqlColumnName))
+            return FinalizeParseStatus::Completed;
```
naronchen, line 786: I skipped it here and handle the expand in SubqueryRefExp::_ExpandSelectAsterisk
Im not sure if this is the best way to do this, this means Im leaving RowValueConstructorListExp::_ExpandSelectAsterisk implemented but not triggered.

iModelCore/ECDb/ECDb/ECSql/ECSqlSelectPreparer.cpp
```diff
@@ -374,7 +374,10 @@ ECSqlStatus ECSqlSelectPreparer::PrepareDerivedPropertyExp(NativeSqlBuilder::Lis
         if (exp.GetExpression()->GetType() == Exp::Type::NavValueCreationFunc ||
             (exp.GetExpression()->GetType() == Exp::Type::PropertyName && exp.GetExpression()->GetAs<PropertyNameExp>().IsPropertyRef() && exp.GetExpression()->GetAs<PropertyNameExp>().GetPropertyRef()->LinkedTo().GetExpression()->GetType() == Exp::Type::NavValueCreationFunc))
             return ECSqlStatus::Success; // it was already generated by PrepareNavValueCreationFuncExp
-
+        if (innerExp->GetType() == Exp::Type::PropertyName && 
+            innerExp->GetAs<PropertyNameExp>().IsPropertyRef() &&
+            innerExp->GetAs<PropertyNameExp>().GetPropertyRef()->IsSqlColumnNameExp())
+            return ECSqlStatus::Success;
```
naronchen, line 380: I skipped create field here for now but Im not sure. In this case for inline talbes, should fields be created here?

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -12749,6 +12749,98 @@ TEST_F(ECSqlStatementTestFixture, InsertUsingOnlyAndAll)
...
+
+    ECSqlStatement simpleSelect;
+    std::string query = "select column1 from (values(1), (2), (3))";
+    ASSERT_EQ(ECSqlStatus::Success, simpleSelect.Prepare(m_ecdb, query.c_str()));
+    
+    // EXPECT_EQ(BE_SQLITE_ROW, simpleSelect.Step());
+    // ASSERT_EQ(1, simpleSelect.GetColumnCount());
+    // ECSqlColumnInfo const& colInfo = simpleSelect.GetColumnInfo(0);
+    // Utf8String colName = colInfo.GetPropertyPath().ToString();
+    // ASSERT_STREQ("column1", colName.c_str());
```
naronchen, line 12763: right now even though its passing all the prepare statement, it has null property and no columns returning when I do step(). Probably it has to do with how results from sqlite are interpreted? where to look for that?

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -12749,6 +12749,98 @@ TEST_F(ECSqlStatementTestFixture, InsertUsingOnlyAndAll)
...
+    query = "SELECT d.column1, d.column2, x.column1, x.column2 FROM (VALUES(1,1), (2,2)) d, (VALUES(4,5), (5,6)) x";
+    ASSERT_EQ(ECSqlStatus::Success, multipleInlineTables.Prepare(m_ecdb, query.c_str()));
+
+    ECSqlStatement withColumnAlias;
+    query = "SELECT column1 myColumn FROM (VALUES(1), (2))";
+    ASSERT_EQ(ECSqlStatus::Success, withColumnAlias.Prepare(m_ecdb, query.c_str()));
+
+    ECSqlStatement nestedValues;
+    query = "SELECT column1 FROM (SELECT column1 FROM (VALUES(1), (2), (3)))";
+    ASSERT_EQ(ECSqlStatus::Success, nestedValues.Prepare(m_ecdb, query.c_str()));
```
naronchen, line 12805: when it comes to nested values, it uses nested alias K0, K1... making the columns come back named 1,2,....

naronchen (commented):
WIP

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
ColinKerr: Add a new issue id for each new issue, if it's an existing issue it's id should remain the same

khanaffan (approved):

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -13009,6 +13009,381 @@ TEST_F(ECSqlStatementTestFixture, Testing_Table_Valued_Functions_without_schemaN
         }
     }
 
+TEST_F(ECSqlStatementTestFixture, ValuesClauseTest) {
```
khanaffan, line 13012: Here is some test we should test
1. `SELECT column1, column2 FROM (VALUES(1,2)) UNION SELECT column1, column2 FROM (VALUES(3,4));`
2. `SELECT * FROM (SELECT column1 a, column2 b FROM  (VALUES(1,2)))`
3. `SELECT * FROM (VALUES(1,2)) a, (VALUES(2,3)) b`

khanaffan: I also noted that following, but column1_1 or column2_1 cannot be accessed in WHERE clause 
```
iModelConsole> SELECT * FROM (VALUES(1,2)) a, (VALUES(2,3)) b;

column1           |column2           |column1_1         |column2_1
---------------------------------------------------------------------------
1                 |2                 |2                 |3
```

khanaffan: There is no type checking e.g. Normally if user use different type for same column in different rows and its constant literal then we should fail if type is different.

```
iModelConsole> .metadata SELECT * FROM (VALUES('dd',2),('333',4)) a;

Column metadata
===============
Index   Name/PropertyPath                   DisplayLabel                        System   Type                      Root class                     Root class alias
------------------------------------------------------------------------------------------------------------------------------------------------------------------
  0     column1                             column1                             no       String                    generated
  1     column2                             column2                             no       Long                      generated
```

khanaffan (changes requested):
I added few mandory work that include add test and validate type of constant literal provided as values if more than one row is specified.

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
akshay05n: >There is no type checking e.g. Normally if a user uses a different type for the same column in different rows and its constant literal then we should fail if the type is different.

I checked this in SQLite playground and `SELECT * FROM (VALUES('dd',2),('333',4)) a;` query passes on 4 different playgrounds.

SQLite uses dynamic typing and a concept called [type affinity](https://www.sqlite.org/datatype3.html#type_affinity). This means that even if you declare a column to be of a certain type (e.g., INTEGER), SQLite will not strictly enforce it. You can insert different types (e.g., string, float, blob) into that column without errors.

PR #1009, opened 2025-02-17
Upgrade FlatBuffers to 1.12.0

yuryleb2:
- Upgrade FlatBuffers SDK copy from 1.0.0 to 1.12.0 (released Mar 12, 2020) from https://github.com/google/flatbuffers/releases/tag/v1.12.0
- Restore TemporaryForceDefaults struct, FlatBufferBuilder::IsForcingDefaults() method and has_XXX() methods for generated FB Table classes from previous FlatBuffers 1.0.0 modifications
- Update prebuilt FlatBuffers compiler and its make files
- Regenerate C++ stuff of GeomLibsSerialization; TypeScript stuff was unchanged so as it was built with the same FlatBuffers compiler v1.12.0 version
- Regenerate FlatBuffers stuff of DgnCore with excluding custom emitVectorTypedefs attrubutes (it was also Bentley extension - generated vector typedefs were just moved to AnnotationFrameStylePersistence.h and other annotation headers) and adding default zero values for enums to Annotations.fbs and Dimension.fbs (only attribute removed); DgnFont.fbs, ElementGraphics.fbs and TextString.fbs were kept unchanged
- Use FlatBufferBuilder::CreateUninitializedVector() to store Parasolid body serialization bits; former CreateVector() call copies serialization buffer byte-by-byte that works very slow for bodies greater 1M


dassaf4 (approved):
Approving for geomlibs. See original PR [here](https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel02/pullrequest/500048).

aruniverse:
> TypeScript stuff was unchanged so as it was built with the same FlatBuffers compiler v1.12.0 version

So just to confirm, both the ts and native flat buffer versions need to be synchronized.
https://github.com/iTwin/itwinjs-core/blob/master/core/geometry/package.json#L52

pmconne (approved):

yuryleb2:
I think I can manage this. Latest FlatBuffer 25.1.24 has this TableKeyComparator operator also deleted and IMHO we can do the same as our yet another modification.

rschili:
@yuryleb2 verified, it works fine! That was quick, thanks.

Issue #997, opened 2025-02-10
Have schema merger return a real status


c-macdonald:
In a nutshell, the IFC4 schema has a relationship class with 3 target classes.  These classes don't share a common base class (in fact, none of them have a base class).  This means that EC 3.0 can't create an abstract constraint class.  We have 2 different versions of the IfcApplicationDomain schema. They each reference the IFC4 schema.  The first thing the new merger does is make a copy of the input schema and all of its references.  It fails to copy the IFC4 referenced schema because of the illegal relationship class.
 
So, technically, not a problem with the merger.  The ironic part is that we don't care at all about the targets.  Once the schemas are consolidated, then we BISify them. And one of the steps of bisification is to remove all target constraints and just add the base Bis:Element class.  Since we aren't creating new relationships, we don't need to actually constrain them. We just need to be able to read existing relationships.
 
I can work around this by pre-processing the schema and removing extra constraint classes before merging.  But that is potentially a lot of overhead that isn’t necessary in most cases.  All of this leads up to my actual question: can we change the SchemaMerger::MergeSchemas to return an actual status instead of just BentleyStatus?  If it returned that the actual ECObjectsStatus, then I can try to fix certain cases. I’d start with this one and look for RelationshipConstraintsNotCompatible.  If so, I’d then run through the schema and its references and remove the extra constraint classes.

ColinKerr:
We can investigate modifying SchemaMerger::MergeSchemas to return the full ECObjectsStatus

aruniverse:
@cammac1803 can you provide an update on this issue? Is this still relevant or needed?

PR #1010, opened 2025-02-19
Fix for MergeSchemas behaviour when merging schemas with multiple constraint classes

Abhijeet-Bentley:
Issue Link: https://github.com/iTwin/imodel-native/issues/997

This is the first task of the said work item.

- MergeSchemas function when merging schemas with multiple constraint classes throws error (Bentley Statuses).
- Now, we have skipped the validations while copying Relationship Classes.

1. Added tests to check CopyClass() and MergeSchemas() behaviour.
2. Added a skipValidation flag inside SchemaMergeOptions object.
3. This flag gets passed along all the way to CopyClass() method.
4. Inside CopyClass(), we instantiated a Relationship Class using CreateRelationshipClass() constructor, where we pass this new skipValidation flag to override the m_verify flag on Relationship Class object in turn, turning off the validations while copying.
5. Made some adjustments to the GetOrCopyReferencedItemforCopy method to accomodate changes made to CopyClass() method.


iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -3831,7 +3831,7 @@ friend struct SchemaJsonWriter;
     //! @param[in]  sourceClass The class to copy
     //! @param[in]  copyReferences If true the method will copy types from the source schema into the target schema, if they do not already exist. If false, there will be a schema reference created to the source schema if necessary.
     //! @param[in]  newName  If not nullptr, this name will be used as the new name instead of the original name
-    ECOBJECTS_EXPORT ECObjectsStatus CopyClass(ECClassP& targetClass, ECClassCR sourceClass, bool copyReferences = false, Utf8CP newName = nullptr);
+    ECOBJECTS_EXPORT ECObjectsStatus CopyClass(ECClassP& targetClass, ECClassCR sourceClass, bool copyReferences = false, Utf8CP newName = nullptr, bool skipValidations = false);
```
rschili, line 3834: let's call it skipValidation (no plural) to match how it's called in SchemaContext

iModelCore/ecobjects/PublicApi/ECObjects/SchemaMerger.h
```diff
@@ -51,10 +51,15 @@ struct SchemaMergeResult final
 struct SchemaMergeOptions final
     {
 private:
+    bool skipValidations = false;
```
rschili, line 54:
```suggestion
    bool m_skipValidation = false;

```

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -1579,7 +1579,7 @@ ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy(const Item & itemWithRe
     if (nullptr == refForCopy)
         {
         if (willCopyRef)
-            status = (this->*copyItem)(refForCopy, *startingRef, copyReferences, nullptr);
+            status = (this->*copyItem)(refForCopy, *startingRef, copyReferences, nullptr, std::forward<Args>(args)...);
```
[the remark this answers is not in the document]
rschili: Let's clean that up a bit. I don't like that this new approach is not typesafe, just forwarding arguments. Better put this in the header:

```cpp
    template<typename T>
    using GetItemFunction = T* (ECSchema::*)(Utf8CP name) const;
    template<typename T>
    using CopyItemFunction = ECObjectsStatus (ECSchema::*)(T*& target, const T& source, bool copyReferences, Utf8CP newName, bool skipValidation);
    
    
    template<typename Item, typename RefItem>
    ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy
        (
            const Item& itemWithRef,
            RefItem*& refForCopy,
            RefItem const* startingRef,
            bool copyReferences,
            ECSchemaElementType refItemType,
            GetItemFunction<RefItem> getItem,
            CopyItemFunction<RefItem> copyItem
        );
```

For that we will have to adjust the signatures of all CopyXXX methods, but it should be fine, just make it a default parameter which defaults to true if not provided.

iModelCore/ecobjects/PublicApi/ECObjects/SchemaMerger.h
```diff
@@ -51,10 +51,15 @@ struct SchemaMergeResult final
 struct SchemaMergeOptions final
     {
 private:
+    bool m_skipValidation = false;
     BeJsDocument m_json;
 public:
     SchemaMergeOptions() {}
 
+    //! Setting this flag instructs the merger to skip validations
+    void SetSkipValidations(bool flag) { m_skipValidation = flag; }
```
rschili, line 60:
```suggestion
    void SetSkipValidation(bool flag) { m_skipValidation = flag; }

```

iModelCore/ecobjects/PublicApi/ECObjects/ECSchema.h
```diff
@@ -3831,7 +3831,7 @@ friend struct SchemaJsonWriter;
     //! @param[in]  sourceClass The class to copy
     //! @param[in]  copyReferences If true the method will copy types from the source schema into the target schema, if they do not already exist. If false, there will be a schema reference created to the source schema if necessary.
     //! @param[in]  newName  If not nullptr, this name will be used as the new name instead of the original name
-    ECOBJECTS_EXPORT ECObjectsStatus CopyClass(ECClassP& targetClass, ECClassCR sourceClass, bool copyReferences = false, Utf8CP newName = nullptr);
+    ECOBJECTS_EXPORT ECObjectsStatus CopyClass(ECClassP& targetClass, ECClassCR sourceClass, bool copyReferences = false, Utf8CP newName = nullptr, bool skipValidation = false);
```
rschili, line 3834: add the documentation for the new parameter

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -1588,7 +1588,21 @@ ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy(const Item & itemWithRe
...
+        ECSchemaP refSchema;
+        ECObjectsStatus status = getOrAddReferencedSchema(this, startingRef->GetSchema(), refSchema, willCopyRef);
+        if (ECObjectsStatus::Success != status)
+            return status;
+    
+        refForCopy = (refSchema->ECSchema::GetClassP)(startingRef->GetName().c_str());
+        if (nullptr == refForCopy)
+            {
+            if (willCopyRef)
+                status = CopyClass(refForCopy, *startingRef, copyReferences, nullptr, false);
```
rschili, line 1601: Setting the flag always to false here is wrong. The method needs to take the flag provided into CopyClass, and pass it along into subsequent calls.

iModelCore/ecobjects/test/NonPublished/SchemaMergerTests.cpp
```diff
@@ -4493,6 +4493,67 @@ TEST_F(SchemaMergerTests, MergeRelationshipConstraints)
 /*---------------------------------------------------------------------------------**//**
 * @bsitest
 +---------------+---------------+---------------+---------------+---------------+------*/
+TEST_F(SchemaMergerTests, MergeRelationshipConstraintsWithNoBaseClass)
```
rschili, line 4496: I suggest not using a SchemaReference for this, put the Relationship right into the actual test schemas.
I'd also like to see some variations: Merging the Relationship if one side has it and the other does not. Merging it if it's in both sides of the schema and identical. Merge it if it's on both sides but has different classes. Merge it if it's on both sides with 1 Constraint class each, so the merge result has two classes.
Also the test should check both, SkipValidation true and false to make sure they behave accordingly

iModelCore/ecobjects/test/NonPublished/CopyTests/ClassCopyTests.cpp
```diff
@@ -284,6 +284,45 @@ TEST_F(ClassCopyTest, CopyRelationshipClassWithConstraintClassesInRefSchemaAndOr
...
+    ECSchemaPtr copySchema;
+    ECObjectsStatus copyStatus = ECSchema::CreateSchema(copySchema, "CopySchema", "cs", 1, 0, 0);
+    ASSERT_EQ(ECObjectsStatus::Success, copyStatus) << "Failed to create copy schema";
+    ECClassP ecClass = schema->GetClassP("foo");
+    ASSERT_NE(ecClass, nullptr);
+    ASSERT_TRUE(ecClass->IsRelationshipClass());
+
+    ECClassP copiedClass;
+    copyStatus = copySchema->CopyClass(copiedClass, *ecClass, true, nullptr, true);
+    ASSERT_EQ(ECObjectsStatus::Success, copyStatus) << "Failed to copy class";
```
rschili, line 323: Add another call to CopyClass with false and expect it to fail

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -1586,9 +1586,23 @@ ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy(const Item & itemWithRe
     return status;
     }
 
-ECObjectsStatus ECSchema::GetOrCopyReferencedClassForCopy(ECClassCR classWithRef, ECClassP& refForCopy, ECClassCP startingRef, bool copyReferences)
+ECObjectsStatus ECSchema::GetOrCopyReferencedClassForCopy(ECClassCR classWithRef, ECClassP& refForCopy, ECClassCP startingRef, bool copyReferences, bool skipValidation)
```
rschili, line 1589: I think you forgot to update the header file, or is it just not showing correctly for me?

iModelCore/ecobjects/src/ECSchema.cpp
```diff
@@ -1566,8 +1566,8 @@ bool areFromSameSchema (NamedItemA const* itemA, NamedItemB const* itemB)
     return itemA->GetSchema().GetName().Equals(itemB->GetSchema().GetName().c_str());
     }
 
-template<typename Item, typename RefItem>
-ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy(const Item & itemWithRef, RefItem *& refForCopy, RefItem const* startingRef, bool copyReferences, ECSchemaElementType refItemType, RefItem *(ECSchema::*getItemP)(Utf8CP), ECObjectsStatus(ECSchema::*copyItem)(RefItem *&, const RefItem &, bool, Utf8CP))
+    template<typename Item, typename RefItem>
+    ECObjectsStatus ECSchema::GetOrCopyReferencedItemForCopy(const Item & itemWithRef, RefItem *& refForCopy, RefItem const* startingRef, bool copyReferences, ECSchemaElementType refItemType, RefItem *(ECSchema::*getItemP)(Utf8CP), ECObjectsStatus(ECSchema::*copyItem)(RefItem *&, const RefItem &, bool, Utf8CP))
```
rschili, line 1570: indentation is wrong here

rschili: also you did not update the method signature like in the header

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -352,6 +352,19 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
         {
         DumpSchemasToFile(result.GetResults(), dumpLocation.c_str(), "Result");
         }
+    
+    if(false == options.GetSkipValidation())
+        {
+        for(const auto& schema : result.GetResults())
+            {
+            if(!const_cast<ECSchemaP>(schema)->Validate(true))
```
rschili, line 360: you can use m_modifiedSchemas from results here then the const cast is not needed.

iModelCore/ecobjects/src/SchemaMerger.cpp
```diff
@@ -352,6 +352,19 @@ BentleyStatus SchemaMerger::MergeSchemas(SchemaMergeResult& result, bvector<ECSc
...
         DumpSchemasToFile(result.GetResults(), dumpLocation.c_str(), "Result");
         }
+    
+    if(false == options.GetSkipValidation())
+        {
+        for(const auto& schema : result.GetResults())
+            {
+            if(!const_cast<ECSchemaP>(schema)->Validate(true))
+                {
+                result.Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECSchema, ECIssueId::EC_0010,
```
rschili, line 362: You need to generate a unique IssueId for this. The one used is already in use by ECSchemaConverter

iModelCore/ecobjects/test/NonPublished/SchemaMergerTests.cpp
```diff
@@ -4493,6 +4493,284 @@ TEST_F(SchemaMergerTests, MergeRelationshipConstraints)
...
+
+    ECSchemaReadContextPtr rightContext = InitializeReadContextWithAllSchemas({schemaXml2}, nullptr, true);
+    bvector<ECN::ECSchemaCP> rightSchemas = rightContext->GetCache().GetSchemas();
+
+    auto schemaMergeOptions = SchemaMergeOptions();
+    SchemaMergeResult result1;
+    EXPECT_EQ(BentleyStatus::ERROR, SchemaMerger::MergeSchemas(result1, leftSchemas, rightSchemas, schemaMergeOptions));
+    schemaMergeOptions.SetSkipValidation(true);
+    SchemaMergeResult result2;
+    EXPECT_EQ(BentleyStatus::SUCCESS, SchemaMerger::MergeSchemas(result2, leftSchemas, rightSchemas, schemaMergeOptions));
```
rschili, line 4768: Like you already mentioned, use an expected result schema to make sure this works as we expect

iModelCore/ecobjects/src/SchemaComparer.cpp
```diff
@@ -946,7 +946,29 @@ BentleyStatus SchemaComparer::CompareRelationshipConstraint(RelationshipConstrai
     change.RoleLabel().Set(oldConstraint.RoleLabel(), newConstraint.RoleLabel());
     change.IsPolymorphic().Set(oldConstraint.IsPolymorphic(), newConstraint.IsPolymorphic());
     change.Multiplicity().Set(oldConstraint.Multiplicity(), newConstraint.Multiplicity());
-    change.AbstractConstraint().Set(oldConstraint.AbstractConstraint(), newConstraint.AbstractConstraint());
+
+    auto abstractConstraintCheck = [&](RelationshipConstraintProxy const& constraint) -> bool
```
rschili, line 950: This is not a great place to put the fix.
Ideally, this method should compare two values and not imply complicated rules.
The root of the problem is that GetAbstractConstraint() hides the real value from us, and you are using implementation detail knowledge here to determine the raw m_abstractConstraint value.
Better expose the raw value from ECRelationshipConstraint. There are several ways to achieve this. Some examples:
- Make SchemaComparer a friend of ECRelationshipConstraint so you can access m_abstractConstraint
- Add a boolean parameter to GetAbstractConstraint which indicates if you prefer the raw value
- Add a separate GetRawAbstractConstraint method alongside the existing one
- Change GetAbstractConstraint to not do this strange behavior and just return m_abstractConstraint no matter what. But that's a break to our public API. It may be fine, but there is the danger of breaking things if we do that.

rschili (approved):

RohitPtnkr1996 (approved):

PR #1012, opened 2025-02-19
ECSqlReader can see uncommitted changes on primary connection.

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7740

Editing application need to read changes as present on primary connection. `ECSqlReader` uses concurrent queries to read data which uses separate connection which does not see current changes on primary db.

This can be solved with `usePrimaryConn` option but we lose the concurrency in doing so.

An alternative method is to use `PRAGMA read_uncommitted=1` which only works if shared cache is on. So, this PR enables read_uncommitted from concurrent query and enables shared cache by default.


khanaffan:
`PRAGMA read_uncommitted=1` is not really a good option as it see uncommitted changes and require shared page cache which is deprecated by sqlite. We settled with using `usePrimaryCon` for design application.

PR #1019, opened 2025-02-26
Updating zlib to version 1.3.1

naveedkhan8067:
Updating zlib to version 1.3.1 from https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz


iModelCore/libsrc/compress/zlib/contrib/minizip/zip.c
```diff
@@ -1936,15 +1891,14 @@ extern int ZEXPORT zipClose (zipFile file, const char* global_comment)
...
 
     return err;
 }
 
-extern int ZEXPORT zipRemoveExtraInfoBlock (char* pData, int* dataLen, short sHeader)
-{
+extern int ZEXPORT zipRemoveExtraInfoBlock(char* pData, int* dataLen, short sHeader) {
   char* p = pData;
   int size = 0;
   char* pNewHeader;
```
naveedkhan8067, line 1904: Adding a comment here as it seems like on expended lines they are not possible.

```
extern int ZEXPORT zipRemoveExtraInfoBlock(char* pData, int* dataLen, short sHeader) {
  char* p = pData;
  int size = 0;
  char* pNewHeader;
  char* pTmp;
  short header;
  short dataSize;

  int retVal = ZIP_OK;

  if(pData == NULL || *dataLen < 4)
    return ZIP_PARAMERROR;
```

The if condition mentioned in the code block above ([line#1911](https://github.com/iTwin/imodel-native/blob/f2b18ce199448213fe9b3ba426742eec13616d3f/iModelCore/libsrc/compress/zlib/contrib/minizip/zip.c#L1911)) was a bit different, when I compared the current code (main branch zlib version 1.2.13) vs zlib-1.2.13 downloaded) to track any Bentley specific changes. The difference is as follows:

Main branch
```
if(pData == NULL || *dataLen < 4)
  return ZIP_PARAMERROR;
```
Downloaded zlib-1.2.13.tar.gz
```
if(pData == NULL || dataLen == NULL || *dataLen < 4)
  return ZIP_PARAMERROR;
```

I don't know the reason why they are different.

While updating to the newer version 1.3.1, I have noticed that it has `if(pData == NULL || dataLen == NULL || *dataLen < 4)`. At the moment while merging, I accepted the current change not the incoming one but I think we should not be modifying the code.

iModelCore/libsrc/compress/zlib/contrib/minizip/unzip.c
```diff
@@ -1586,7 +1456,7 @@ extern int ZEXPORT unzOpenCurrentFile3 (unzFile file, int* method,
         pfile_in_zip_read_info->stream_initialised=Z_DEFLATED;
       else
       {
-        TRYFREE(pfile_in_zip_read_info);
+        free(pfile_in_zip_read_info);
         return err;
       }
```
naveedkhan8067, line 1461: Similarly, I found difference in two else blocks, when I compared the current code (main branch zlib version 1.2.13) vs zlib-1.2.13 downloaded) to track any Bentley specific changes. The main branch was missing with `TRYFREE(pfile_in_zip_read_info->read_buffer);` statement in the two else blocks mentioned below:

Main branch
```
err=BZ2_bzDecompressInit(&pfile_in_zip_read_info->bstream, 0, 0);
      if (err == Z_OK)
        pfile_in_zip_read_info->stream_initialised=Z_BZIP2ED;
      else
      {
        TRYFREE(pfile_in_zip_read_info);
        return err;
      }
#else
      pfile_in_zip_read_info->raw=1;
#endif
    }
    else if ((s->cur_file_info.compression_method==Z_DEFLATED) && (!raw))
    {
      pfile_in_zip_read_info->stream.zalloc = (alloc_func)0;
      pfile_in_zip_read_info->stream.zfree = (free_func)0;
      pfile_in_zip_read_info->stream.opaque = (voidpf)0;
      pfile_in_zip_read_info->stream.next_in = 0;
      pfile_in_zip_read_info->stream.avail_in = 0;

      err=inflateInit2(&pfile_in_zip_read_info->stream, -MAX_WBITS);
      if (err == Z_OK)
        pfile_in_zip_read_info->stream_initialised=Z_DEFLATED;
      else
      {
        TRYFREE(pfile_in_zip_read_info);
        return err;
      }
```
Downloaded zlib-1.2.13.tar.gz
```
      err=BZ2_bzDecompressInit(&pfile_in_zip_read_info->bstream, 0, 0);
      if (err == Z_OK)
        pfile_in_zip_read_info->stream_initialised=Z_BZIP2ED;
      else
      {
        TRYFREE(pfile_in_zip_read_info->read_buffer);
        TRYFREE(pfile_in_zip_read_info);
        return err;
      }
#else
      pfile_in_zip_read_info->raw=1;
#endif
    }
    else if ((s->cur_file_info.compression_method==Z_DEFLATED) && (!raw))
    {
      pfile_in_zip_read_info->stream.zalloc = (alloc_func)0;
      pfile_in_zip_read_info->stream.zfree = (free_func)0;
      pfile_in_zip_read_info->stream.opaque = (voidpf)0;
      pfile_in_zip_read_info->stream.next_in = 0;
      pfile_in_zip_read_info->stream.avail_in = 0;

      err=inflateInit2(&pfile_in_zip_read_info->stream, -MAX_WBITS);
      if (err == Z_OK)
        pfile_in_zip_read_info->stream_initialised=Z_DEFLATED;
      else
      {
        TRYFREE(pfile_in_zip_read_info->read_buffer);
        TRYFREE(pfile_in_zip_read_info);
        return err;
      }
```

I am not sure why.

While updating to the newer version 1.3.1, I have noticed that it has that `free(pfile_in_zip_read_info->read_buffer);` statement. At the moment while merging, I accepted the current change not the incoming one but I think we should not be modifying the code.

iModelCore/libsrc/compress/zlib/contrib/minizip/zip.c
chuckkir: I don't have any explanation for the minor differences. My gut says go with whatever is in the newest versions. It's possible that there was a patch where they didn't increment the version number or something like that, or it's possible someone here tweaked it for some unknown reason.

Probably best to ask @khanaffan too because he has made all the recent changes to this lib.

naveedkhan8067: Agreed!

khanaffan (approved):

iModelCore/libsrc/compress/zlib/contrib/minizip/unzip.c
khanaffan: As long as I understand we only use LZMA & Snappy, I think same is used by PNG (LZMA). I am not sure which other library uses what part of compress but if we can just trim it down to essential it be great. 

I did not find any reference to  minizip on imodel-native.

PR #1022, opened 2025-02-27
Schema Merger returns actual status instead BentleyStatus

Abhijeet-Bentley:
Issue Link: https://github.com/iTwin/imodel-native/issues/997

This is the second part of the work item.

To get more detailed errors, all methods from SchemaMerger now, have a ECObjectsStatus return type
instead of just BentleyStatus.


RohitPtnkr1996 (approved):

rschili (changes requested):
I don't like how this pretty much returns ECObjectStatus::Success and ECObjectsStatus::Error now instead of BentleyStatus. The only other two codes I saw in the tests are NamedItemAlreadyExists (once) and RelationshipConstraintsNotCompatible (once). I don't think this level of detail is what Carole had in mind when she requested the change.
@c-macdonald please have a brief look. With this level of detail, the change seems like an improvement to me, but only a miniature one. I believe for this to be working right we should not just return ::Error and instead find a correct code to use for each spot.

rschili (approved):

pmconne:
> I believe for this to be working right we should not just return ::Error and instead find a correct code to use for each spot.

@rschili did you change your mind?

rschili:
> @rschili did you change your mind?

@pmconne yeah, sort of. Colin pointed out that even with that, this is a tiny improvement and we can later iterate on it to return more concrete errors. My worry was that if the errors are now useless 95% of the time and only useful 5%, consumers will be educated to just consider them useless.

Anyways, it's a minor concern so I'm fine with pushing the tiny improvement for now.

pmconne (approved):

PR #1023, opened 2025-02-27
Revert a recent change to constraint and fix in a different place

rschili:
A recent change made in https://github.com/iTwin/imodel-native/pull/1010 caused side effects. Reverting the change and fixing it in a different spot now.
Presentation is using the method which had been modified to identify some candidate classes.


ColinKerr (commented):
This change needs an associated test to codify the correct behavior

PR #1024, opened 2025-02-27
Implement LineStyleElement::_CloneForImport()

DanRod1999:
Brought up in this [issue](https://github.com/iTwin/itwinjs-core/discussions/7642), it was found that when importing line style elements we currently do not copy over style data in the Be Prop table from the source db to the target. Note: I used [this issue ](https://github.com/iTwin/imodel-transformer/issues/231)to update my work.

To fix this I implemented CloneForImport() for line style elements and used it to call RemapLineStyleID. This function is called during GeometryStreamIO::Import however, at this point is is too late, so it skips the portion of code that imports the necessary data from the line style element.

The structure for this function was loosely copied from the Clone for import functions for Text Annotation elements.

After adding this function I also ran into problems with some of the current implementations of line style import functions not having correctly remapping, or not updating the necessary values on the element based on the remapping.

I tested this in the transformer repo using these two tests: [simple line style](https://github.com/iTwin/imodel-transformer/blob/dan/line-style-v5/packages/transformer/src/test/standalone/IModelTransformer.test.ts#L5150), [compound line style](https://github.com/iTwin/imodel-transformer/blob/dan/line-style-v5/packages/transformer/src/test/standalone/IModelTransformer.test.ts#L5280). I verified visually that the line styles appeared correctly using a viewer, and also checked the the elements within the be_prop and bis_element tables in the bim files. If these tests are worth keeping in the transformer repo I will modify them to run proper assertions on the cloned elements.


iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -562,10 +570,10 @@ DgnStyleId DgnImportContext::_RemapLineStyleId(DgnStyleId sourceId)
         return dest;
 
     
-#if defined(NEEDSWORK_LINESTYLES) //  importers are not tested so don't risk passing along bad data.
+// #if defined(NEEDSWORK_LINESTYLES) //  importers are not tested so don't risk passing along bad data.
```
aruniverse, line 573: should this be commented out? if so lets remove it

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -631,9 +637,37 @@ DgnStyleId LineStyleElement::ImportLineStyle(DgnStyleId srcStyleId, DgnImportCon
...
+    DgnDbStatus _status;
+    if (nullptr == status)
+        status = &_status;
+
+
+    DgnElementPtr newElem = T_Super::_CloneForImport(status, destModel, context);
+    if (!newElem.IsValid() || (DgnDbStatus::Success != *status))
+        return newElem;
+
+    DgnStyleId destStyleId = context.RemapLineStyleId(DgnStyleId(this->m_elementId.GetValue()));
```
bbastings, line 658: This looks wrong to me? Should be getting data from "this", copying it to newElem and using the DgnImportContext to remap any ids it contains if context.IsBetweenDbs is true. Then you just return newElem. Don't think that call to Update should be here either...newElem will be inserted the first time this is called...and then never called again as a geometric element that references this style should find an existing id from the DgnImportContext.

bbastings:
Seems like something is awry here. If this code appears to be working then I surmise that the line style definition element is being erroneously copied without going through _CloneForImport and then later getting fixed up by a call to _CloneForImport because it's calling Update. Likely getting a 2nd definition inserted from the newElem being returned and just no one references it...

bbastings (commented):
Seems like something is awry here. If this code appears to be working then I surmise that the line style definition element is being erroneously copied without going through _CloneForImport and then later getting fixed up by a call to _CloneForImport because it's calling Update. Likely getting a 2nd definition inserted from the newElem being returned and just no one references it.

Look at DgnElement::Import you see what I mean. For the LineStyleElement you can probably just implement _RemapIds to handle the remap of the be_props style id...it doesn't have anything complicated to do that requires overriding _CloneForImport.

```
DgnElementPtr DgnElement::_CloneForImport(DgnDbStatus* inStat, DgnModelR destModel, DgnImportContext& importer) const
    {
    DgnDbStatus ALLOW_NULL_OUTPUT(stat, inStat);

    DgnElement::CreateParams params = GetCreateParamsForImport(destModel, importer);
    params.m_modelId = destModel.GetModelId();

    if (!params.IsValid())
        {
        BeNapi::ThrowJsException(
            m_dgndb.GetJsIModelDb()->Env(),
            params.m_classId.IsValid() ? "invalid create params" : "attempt to clone with unknown class",
            (int) (params.m_classId.IsValid() ? DgnDbStatus::BadRequest : DgnDbStatus::WrongClass)
        );
        }

    DgnElementPtr cloneElem = GetElementHandler().Create(params);

    if (!cloneElem.IsValid())
        {
        stat = DgnDbStatus::BadRequest;
        return nullptr;
        }

    cloneElem->CopyForCloneFrom(*this);

    if (importer.IsBetweenDbs())
        {
        cloneElem->_RemapIds(importer);
        cloneElem->_AdjustPlacementForImport(importer);
        }
    else if (GetParentId().IsValid())
        {
        cloneElem->SetParentId(importer.FindElementId(GetParentId()), GetParentRelClassId());
        }

    stat = DgnDbStatus::Success;
    return cloneElem;
    }
```

DanRod1999:
> Seems like something is awry here. If this code appears to be working then I surmise that the line style definition element is being erroneously copied without going through _CloneForImport and then later getting fixed up by a call to _CloneForImport because it's calling Update, which is very wrong

Ah I see. I did not make that connection before. `Update` did feel wrong, because I noticed other clone for import funcs didn't seem to need it. I did not realize until now that `newElem` was probably due to the definition element not being cloned correctly.

However, I am still a little confused because I thought the definition element was inserted when RemapLineStyleId inserted the linestyle element
LsDb.cpp: Ln640
`BentleyStatus result = importer.GetDestinationDb().LineStyles().Insert(dstStyleId, name.c_str(), compId, LsDefinition::GetAttributes(jsonObj), LsDefinition::GetUnitDef(jsonObj));`

Wouldn't this create 2 definition elements? One from the clone for import and a new one from this linestyle insert?

bbastings:
> Wouldn't this create 2 definition elements? One from the clone for import and a new one from this linestyle insert?

No, look at the implementation of ImportElement and _CloneForImport on the base element class.

Whatever causes the element to be clones first put the entry for the new id into the remap table so any other reference to it just gets that id.

DanRod1999:
rewrote _CloneForImport() using pieces of `ImportLineStyle()` and `LineStyles().Insert(...)`. This removes RemapLineStyleId() because that function prematurely inserts the line style element that is being cloned. I essentially took everything from ImportLineStyle to remap the component ID of the new element, and then from `Insert()` Initialized the new json object, and assigned all the data from the json object to the `newElem` that I returned. Then I omitted the actual insert of the element, as that will take place after the clone occurs.

Should be good to review again

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -631,9 +638,69 @@ DgnStyleId LineStyleElement::ImportLineStyle(DgnStyleId srcStyleId, DgnImportCon
...
+        }
+
+    DefinitionModelPtr model = m_dgndb.Models().Get<DefinitionModel>(DgnModel::DictionaryId());
+
+    LsDefinition::InitializeJsonObject(jsonObj, compId, LsDefinition::GetAttributes(jsonObj), LsDefinition::GetUnitDef(jsonObj));
+    Utf8String data = Json::FastWriter::ToString(jsonObj);
+
+    LineStyleElementPtr lsElement = dynamic_cast<LineStyleElement*>(newElem.get());
+    lsElement->SetName(name.c_str());
+    lsElement->SetDescription(nullptr);
```
bbastings, line 700: Name and description aren't changing. If someone added a description somehow, should leave it be...

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -631,9 +638,59 @@ DgnStyleId LineStyleElement::ImportLineStyle(DgnStyleId srcStyleId, DgnImportCon
...
+    if (nullptr == status)
+        status = &_status;
+
+
+    DgnElementPtr newElem = T_Super::_CloneForImport(status, destModel, context);
+    if (!newElem.IsValid() || (DgnDbStatus::Success != *status))
+        return newElem;
+
+    if(!context.IsBetweenDbs())
+        return newElem;
```
DanRod1999, line 660: can this !context.IsBetweenDbs() be added to the above conditional, or better practice to keep in its on condition?

DanRod1999:
I've noticed that elements with an unspecified unit def are supposed to default to 1. The elements I am using in my source db do not have a unit def field in their data in the bis_defintionElement table, but the cloned elements do add a unit def of 0. I do not see any issues visually with this, but it does seem wrong? Should unit defs of 0 be converted to 1 so they are more accurately entered, or is the 0 intentional?
src:
  data: '{"compId":0,"compType":2,"flags":4}',
clone:
  data: '{"compId":8,"compType":2,"flags":4,"unitDef":0.0}',
  
Note: Non 0 Unit defs correctly copy unit defs

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
bbastings: Doesn't matter, but I'd leave it separate, easier for someone to spot if they find this as an example for doing another element.

bbastings:
> Should unit defs of 0 be converted to 1

Code has checks for 0 so it doesn't matter (below snippet from LineStyleSymb::Init)

```
    // now adjust scale for units in linestyle definition
    double unitDef = nameRec->GetUnitsDefinition();
    if (unitDef == 0)
        unitDef = 1.0;
```

DanRod1999:
> Code has checks for 0 so it doesn't matter (below snippet from LineStyleSymb::Init)

Got it thanks. Didn't know if it was a good thing to keep in the tables, so just wanted to check.

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -631,9 +638,59 @@ DgnStyleId LineStyleElement::ImportLineStyle(DgnStyleId srcStyleId, DgnImportCon
...
+
+    DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
+    if (dstStyleId.IsValid())
+        {
+        //  *** TBD: Check if the line style definitions match. If not, rename and remap
+        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
+        return const_cast<DgnElement*>(context.GetDestinationDb().Elements().GetElement(dstStyleId).get());
```
bbastings, line 667: Can't return an element, remember caller is going to insert the element that is returned...

Calling context.AddLineStyleId is sufficient for geometric elements to remap the style, but should probably call context.AddElementId with this same pair of ids too so that generic element id remap lookup finds it too.

Then you need to return an invalid element to stop DgnElement::Import from doing the insert...

bbastings (changes requested):

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -631,9 +638,59 @@ DgnStyleId LineStyleElement::ImportLineStyle(DgnStyleId srcStyleId, DgnImportCon
...
+    LsComponentId compId = LsDefinition::GetComponentId (jsonObj);
+    compId = LsComponent::Import(compId, context);
+    if (!compId.IsValid())
+        {
+        //  *** TBD: Check if the line style definitions match. If not, rename and remap
+        BeAssert(false && "unable to import component for line style");
+        return newElem;
+        }
+
+    DefinitionModelPtr model = m_dgndb.Models().Get<DefinitionModel>(DgnModel::DictionaryId());
```
bbastings, line 685: model isn't used...

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
DanRod1999: Is returning `DgnElementPtr()` sufficient or should it just be nullptr? Also based on `DgnElements::InsertElement()` I noticed
`// don't allow elements that already have an id unless the "preassignedId" flag is set`
`if (element.m_elementId.IsValid() && !element.m_flags.m_preassignedId)`
I assume its still safer to just return an empty element in _CloneForImport?

bbastings: nullptr is fine. I just think it's cleaner to bail as soon as possible and not rely on the id check on insert...

bbastings (approved):

aruniverse:
@merigfyio backport release/4.10.x release/4.11.x

aruniverse:

PR #1031, opened 2025-03-06
Fix: Ensure `ECSqlReader` returns correct metadata for instance properties

akshay05n:
This PR fixes an issue where `ECSqlReader` does not return the correct metadata for instance properties when selecting dynamic columns using ECSQL queries like:

``` sql
SELECT $->prop FROM Bis.Element
```

### DIFF
--- a/iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
+++ b/iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
@@ -1769,6 +1769,88 @@ TEST_F(InstanceReaderFixture, dynamic_meta_data) {
     }
 }
 
+TEST_F(InstanceReaderFixture, ResetDynamicMetadata) {
+    ASSERT_EQ(SUCCESS, SetupECDb("ResetDynamicMetadata.ecdb", SchemaItem(
+        R"xml(<ECSchema schemaName='TestSchema' alias='ts' version='1.0.0' xmlns='http://www.bentley.com/schemas/Bentley.ECXML.3.2'>
+            <ECEntityClass typeName='A'>
+            </ECEntityClass>
+            <ECEntityClass typeName='B'>
+                <BaseClass>A</BaseClass>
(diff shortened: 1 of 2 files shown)


iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
```diff
@@ -1769,6 +1769,87 @@ TEST_F(InstanceReaderFixture, dynamic_meta_data) {
...
+            select
+                $->prop
+            from ts.A
+        )x";
+
+        ECSqlStatement stmt;
+        ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, sql));
+        assertDefault(stmt, 0 , "$->prop", "__x0024____x002D____x003E__prop");
+        while(stmt.Step() == BE_SQLITE_ROW) {
+            auto ecInstanceId = stmt.GetValueInt(0);
```
khanaffan, line 1841: `stmt.GetValueInt(0) is $->prop` not `ECInstanceId` which is not selected. The loop runs but I bet none of `assertDynamic()` ever executed. `ecInstanceId` be likely `0`. 

You can just do 

```
ASSERT_EQ(BE_SQLITE_ROW, stmt.Step())
assertDynamic(stmt, 0 , "prop", "prop", "", "B", PrimitiveType::PRIMITIVETYPE_Integer);
ASSERT_EQ(BE_SQLITE_ROW, stmt.Step())
assertDynamic(stmt, 0 , "prop", "prop", "", "C", PrimitiveType::PRIMITIVETYPE_Double);
```

khanaffan (changes requested):

iModelCore/ECDb/Tests/NonPublished/InstanceReaderTest.cpp
akshay05n: I have modified the test according to your comments. I missed this part when I removed `ecInstanceId` from the select statement. Thanks for pointing it out.

khanaffan (approved):

PR #1051, opened 2025-03-20
Fix for unhandled crash in ECSqlFieldFactory::CreateTopLevelColumnInfo()

Abhijeet-Bentley:
Issue link: https://github.com/iTwin/itwinjs-backlog/issues/1308

Test case reveals a missing null check inside ECSqlFieldFactory::CreateTopLevelColumnInfo() for ecProperty pointer.
Upon further investigation the error seems to be originating from TryGetPropertyMap(). The propertyMap returned by said function, when resolved consists of lesser members inside resolvePath than path. Hence, a check is added to see if the resolvePath's size is lesser than that of path's, if true we throw ERROR.


iModelCore/ECDb/ECDb/ECSql/ECSqlFieldFactory.cpp
```diff
@@ -299,11 +299,15 @@ ECSqlColumnInfo ECSqlFieldFactory::CreateColumnInfoForProperty(ECSqlPrepareConte
...
 ECSqlColumnInfo ECSqlFieldFactory::CreateTopLevelColumnInfo(IssueDataSource const& issues, bool isSystemProperty, bool isGeneratedProperty, ECSqlPropertyPath const& propertyPath, ECSqlColumnInfo::RootClass const& rootClass, ECPropertyCP originalProperty, bool isDynamic)
     {
-    BeAssert(propertyPath.Size() > 0);
-    ECPropertyCP ecProperty = propertyPath.GetLeafEntry().GetProperty();
-    BeAssert(ecProperty != nullptr);
     DateTime::Info dateTimeInfo;
     ECStructClassCP structType = nullptr;
+    BeAssert(propertyPath.Size() > 0);
+    ECPropertyCP ecProperty = propertyPath.GetLeafEntry().GetProperty();
+    //BeAssert(ecProperty != nullptr);
```
pmconne, line 306:
```suggestion
```

pmconne:
@khanaffan can you review? From tracing through the callers it seems nobody checks if the returned ECSqlColumnInfo is valid, they just assume it is. For example, later on this will crash for same reason (assumes property pointer is non-null):
```
void PrimitiveECSqlField::UpdateDateTimeMetaData() {
    auto& columnInfo = GetColumnInfo();
    if (columnInfo.GetDataType().GetPrimitiveType() == PRIMITIVETYPE_DateTime) {
        ECPropertyCP property = columnInfo.GetProperty();
        BeAssert(property != nullptr && "ColumnInfo::GetProperty can return null. Please double-check");
        if (CoreCustomAttributeHelper::GetDateTimeInfo(m_datetimeMetadata, *property) != ECObjectsStatus::Success)
```

rschili:
Mentioning https://github.com/iTwin/itwinjs-backlog/issues/1079 here.
We do want to consider removing property from columnInfos altogether soon. But discussions are needed.

iModelCore/ECDb/ECDb/ECSql/ECSqlFieldFactory.cpp
```diff
@@ -299,11 +299,15 @@ ECSqlColumnInfo ECSqlFieldFactory::CreateColumnInfoForProperty(ECSqlPrepareConte
...
+    if (ecProperty == nullptr)
+        {
+        return ECSqlColumnInfo(ECTypeDescriptor(), dateTimeInfo, structType, ecProperty, originalProperty, isSystemProperty, isGeneratedProperty, propertyPath, rootClass, isDynamic);
```
khanaffan, line 309: So this is not really a fix. The type of descriptor is empty. This should never be empty of any property. For property p2d.x it should be primitive and double.

Without this information we will not be able to serialize this into json.

iModelCore/ECDb/Tests/NonPublished/ECSqlStatementTests.cpp
```diff
@@ -259,6 +259,25 @@ TEST_F(ECSqlStatementTestFixture, MultilineStringLiteralOrName) {
...
+                        <ECEntityClass typeName="TestElement" modifier="None">
+                            <BaseClass>IPrimitive</BaseClass>
+                        </ECEntityClass>
+                        <ECEntityClass typeName="IPrimitive" modifier="Abstract">
+                            <ECProperty propertyName="p2d" typeName="point2d"/>
+                        </ECEntityClass>
+                </ECSchema>)xml")));
+
+    ECSqlStatement stmt2;
+    ASSERT_EQ(ECSqlStatus::Success, stmt2.Prepare(m_ecdb, "SELECT p2d.X FROM (SELECT * FROM (SELECT * FROM aps.TestElement LIMIT 1))"));
```
khanaffan, line 276: You have changed ColumnInfo but have not tested it here to see what it is? Add Assertion to poke stmt2.GetValue(0).GetColumnInfo() and verify it content. It should match p2d.x in a regular schema.

iModelCore/ECDb/ECDb/ECSql/ECSqlFieldFactory.cpp
khanaffan: Also as a rule top level property must have ECProperty set not matter what. We create a statement schema just for this purpose where any expr that is not a direct EC Property we just create a runtime property and add to the statement schema.

khanaffan: Also add test on itwin.js side with same schema and query. I am confident with your fix it will still crash but at a different place. ConcurrentQuery uses ECSqlRowAdaptor to convert rows to json. It also assumes top level properties will not be null. As its either ECProperty from class or a generated ECProperty but never null.

tcobbs-bentley:
@Abhijeet-Bentley Is it ok for me to backport this to iTwin 4.x?

Abhijeet-Bentley:
@tcobbs-bentley Sure, it is not a long-term fix but, failing and returning error is always better than crashing.

tcobbs-bentley:

Issue #912, opened 2024-11-07
ECSQL -> SQL conversion adds unnecessary clauses, causing poor query performance


grigasp:
I have the following ECSQL query:
```sql
SELECT count(1)
FROM bis.SpatialCategory c
WHERE EXISTS (
  SELECT 1
  FROM bis.GeometricElement3d element
  WHERE
    element.Model.Id = ?
    AND element.Category.Id = c.ECInstanceId
    AND element.Parent IS NULL
)
```

For some specific Models in my test iModels the query takes a lot of time to execute, e.g.:

| iModel | Model.Id | Time it takes to execute query |
| --- | --- | --- |
| A | 0x1700000001c7 | 4.97 s. |
| B | 0x240000000018 | 13.67 s. |
| C | 0x20000000030  | 12.43 s. |
| D | 0x20000001971  | 16.73 s.|

When looking at why it's slow, I found that it includes a calculated `ParentRelECClassId` column into the WHERE clause, which makes the query slow:

```sql
SELECT COUNT(1)
FROM (
    SELECT [ElementId] ECInstanceId,
      [ECClassId]
    FROM [main].[bis_DefinitionElement]
    WHERE [bis_DefinitionElement].ECClassId = 169
  ) [c]
WHERE EXISTS(
    SELECT 1
    FROM (
        SELECT [bis_GeometricElement3d].[ElementId] ECInstanceId,
          [bis_GeometricElement3d].[ECClassId],
          [bis_Element].[ModelId],
          [bis_Element].[ParentId],
          (
            CASE
              WHEN [bis_Element].[ParentId] IS NULL THEN NULL
              ELSE [bis_Element].[ParentRelECClassId]
            END
          ) [ParentRelECClassId],
          [bis_GeometricElement3d].[CategoryId]
        FROM [main].[bis_GeometricElement3d]
          INNER JOIN [main].[bis_Element] ON [bis_GeometricElement3d].[ElementId] = [bis_Element].[Id]
      ) [element]
    WHERE [element].[ModelId] = ?
      AND [element].[CategoryId] = [c].[ECInstanceId]
      AND (
        [element].[ParentId] IS NULL
        AND [element].[ParentRelECClassId] IS NULL
      )
  )
```

As a workaround, I found that I can add `.Id` to the parent nullity check (`... AND element.Parent.Id IS NULL`), in which case the unnecessary clause in the SQL is not added:

```sql
SELECT COUNT(1)
FROM (
[...]

PR #1073, opened 2025-04-14
Fix: Skip redundant [Element].[ParentRelECClassId] = NULL expression and avoid unnecessary SQL parentheses.

rushikesh-bentley:
Issue Link: https://github.com/iTwin/imodel-native/issues/912 

- Identified that [Element].[ParentRelECClassId] = NULL expressions were leading to redundant SQL.
- Added a conditional check to skip such expressions in the loop building SQL snippets.
- Introduced a fallback condition TRUE to avoid generating empty SQL when all snippets are skipped.
- Moved isFirstSnippet logic outside the loop to determine if any expressions were added.
- Updated logic to avoid adding parentheses when all snippets are skipped, preventing unnecessary wrapping.


RohitPtnkr1996:
Superseded by https://github.com/iTwin/imodel-native/pull/1380

PR #1076, opened 2025-04-17
Fixes for Clang/LLVM 20 detected problems

rschili:
I fixed the things that were easy, and left some TODO comments for the things that require additional work.

I created a new story to work off the remaining items: https://github.com/iTwin/itwinjs-backlog/issues/1473

PR #1077, opened 2025-04-17
Geomlibs cleanup for clang/LLVM 20

dassaf4:
Follows up https://github.com/iTwin/imodel-native/pull/1076
Addresses the geomlibs portion of https://github.com/iTwin/itwinjs-backlog/issues/1473

### DIFF
--- a/iModelCore/GeomLibs/PublicAPI/Geom/GeomApi.h
+++ b/iModelCore/GeomLibs/PublicAPI/Geom/GeomApi.h
@@ -398,19 +398,6 @@ struct  GeometryValidator : public RefCountedBase
 
 };
 
-//=======================================================================================
-//! Inlinable template for initial zeroing an object of size T.
-//!@bsiclass
-//=======================================================================================
-template<typename T> struct ZeroInit
-{
-    ZeroInit() 
-        {
-        static_assert(std::is_trivially_copyable<T>::value, 
-                      "ZeroInit can only be used with trivially copyable types.");
-        memset(this, 0, sizeof(T));
-        }
-};
 //! POD struct for pair of size_t values
 struct SizeSize
 {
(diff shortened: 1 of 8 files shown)


saeeedtorabi (approved):

rschili (approved):
Cool, that was fast, nice to see a follow up.
I'll update https://github.com/iTwin/itwinjs-backlog/issues/1473 and remove those things from the list of things to do.

May I ask why the choice to use explicit constructor initialization over field initializers? Lately I tend towards preferring the field initializers over ctor because it's much easier to miss when adding or changing fields.

I wonder if at some point we should replace that custom BSIBaseGeom:Malloc/Free with standard library smart pointers. Is anybody still utilizing the feature to set custom allocators instead of malloc?

dassaf4:
> May I ask why the choice to use explicit constructor initialization over field initializers? Lately I tend towards preferring the field initializers over ctor because it's much easier to miss when adding or changing fields.

Sure! And thanks for taking the first crack. I certainly learned something today. In the `SolidPrimitive` classes, there already were constructors present, so it's more efficient to let them do the initialization. (I've stepped through this kind of thing before. They do get initialized twice if both types of initialization are present!) But to your point, one of the constructors did miss a field, soo.... Anyway, I could've gone either way in the classes that had no user-defined constructors; I went with constructors because it felt cleaner, and a developer who adds/changes a field should be checking non-default constructors and destructors regardless of whether or not the other fields have initializers.

> I wonder if at some point we should replace that custom BSIBaseGeom:Malloc/Free with standard library smart pointers. Is anybody still utilizing the feature to set custom allocators instead of malloc?

Yeah, that part hasn't aged well. Geomlibs has some dark dusty corners. The newer parts consistently use Bentley smart pointers via RefCounted.h. I doubt the facility is used. But replacing memory management throughout might introduce more bugs than could be lurking, so this is low priority.

PR #1080, opened 2025-04-22
Clone Element returns Error on linestyle Elements that already exist in Dest Imodel

DanRod1999:
Resolves:
https://github.com/iTwin/imodel-transformer/issues/250

When cloning a linestyle element that already exists currently we remap id's but return null. This throws an error during 
```
Napi::Value CloneElement(NapiInfoCR info)
...
        DgnElementPtr targetElement = sourceElement->CloneForImport(&cloneStatus, *targetModel, *m_importContext);
        if (cloneStatus != DgnDbStatus::Success || !targetElement.IsValid())
            THROW_JS_EXCEPTION("Unable to clone element");
...
```
Instead of Null I returned the dest element, to avoid this error. Since we typically return `newElem` which becomes the element in the dest db I believe this should be equivalent.


bbastings (commented):
_CloneForImport should be returning an element unless there is an error, however:

> When cloning a linestyle element that already exists currently we remap id's but return null.

Why are we trying to clone an element that already exists in the destination? This is certainly wrong...

If you look at ElementImporter:ImportElement you'll see that it doesn't call clone when the element is already found in the destination.

```
DgnElementCPtr ElementImporter::ImportElement(DgnDbStatus* statusOut, DgnModelR destModel, DgnElementCR sourceElement)
    {
    auto destElementId = m_context.FindElementId(sourceElement.GetElementId());
    if (destElementId.IsValid()) // If source element was already copied, just return the existing copy. This happens, for example, when a parent deep-copies its children immediately.
        return m_context.GetDestinationDb().Elements().GetElement(destElementId);

    DgnElementCPtr destElement = sourceElement.Import(statusOut, destModel, m_context);
    if (!destElement.IsValid())
        return nullptr;
...
```

DanRod1999:
Just to Clarify, this if stmt is checking if the StyleID already exists in the destDb, which is the same thing as checking if the lines style element already exists in the dest db? If we shouldn't be cloning elements multiple times, I'm not sure I understand the purpose of this stmt
```
    DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
    if (dstStyleId.IsValid())
        {
        //  *** TBD: Check if the line style definitions match. If not, rename and remap
        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
        DgnElementId elemId = context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
        
        return context.GetDestinationDb().Elements().GetForEdit<LineStyleElement>(elemId);
        }
```
In terms of cloning elements twice, I agree this is not great behavior. @ViliusRuskys looked through the transformer and saw that we check if the element already exists after we clone it, so if avoidable we should fix that problem there. I'm not sure if its problematic, but it does seem wasteful at least.

DanRod1999:
> Why are we trying to clone an element that already exists in the destination? This is certainly wrong...
In terms of cloning elements twice, I agree this is not great behavior. @ViliusRuskys looked through the transformer and saw that we check if the element already exists after we clone it, so if avoidable we should fix that problem there. I'm not sure if its problematic, but it does seem wasteful at least.

However, if an element was updated in the SourceDb then I believe this is an instance where we want to clone the element again. I believe this is a reason why we cannot skip an element if it already exists. Our current tests do not update an existing element, so I have not tested this, but in that case, wouldn't this still return true?
```
    DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
    if (dstStyleId.IsValid())
```

bbastings:
> If we shouldn't be cloning elements multiple times, I'm not sure I understand the purpose of this stmt

A line style is identified by it's name. It's trying to check that a line style with the same name in the destination as in the source has the same properties, i.e. is the same line style.

This shouldn't come up with the transformer as the destination is always created from the same source so there should never be any name collisions.

bbastings:
> However, if an element was updated in the SourceDb then I believe this is an instance where we want to clone the element again.

The transformer is being called in some sort of incremental update mode? It doesn't always create a new destination model from the source? I don't see how this could work. You certainly still don't want to clone an element multiple time because that's an insert when you'd want an update?!?

JulijaRamoskiene:
> > However, if an element was updated in the SourceDb then I believe this is an instance where we want to clone the element again.
> 
> The transformer is being called in some sort of incremental update mode? It doesn't always create a new destination model from the source? I don't see how this could work. You certainly still don't want to clone an element multiple time because that's an insert when you'd want an update?!?

The way imodel-transofrmer was implemented, it currently: 
1. Iterates elements in source. For each processed element cloneElement() is called. 
2.  Then `imodel-transformer` checks if element already exists in target, if element is found properties returned from clone element are updated accordingly, and element is updated in target: 

```
       // `onTransformElement()` calls `cloneElement()`
      const targetProps = this.onTransformElement(sourceElement);
      this.targetDb.elements.updateElement({ ...targetProps, id: targetId });
```

JulijaRamoskiene:
> > Is there something wrong with cloning element if it already exists in target?
> 
> No, but you don't want to return the element in the destination which is what this PR is doing. Element changed in source -> clone source -> update existing destination element, is perfectly fine. Either way, for insert or update I would only expect the line style destination to be cloned once in a given run of the transformer.

To clarify  "once in given run of transformer" =>  In scenario described above there would be 2 runs of transformer and each run would call cloneElement on line style once: 

1. First run:  all elements are processed and exported to target. Lines style is cloned and inserted into target. 
2. Source iModel is modified by user. Line style is updated in source.
3. Second run: transformer  is executed again, it reprocesses changed elements, since it sees that line style was updated, it will attempt to clone linestyle by calling cloneElement() and would then attempt to update it in target with returned props .


@bbastings  Is it correct that  `_CloneForImport` should not return  if line style is found in destination in this `if`:
```
if (dstStyleId.IsValid())
        {
        //  *** TBD: Check if the line style definitions match. If not, rename and remap
        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
        context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
        return nullptr;
        }
```

Instead it should just remap ids and then continue with clonning? Something like: 

```
DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
    if (dstStyleId.IsValid())
        {
        //  *** TBD: Check if the line style definitions match. If not, rename and remap
        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
        context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
        }

    Utf8String  srcData (this->GetData());

    Json::Value  jsonObj (Json::objectValue);
    if (!Json::Reader::Parse(srcData, jsonObj))
        return newElem;
    >....<
```

bbastings:
> There can be name collisions.

In order to properly support this, that TBD is important, would need to implement that comparison to see if the styles with the same name are actually different AND if so, actually rename it. You then need to be able to find that rename somehow so every update doesn't end up looking for the "original" name and creating another renamed definition.

Don't know if transformer is already storing some form of provenance that could preserve this information. Another approach would be using a naming convention for the rename that allows you to find the correct definition to compare, ex. style name + model name or something. Without this you'll not be able to distinguish between a new style with a name collision or a style that changed/needs to be updated, right?

People generally don't modify their line style definitions, so supporting that is a pretty fringe case imo. Without implementing the TBD, you aren't handling name collisions from multiple sources anyway, better just to remove this check entirely as it's certainly not helping in it's current state.

```
    DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
    if (dstStyleId.IsValid())
        {
        //  *** TBD: Check if the line style definitions match. If not, rename and remap
        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
        DgnElementId elemId = context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
        
        return context.GetDestinationDb().Elements().GetForEdit<LineStyleElement>(elemId);
        }
```

JulijaRamoskiene:
> > There can be name collisions.
> 
> In order to properly support this, that TBD is important, would need to implement that comparison to see if the styles with the same name are actually different AND if so, actually rename it. You then need to be able to find that rename somehow so every update doesn't end up looking for the "original" name and creating another renamed definition.
> 
> Don't know if transformer is already storing some form of provenance that could preserve this information. Another approach would be using a naming convention for the rename that allows you to find the correct definition to compare, ex. style name + model name or something. Without this you'll not be able to distinguish between a new style with a name collision or a style that changed/needs to be updated, right?
> 
> People generally don't modify their line style definitions, so supporting that is a pretty fringe case imo. Without implementing the TBD, you aren't handling name collisions from multiple sources anyway, better just to remove this check entirely as it's certainly not helping in it's current state.
> 
> ```
>     DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
>     if (dstStyleId.IsValid())
>         {
>         //  *** TBD: Check if the line style definitions match. If not, rename and remap
>         context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
>         DgnElementId elemId = context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
>         
>         return context.GetDestinationDb().Elements().GetForEdit<LineStyleElement>(elemId);
>         }
> ```
@bbastings 
We resolve name collisions at a different level in our service. before these cloneElement() is even called. i think it should not be responsibility of this method. Also, transformer will call remap element once it finds target element on its own. In this case, we just want to  clone source line style, and then update line style element in target with these cloned properties. I think removing this remapping logic makes most sense. Thanks a lot for your help.

bbastings:
> We resolve name collisions at a different level in our service.

Great, yeah definitely just remove that check it's doing nothing good now. :)

DanRod1999:
Thanks everyone. I'm going to then remove this condition

iModelCore/iModelPlatform/DgnCore/linestyle/LsDb.cpp
```diff
@@ -659,29 +659,28 @@ DgnElementPtr LineStyleElement::_CloneForImport(DgnDbStatus* status, DgnModelR d
...
     DgnStyleId dstStyleId = QueryId(context.GetDestinationDb(), this->GetName().c_str());
     if (dstStyleId.IsValid())
         {
-        //  *** TBD: Check if the line style definitions match. If not, rename and remap
-        context.AddLineStyleId(DgnStyleId(this->GetElementId().GetValue()), dstStyleId);
-        DgnElementId elemId = context.AddElementId(this->GetElementId(), DgnElementId(dstStyleId.GetValue()));
-        
-        return context.GetDestinationDb().Elements().GetForEdit<LineStyleElement>(elemId);
-        }
+        LineStyleElementPtr dstElem = context.GetDestinationDb().Elements().GetForEdit<LineStyleElement>(dstStyleId);
```
bbastings, line 672: Should not be using GetForEdit, since you aren't going to change this element. Should also add a comment to explain why this is necessary.

bbastings (changes requested):

bbastings (approved):

DanRod1999:

PR #1083, opened 2025-04-24
Presentation: Fix property overrides always taking precedence over class-based overrides

grigasp:
Fixes https://github.com/iTwin/itwinjs-core/issues/7983

There were 2 issues:
1. If a property-specific override was found, we didn't even look at class-level overrides. That's wrong, as we should pick the override with the highest priority.
2. The "hide the rest" class-level property display overrides, which are added with lowest possible priority, were overwriting higher priority overrides that could possibly already be there.


grigasp:

grigasp:

saskliutas (approved):

PR #1084, opened 2025-04-24
Fix for: Clearing ECDb cache causes repeated ConcurrentQuery to fail

rschili:
fixes: https://github.com/iTwin/itwinjs-core/issues/7984

The cleared cache did not properly clean ConcurrentQuery's internal caches, so if an identical query was sent again, the statement would be reused, which triggers an internal failsafe to protect against that.

Now we clean concurrent query caches.

@khanaffan please check. There are so many classes and caches in concurrent query, I'm only 80% sure that I hit the right spot and didn't miss anything. My added test passes after fix and fails without.

### DIFF
--- a/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
+++ b/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
@@ -397,6 +397,9 @@ void ConnectionCache::Interrupt(bool reset_conn, bool detach_dbs) {
         std::this_thread::yield();
     }
     if (reset_conn) {
+        if(m_syncConn)
+            m_syncConn->Reset(detach_dbs);
+
         for (auto& it : m_conns) {
             it->Reset(detach_dbs);
         }
--- a/iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
+++ b/iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
@@ -1234,4 +1234,60 @@ TEST_F(ConcurrentQueryFixture, ReaderBindingForIdSetVirtualTable) {
         }
 }
 
+//---------------------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -328,6 +328,15 @@ void ConnectionCache::InterruptIf(std::function<bool(RunnableRequestBase const&)
             conn->InterruptIf(predicate, cancel);
     }
 }
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//---------------------------------------------------------------------------------------
+void ConnectionCache::Clear() {
+    m_conns.clear();
```
khanaffan, line 336: Change the state of this class requires lock on mutex. But that could cause race condition.

May be the simple solution to this is to restart concurrent query on schema import

rschili: I don't think Shutdown does what you suggested it does. It just clears some cache directories on disk, it does not reset or clear connections, so it does not help.
I've put a different fix now. It seems like there already is a reset logic inside Interrupt().
Allow me the criticism, I think this concurrent query architecture is too complicated.
The calltree is

```
ConcurrentQueryMgr::Impl::_OnBeforeClearECDbCache()
-> ConcurrentQueryMgr::Impl::Suspend(ClearCacheOption clearCache, DetachAttachDbs detachDbs)
-> ConnectionCache::Interrupt(bool reset_conn, bool detach_dbs)
-> CachedConnection::Reset(bool detachDbs) // This one even calls m_db.ClearECDbCache() again resursively!!
-> QueryAdaptorCache::Reset()
-> (clears vector of CachedQueryAdaptor)
```

There are so many managers adaptors and caches, reminds me of the C# days :-)

khanaffan (approved):

aruniverse:

PR #1089, opened 2025-04-25
Fix: Issue where StructProperty with new properties fail schema upgrade

khanaffan:
This is a customer issue. Reported in connector. 

[[CS0421167] Cannot re-sync iModel - Error with no information in report](https://dev.azure.com/bentleycs/ProductAdvancement/_workitems/edit/1646730)

Backport PR: https://dev.azure.com/bentleycs/iModelTechnologies/_git/imodel02/pullrequest/520495

When Struct definition of structType property is updated and now it has new properties and collectively it does not fit in joined table we moved it into overflow table.

The bug was that when structType property was already in overflow table we allocated columns in joined table. Then attempt to move the whole struct to different columns in overflow table along with newly allocated joined table column.

The fix is struct that already is in overflow always allocate column in overflow table and thus its not moved at all.
```cpp
    // any new property of struct should go into overflow table if
    // previsouly older properties were mapped into overflow
    auto baseLineTable = compoundPropDiff.GetBaselineTable();
    if (baseLineTable != nullptr && baseLineTable->GetType() == DbTable::Type::Overflow) {
        ctx.m_classMap.GetColumnFactory().EnsurePropertyGoesToOverflow(property.GetName(), *ctx.m_importCtx);
    }
```
Note: this will be ported back to imodel02


rschili (approved):

rschili:
I don't think we should be pushing huge schemas like these. In the past I always went through the effort of stripping the scenario down to the bare minimum. These tests are adding ~10,000 lines of testdata just for one scenario. But I know how much work it is to clean those out, so I'm fine pushing it as is.

PR #1102, opened 2025-05-02
Improve error message when deleting element that is a definition element

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/8035

closes https://github.com/iTwin/itwinjs-core/issues/8029

This pull request enhances the error message displayed when trying to delete a definition element. Instead of a generic error, it now provides a specific message indicating that definition elements cannot be deleted directly. Developers are advised to use the `deleteDefinitionElements()` method instead. This change offers clearer guidance and helps prevent accidental deletions of crucial elements.

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/DgnCategory.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/DgnCategory.cpp
@@ -82,14 +82,6 @@ DgnCategoryId DgnCategory::QueryCategoryId(DgnDbR db, DgnCodeCR code)
     return DgnCategoryId(db.Elements().QueryElementIdByCode(code).GetValueUnchecked());
     }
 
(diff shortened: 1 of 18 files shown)


ColinKerr:
@pmconne and @bbastings do you see any issues with this change?

pmconne:
If trying to delete a subclass of `Bis.DefinitionElement` is **always** supposed to be an error then the changes are fine and consolidating the checks to the base class method is an improvement, though it shows that our API is faulty (`Element` should not provide a `delete` method if deleting some subclasses of `Element` is forbidden - Liskov subtitution principle etc).

Are people who have implemented their own subclasses of `DefinitionElement` in TypeScript going to be surprised to discover after updating to the next version of iTwin.js that their applications are suddenly throwing exceptions when they attempt to delete those elements?

PR #1103, opened 2025-05-02
Optionally return world CRS 

ben-polinsky:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/8037

There are a number of critical "world" CRS ("EPSG:4326" for example) which are not available from [getAvailableCoordinateReferenceSystems](https://github.com/iTwin/itwinjs-core/blob/06313ff2dc8671959008c377d92437e3bbe2f57a/core/backend/src/GeographicCRSServices.ts#L48). This is because [GeoServicesInterop::GetListOfGCS](https://github.com/iTwin/imodel-native/blob/644642cf7a86d9926c7c64aee51f4491045cca97/iModelJsNodeAddon/GeoServicesInterop.cpp#L75) filters out CRS without extents or those with extents spanning the globe.

This PR exposes an optional `includeWorld` argument to circumvent the above check.


ben-polinsky:
@bbastings Could you take a look?

iModelJsNodeAddon/GeoServicesInterop.cpp
```diff
@@ -85,8 +85,10 @@ bvector<CRSListResponseProps> GeoServicesInterop::GetListOfCRS(DRange2dCP extent
         DRange2d crsRange(DRange2d::From(crs->GetMinimumLongitude(),
                         crs->GetMinimumLatitude(),
                         crs->GetMaximumLongitude(),
-                        crs->GetMaximumLatitude()));
-        if (crsRange.IsEmpty() || crsRange.Area() == 64800) //64800 => Worldwide, not wanted
+                    crs->GetMaximumLatitude()));
+
+        // Don't include world crs (area == 64800) by default
+        if (!includeWorld && (crsRange.IsEmpty() || crsRange.Area() == 64800))
```
bbastings, line 91: As there was already the option to filter by "extent" it seems like you could have supported this by checking if extent is a null range (not pointer). It doesn't seem like passing both true for includeWorld and a non-null extent are compatible anyway (extent check can exclude large or non-existent "world" range). So something like this:

```
boolean isNullExtent = (extent && extent.IsNull());

if (!isNullExtent && (crsRange.IsEmpty() || crsRange.Area() == 64800)) // 64800 => Worldwide, not wanted
  {
  continue;
  }

// Don't include CRS if it does not contain the extent
if (extent && !isNullExtent)
  {
  DRange2d extentRange(DRange2d::From(extent->low.x, extent->low.y, extent->high.x, extent->high.y));
  if (!extentRange.IntersectsWith(crsRange))
    continue;
  }
```
Would just need to document the null range behavior for GetAvailableCoordinateReferenceSystemsArgs on the ts side...

Maybe @AlainRobertAtBentley has an opinion?

ben-polinsky: Reasonable.

> It doesn't seem like passing both true for includeWorld and a non-null extent are compatible anyway

I was thinking If someone wants to query by extent but also include "world" in the response - technically the extent range would then always intersect.

bbastings: > I was thinking If someone wants to query by extent but also include "world" in the response - technically the extent range would then always intersect.

Your comment said "CRS without extents" exist, so those would still be excluded...don't know if that's good/desired behavior or not.

ben-polinsky: Right, my assumption was that an empty `crsRange` also indicated a world CRS. However, since this code when isolated never hits with our current set of CRS:

```c++
  if (crsRange.IsEmpty()) 
  {
      continue; // never
  }
```

I'm not sure of the intent.

I think allowing world to be included in extent range queries makes a bit more sense than passing null. It's more flexible and does allow someone to get world crs back if querying a large extent range.

bbastings: Okay, sounds like the empty crsRange isn't a problem.

bbastings (approved):

ben-polinsky:

PR #1106, opened 2025-05-07
Add special optimization of ExternalSourceAspect and remove some unnecessary code

khanaffan:
Closes: https://github.com/iTwin/itwinjs-backlog/issues/1479

### Changes
* BloomFilter is enabled again. (Need Perf testing if it has any negative effect anymore)
* Disqualifing ECClassId lookup index for ExternalSourceAspect is removed.
* Disqualify JOIN expression for ExternalSourceAspect.Element.Id is added.

This query optimization involves ExternalSourceAspect, which is typically slow. A previously mentioned method was suggested to accelerate its performance.

### Reason:
ExternalSourceAspect.Element.Id is associated with the index (ix_bis_ElementMultiAspect_fk_bis_ElementOwnsMultiAspects_target), which closely matches Element.Id. Consequently, using this index for looping yields performance akin to a SCAN operation. Therefore, the index should be used as a last resort, not as a primary or secondary option.

The patch identifies the search expression 'ExternalSourceAspect.Element.Id' and appends a '+' to invalidate it. Some earlier optimizations related to this have been eliminated.

- [ X] Need ECPresentation performance testing as prerequisite to merge. @grigasp


khanaffan:
changes move to branch affank/multiaspect-optimization

PR #1107, opened 2025-05-07
macOS: Fix crash during shutdown

tcobbs-bentley:
During shutdown, if there are any SQLite Cloud Containers still open, they get closed as part of the `BeSQLite::CloudCache` destructor. As part of this, two messages get sent to the JavaScript code (`onDisconnect` and `onDisconnected`). Both of these cause a crash on macOS. I would expect them to also crash on other platforms, but since I haven't seen any evidence of that, I made the fix Apple-specific. If it is determined that the crash happens on all platforms, the fix can be changed to apply to all platforms.


tcobbs-bentley:
@DanRod1999 confirmed that this crash does not happen on Windows. I don't know why it only happens on macOS, but I am leaving the changes Apple-specific.

aruniverse (approved):
approving, but @khanaffan plz review as well

PR #1111, opened 2025-05-13
Prefix sqlite3 symbols

mathieu-lemuzic:
Prefix sqlite3 symbols to integrate imodel-native libs in cesium tilers repo.


pmconne:
@khanaffan thoughts?

mathieu-lemuzic:
> @khanaffan thoughts?

I checked with affan offline I'm going to try a different approach before resorting to rename in the repo

PR #1124, opened 2025-05-21
Improved error message for ambiguous instance queries

Abhijeet-Bentley:
Issue Link: https://github.com/iTwin/itwinjs-core/issues/7565

-Added test to reproduce and test behavior
-Added improve error message

### DIFF
--- a/iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
+++ b/iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
@@ -404,10 +404,11 @@ BentleyStatus PropertyNameExp::ResolveColumnRef(ECSqlParseContext& ctx)
     const auto inherited = matchProps.size() - local;
 
     if (!(local == 1 || inherited == 1)) {
-
-        if (this->GetParent()->GetType() == Exp::Type::ExtractProperty){
+        auto parent = this->GetParent();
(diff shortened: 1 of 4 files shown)


iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
```diff
@@ -402,7 +402,15 @@ BentleyStatus PropertyNameExp::ResolveColumnRef(ECSqlParseContext& ctx)
 
     const auto local = countLocalRefs(matchProps);
     const auto inherited = matchProps.size() - local;
+
     if (!(local == 1 || inherited == 1)) {
+
+        if (this->GetParent()->GetType() == Exp::Type::ExtractProperty){
```
rschili, line 408: I guess we cannot his a PropertyNameExp with no parent, still GetParent returns a pointer so a nullcheck seems in order.

rschili (approved):

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
Abhijeet-Bentley: I have added a null check to "parent" and also, the ToString() value is stored in a variable "TargetPath"

rschili (approved):

PR #1125, opened 2025-05-21
Adjust confusing log messages during schema import

rschili:
OpenSite+ while migrating to iTwin.js 5 noticed some confusing log messages. Turned out they were suggesting a failed schema import which actually worked fine.

Our new mechanism of copying schemas during import which are located through ECDb caused a cascade of log messages complaining about modified schemas that do not have their version incremented. This message is now skipped if the only changes is to the schema's references.

Also cleaned up some mixup of wchar and char in log messages.


ColinKerr (approved):

DanRod1999:

PR #1135, opened 2025-05-28
Update rapidjson

mathieu-lemuzic:


DanRod1999:
updating third parties should be done following our wiki on UpdatingThirdPartyLibraries. The branches for this process were temporarily deleted, but were just restored. We should update those branches as well before proceeding I think. @chuckkir @naveedkhan8067 do you guys know the best way to do that if this PR/branch is already opened? Is cherry picking certain commits an option?

mathieu-lemuzic:
@DanRod1999 FYI, the `libsrc-Main` branch built successfully: https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=3985932&view=results

PR #1139, opened 2025-05-30
curl 8.14.0

DanRod1999:
update curl from 8.12 to 8.14 to respond to cve's:
https://www.mend.io/vulnerability-database/CVE-2025-5025
https://www.mend.io/vulnerability-database/CVE-2025-4947


aruniverse:

MichaelSwigerAtBentley:
Also, do we need to pull in the rapidjson update from libsrc-Main: https://github.com/iTwin/imodel-native/commits/libsrc-Main/

tcobbs-bentley:
It seems like the commits come from the way we update 3rd-party libraries (libsrc). Having said that, there are a bunch of changes that have nothing to do with curl in this PR, and I don't think that is correct. I don't know if the problem stems from previous changes being made to libsrc without following the process, or something else, but it definitely feels like this PR is going to break things due to all of its non-curl changes.

tcobbs-bentley:
Looks like there weren't a bunch of changes, but only the addition of 5 header files in `libsrc/compress/zlib`. It doesn't seem like adding these should break anything, but it also doesn't seem like they should be showing up as new files in a libcurl PR.

PR #1146, opened 2025-06-04
Update sqlite to 3.50

khanaffan:


pmconne (approved):

khanaffan (commented):
This upgrade triggers a UPSERT operation on `propertyMap` during schema synchronization, causing SQLite to crash due to a rehashing bug that runs indefinitely and eventually fails from memory exhaustion.

This issue appears to be data-driven, indicating a logical error in SQLite.

The bug has been reported to SQLite.

PR #1160, opened 2025-06-19
libxml2 2.14.4

DanRod1999:
Attempt to resolve:
[CVE-2025-49795](https://nvd.nist.gov/vuln/detail/CVE-2025-49795)

This does not resolve the recent cve's, however it does fix some newly deprecated apis introduced in 2.14, so upgrading once the cve's are fixed should be easy. However, not sure if we want to proceed with this PR. Currently there is no timeline for resolving them.


iModelCore/libsrc/libxml2/vendor/libxml.h
```diff
@@ -34,22 +34,43 @@
 #else
     /* BENTLEY CHANGES */
     #if defined (_MSC_VER)
-        #include "include/win32config.h"
+        #include "win32/win32config.h"
     #else
         #include "../BeLibxml-unix-config.h"
+        #define XML_SYSCONFDIR "/etc"
```
DanRod1999, line 40: this was added by myself to get this to work in linux/mac. I don't think its the correct placement, @tcobbs-bentley I know you resolved this issue on your side. Where did you put it/any suggestions

DanRod1999: i also don't like introducing changes to src code, but I just realized we already have our own set of custom changes to this file, so maybe this is fine? Or maybe worth seeing why/if these changes are still needed?

tcobbs-bentley: (With the ifndef wrapper.)

tcobbs-bentley:
@DanRod1999  Do have a list of changes you made inside `libxml2/vendor`? I don't think there's any point reviewing the actual libxml2 changes, but it would be good to review any change you made to that code.

iModelCore/libsrc/libxml2/vendor/libxml.h
DanRod1999: oh sorry duh. Didn't realize that was our, but windows config came with the pkg

iModelCore/libsrc/libxml2/BeLibxml-unix-config.h
```diff
@@ -337,3 +337,8 @@
 
 /* Include multi-threading support like win32 already does. This triggers more in xmlversion.h. */
 #define __MT__
+
+/* Include XML_SYSCONFDIR for unix. As of 2.14.4 this Macro is no longer defined in libxml.h */
+#ifndef _MS_VER
```
tcobbs-bentley, line 342:
```suggestion
#ifndef XML_SYSCONFDIR
```

tcobbs-bentley: That way if it ever gets defined somewhere else (like a makefile), this will leave it alone.

tcobbs-bentley (approved):

pmconne (approved):

PR #1180, opened 2025-07-30
TxnManager.deleteAllTxns not deleting briefcase changes

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-core/issues/8197

itwinjs-core: <TBD>

The function DeleteAllTxns was simply clearing the Dgn_Txns table without reverting/reversing any of the local changes.

This PR updates the function DeleteAllTxns to now reverse the changes before clearing the Dgn_Txns table.
A new function ClearAllTxns has been added which does that DeleteAllTxns used to do.
This was needed as we have tests where we want to setup data and clear the txns before starting a test.

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
@@ -1998,12 +1998,27 @@ Utf8String TxnManager::GetRedoString() {
 /*---------------------------------------------------------------------------------**//**
  @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
+void TxnManager::ClearAllTxns() {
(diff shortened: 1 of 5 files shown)


iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
```diff
@@ -2004,6 +2018,7 @@ void TxnManager::DeleteAllTxns() {
     }
 
     m_dgndb.SaveChanges(); // in case there are outstanding changes that will create a new Txn
+    ReverseAll();
```
khanaffan, line 2021: This only reverse txn in current session. Again discardChanges() does full reverse down to first session.

PR #1187, opened 2025-08-06
When two object are changed by two users rebase does not work as intended

RohitPtnkr1996:
Fixes https://github.com/iTwin/itwinjs-core/issues/7764

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/8396


iModelCore/BeSQLite/SQLite/sqlite3.c
```diff
@@ -233032,7 +233032,7 @@ static int sessionApplyOneOp(
 
     sqlite3_step(p->pDelete);
     rc = sqlite3_reset(p->pDelete);
-    if( rc==SQLITE_OK && sqlite3_changes(p->db)==0 && p->bIgnoreNoop==0 ){
```
khanaffan, line 233035: You cannot do that we override this file. We need to send this to sqlite dev and let them push a fix they will also cover it with test.

PR #1198, opened 2025-08-21
Update geometry guid on subcategory appearance change

pmconne:
Fixes https://github.com/iTwin/itwinjs-core/issues/8176.

When a subcategory's appearance changes, treat it as a change to the geometry of all elements belonging to that subcategory's category.
This will always be true if the category only has a single subcategory, which seems to be where authoring apps are headed since people don't really find subcategories useful/intuitive. It will not necessarily be true for categories with multiple subcategories, but to identify the exact elements affected by a subcategory change would require parsing the geometry stream of every element belonging to the category, which would be way too slow.

Update the geometry GUID of every geometric model containing at least one element belonging to the affected category. This invalidates previously-generated tiles for those models.

If we're tracking element geometry changes (i.e., inside a `GraphicalEditingScope`), also dispatch geometry-changed events for every affected element, so the front-end will know to obtain updated graphics for them.

`ModelChanges` was using this `ChangedIds` type which stored a `fromCommit` flag for each changed element or model. I reduced this to a single `fromCommit` flag on `ModelChanges` itself, because you're always either committing (`fromCommit=true`) or applying a changeset (`fromCommit=false`) - you'll never have a mix of both.


bbastings (dismissed):
Non-query changes LGTM.

aruniverse:
@mergifyio backport release/5.1.x

pmconne:
@Mergifyio backport release/5.1.x

Issue #1203, opened 2025-08-28
IdSet does not work with UNION or CTE


simihartstein:
**To Reproduce**
```sql
WITH RECURSIVE cte (ECInstanceId) AS (
    SELECT Id from IdSet('["1"]')
    UNION
    SELECT Id from IdSet('["1"]') ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES
) SELECT * FROM cte ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES
```

- iTwin.js Version 5.X

PR #1231, opened 2025-10-03
Fixes the bug of IdSet

soham-bentley:
Fixes https://github.com/iTwin/imodel-native/issues/1203


khanaffan (approved):

rschili:
Quick question: I see you fixed a bug in iModelCore/ECDb/ECDb/ECSql/SelectStatementExp.h, but I wonder how that broke IdSet but nothing else. What you fixed seems like a more generic bug unrelated to IdSet.

PR #1232, opened 2025-10-06
Update openssl to 3.2.6

DanRod1999:
Update openssl to 3.2.6

Notable changes made:
- needed to run openssl configure for windows to process .in files
- had revert Configuration.h to keep bentley changes but add new vals from this minor update
- also reverted changes made to aes-x86_64.asm as they broke the build
- update mke and def file to add/remove new files and defs

Configuring .in files may have caused unnecessary files to be generated. These can be potentially removed


pmconne (approved):

DanRod1999:
@mergifyio backport release/5.2.x

PR #1234, opened 2025-10-08
ECSqlReader/ConcurrentQuery unable to see modification to file

khanaffan:
`ConcurrentQuery` relies on `PRAGMA data_version` to detect if the database file has been modified by another connection. However, this mechanism doesn't work reliably when there are active statements, as these likely lock the database and prevent `PRAGMA data_version` from reflecting any changes until those statements are closed.

In ECDb, `InstanceReader` maintains active statements for performance reasons. For example, executing a query like `SELECT $ FROM Bis.Element` creates an active statement, which in turn blocks `PRAGMA data_version` from updating.

As a workaround, we now check the primary connection to determine if it has changed since the last time a query was executed via `ConcurrentQuery`. If a change is detected, we clear the cache to ensure consistency.

### DIFF
--- a/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
+++ b/iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
@@ -176,6 +176,18 @@ void CachedConnection::SyncAttachDbs() {
 // @bsimethod
 //---------------------------------------------------------------------------------------
 void CachedConnection::Execute(std::function<void(QueryAdaptorCache&,RunnableRequestBase&)> cb, std::unique_ptr<RunnableRequestBase> request) {
+    if (m_db.IsDbOpen()) {
+        // Check if the primary file data version has changed
+        uint32_t dataVersion;
+        if (BE_SQLITE_OK == GetPrimaryDb().GetFileDataVersion(dataVersion)) {
+            if (dataVersion != m_primaryFileDataVer) {
+                m_db.ClearECDbCache();
+                m_db.ClearDbCache();
+                m_primaryFileDataVer = dataVersion;
+            }
+        }
+    }
+
     SyncAttachDbs();
     SetRequest(std::move(request));
     cb(m_adaptorCache, *m_request);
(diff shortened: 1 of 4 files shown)


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -176,6 +176,18 @@ void CachedConnection::SyncAttachDbs() {
...
+        // Check if the primary file data version has changed
+        uint32_t dataVersion;
+        if (BE_SQLITE_OK == GetPrimaryDb().GetFileDataVersion(dataVersion)) {
+            if (dataVersion != m_primaryFileDataVer) {
+                m_db.ClearECDbCache();
+                m_db.ClearDbCache();
+                m_primaryFileDataVer = dataVersion;
+            }
+        }
+    }
```
grigasp, line 189: Is this something we should be doing in ECPresentation as well with our secondary connections?

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -176,6 +176,18 @@ void CachedConnection::SyncAttachDbs() {
 // @bsimethod
 //---------------------------------------------------------------------------------------
 void CachedConnection::Execute(std::function<void(QueryAdaptorCache&,RunnableRequestBase&)> cb, std::unique_ptr<RunnableRequestBase> request) {
+    if (m_db.IsDbOpen()) {
+        // Check if the primary file data version has changed
+        uint32_t dataVersion;
+        if (BE_SQLITE_OK == GetPrimaryDb().GetFileDataVersion(dataVersion)) {
+            if (dataVersion != m_primaryFileDataVer) {
+                m_db.ClearECDbCache();
```
rschili, line 184: This fix looks correct to me, however I'm a bit worried about performance regression. Clearing the cache is not always cheap, for example it calls `ConcurrentQueryMgr::Shutdown` and uses a Mutex and triggers OnBefore/OnAfterClear events, god knows what this all entails. We should monitor if this affects performance, and maybe put some tracker here that helps us understand how often we clear the caches now.

rschili (approved):

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
grigasp: Sounds good. I don't think we're going to make a switch to instance queries in the native ECPresentation library, but I'll keep this in mind if we do.

PR #1247, opened 2025-10-14
Incorporate blockcache changes from cloud sqlite trunk

tcobbs-bentley:
These changes came from downloading the zip in [this](https://sqlite.org/cloudsqlite/info/f9629a851205c307) Cloud SQLite check-in and incorporating the bcv-related updates. There were other changes in other parts of the codebase, but I didn't understand them sufficiently to know if brining them in was acceptable.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/8696


tcobbs-bentley:
> > There were other changes in other parts of the codebase, but I didn't understand them sufficiently to know if brining them in was acceptable.
> 
> Seems wrong for us to not be copying changes from upstream. What are you doing to try and understand what its doing? Have you talked to anyone?
> 
> @wgoehrig plz advise and review

There are hundreds (perhaps thousands) of changes in places like sqlite3.c. Maybe all of them need to be brought in, but I don't know enough about SQLite to know. I agree that this is something that needs to be investigated, but this PR simply updates the bcv files so that if Cloud SQLite accepts my changes, the files will match. There were only a few changes in these files.

wgoehrig:
Typically when updating the "Cloud Backed SQLite" extension I only pull in files that start with `bcv` or `blockcache`.

We should only update changes to SQLite itself from their official releases; the branch that the SQLite team uses to build and test the cloud SQLite extension is not always up to date.

@khanaffan usually handles updating SQLite, I'll defer to him on whether we're due for an update.

PR #1256, opened 2025-10-22
Added support for CTE in QueryView

soham-bentley:
Fixes https://github.com/iTwin/itwinjs-core/issues/8573


khanaffan (approved):

rschili:
> What version do I have to increment for this? Because CTE support is now being added for Views Like for ECSQL new features we upgrade ECSql version. But according to me, this is not a new feature of ECSql, right?
> 
> @khanaffan @rschili @ColinKerr @RohitPtnkr1996

It's not exactly an ECSql feature strictly speaking.... it's more like an ECDb feature.... mmh, maybe we should increment the ECDb version instead. It seems drastic, but thinking about it, yeah, it may be the more fitting number.

In any case, we should increment something, so we can tell if this is supported.

soham-bentley:
> > > What version do I have to increment for this? Because CTE support is now being added for Views Like for ECSQL new features we upgrade ECSql version. But according to me, this is not a new feature of ECSql, right?
> > > @khanaffan @rschili @ColinKerr @RohitPtnkr1996
> > 
> > 
> > It's not exactly an ECSql feature strictly speaking.... it's more like an ECDb feature.... mmh, maybe we should increment the ECDb version instead. It seems drastic, but thinking about it, yeah, it may be the more fitting number.
> > In any case, we should increment something, so we can tell if this is supported.
> 
> I would ask how incrementing the profile version would help. e.g. if we increment the profile version then someone could mark the schema with the CA saying that it needs that version at runtime. What would the behavior be if the schema designer did that? How would that compare to what would happen if we did nothing. I support updating the version so long as we define the behavior it would allow and we agree it is the behavior we want.
> 
> This would be an increment to the 4th, least significant digit of the profile so it should be relatively low disruption.

So the final call is incrementing profile to 4.0.0.6, right? @ColinKerr  @rschili

PR #1260, opened 2025-10-29
remove assert(false)

DanRod1999:
Remove assert thats causing local windows build to fail


aruniverse:
run git blame and find who added this and find out why

pmconne:
> run git blame and find who added this and find out why

It's ancient. This was a common practice in our native code by certain people - they handle the unexpected situation, but also assert so that they will get notified of the situation in debug builds.

aruniverse (approved):

khanaffan:
@aruniverse  note that `BeFileName` is not even required. C++17 now have [std::filesystem](https://en.cppreference.com/w/cpp/filesystem.html).

PR #1312, opened 2025-12-24
Add MsgPack library (do not review or merge)

khanaffan:
This is for experimentation


khanaffan:
msgpack was to investigate alternative way to store dynamic properties. This was 6 months ago now we have msgpack extension for sqlite so we can use that instead of the c lib.

aruniverse:
> This was 6 months ago now we have msgpack extension for sqlite

can you link to this?

PR #1368, opened 2026-03-26
Add native ChangesetReader

soham-bentley:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/1865
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9140


khanaffan (approved):

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,944 @@
...
+
+    const auto& sysMap = propertyMap.GetAs<SystemPropertyMap>();
+    const auto* dataMap = sysMap.FindDataPropertyMap(dbTable);
+    if (dataMap == nullptr) {
+        LOG.infov("No data property map found for system property '%s' in table '%s'.",
+                   propertyMap.GetProperty().GetName().c_str(), dbTable.GetName().c_str());
+        return BE_SQLITE_OK;
+    }
+
+    if (!IsInMap(dataMap->GetColumn().GetName(), columnValues)) {
```
soham-bentley, line 472: In some case the actual column behind a system property is virtual. If the column is virtual it will not be in the changeset, in that case the current logic will skip it
Is this fine?
@khanaffan

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,944 @@
...
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+bool ChangesetValueFactory::TryResolveClassMapFromDbSeek(
+    DbTable const& dbTable,
+    ColumnValueMap const& columnValues,
+    ECDbCR conn, const ClassMap*& classMapOut, ECClassId& classIdOut) {
+
+    // A virtual class-id column has no physical storage — nothing to read from the DB.
+    DbColumn const& classIdCol = dbTable.GetECClassIdColumn();
+    if (classIdCol.IsVirtual())
```
soham-bentley, line 697: Also if the classIdCol is virtual we don't try to read it from changeset or from live db because it is virtual and we finally fallback to the rootExclusiveclassId of the current table being changed.
This seems like a logical thing to do, right?
@khanaffan

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,944 @@
...
+    const auto& idPropMap = navMap.GetIdPropertyMap();
+    const DbColumn& idCol = idPropMap.GetAs<SingleColumnDataPropertyMap>().GetColumn();
+    const auto& relClassIdMap = navMap.GetRelECClassIdPropertyMap();
+
+    const bool hasIdInChangeset = IsInMap(idCol.GetName(), columnValues);
+    const bool relClassIdIsVirtual = relClassIdMap.GetColumn().IsVirtual();
+    const bool hasPhysicalRelClassIdInChangeset =
+        !relClassIdIsVirtual && IsInMap(relClassIdMap.GetColumn().GetName(), columnValues);
+
+    if (!hasIdInChangeset && !hasPhysicalRelClassIdInChangeset) {
```
soham-bentley: In case of Nav props we skip if both the `Id` and `RelECClassId` are not in changeset. If one of them is there we don't skip it
Sounds right?
@khanaffan

khanaffan (approved):

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,959 @@
...
+            continue;
+        const auto extType = ExtendedTypeHelper::GetExtendedType(prim->GetExtendedTypeName());
+        if (extType != ExtendedTypeHelper::ExtendedType::Id)
+            continue;
+        if (!propertyMap->GetProperty().GetName().EqualsIAscii(ECDBSYS_PROP_ECInstanceId))
+            continue;
+
+        const auto& sysMap = propertyMap->GetAs<SystemPropertyMap>();
+        const SystemPropertyMap::PerTableIdPropertyMap* dataMap = sysMap.FindDataPropertyMap(primaryDbTable);
+        if(dataMap == nullptr && primaryDbTable.GetType() == DbTable::Type::Overflow) {
```
soham-bentley, line 735: I checked in a test that we don't find data property maps of the ECInstanceId system property map in overflow tables, although the corresponding column data is still there in the changeset (which means the ECisnatnceIfd column is there in the overflow table as well) [I don't know if this is expected behavior or is a bug]
So for overflow tables
I am finding the data property map from the parent table of this overflow table and then getting the data property map.

Then I am getting the column associated with that data property map and then finding value for that column in the changeset.

According to me this works because I think for overflow tables we just take the primary keys of the parent table and have it as is in the overflow table, so the primary key column names remain the same
And we fetch the data from the changeset using the same column name

Does this logic sound right? @khanaffan

iModelCore/ECDb/ECDb/ChangesetReaderImpl.cpp
```diff
@@ -0,0 +1,204 @@
...
+    return m_prepared->GetColumnCount(stage);
+}
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+BentleyStatus ECChangesetReader::Impl::GetInstanceKey(Stage stage, Utf8StringR key) const {
+    if (!IsPrepared())
+        return ERROR;
+    return m_prepared->GetInstanceKey(stage, key);
```
rschili, line 114: On the N-API side, out["key"] = instanceKey.c_str() assigns "", which flows into $meta.nativeKey = "". In the unifier this builds a key of "-new" or "-old", making all such instances hash into the same cache slot - one overwrites the next, silently. Should be return ERROR so the N-API wrapper throws and it fails loudly. In practice this shouldn't happen if ChangesetValueFactory::Create always inserts both system properties.

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
```diff
@@ -0,0 +1,84 @@
...
+    // data fields
+    std::unordered_map<Stage, std::vector<std::unique_ptr<IECSqlValue>>> m_fields;
+    std::vector<Utf8String> m_changedPropNames;
+
+    //filters
+    std::vector<Utf8String> m_tableFilters;
+    std::vector<DbOpcode> m_opcodeFilters;
+    std::vector<ECN::ECClassId> m_ecclassIdFilters;
+
+    PreparedECChangesetReader(PreparedECChangesetReader const&) = delete;
```
rschili: The issue requirement says "no excessive memory usage for very large changesets." A ChangeSet group squashed to one in-memory ChangeSet is the opposite of streaming. Should we stream here? Or is this just a first version that will later be optimized?

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
```diff
@@ -0,0 +1,519 @@
...
+    if (!input.DoesPathExist())
+        return BE_SQLITE_CANTOPEN;
+
+    bvector<BeFileName> files{input};
+    auto reader = std::make_unique<ChangesetFileReaderBase>(files);
+    DdlChanges ddlChanges;
+    bool hasSchemaChanges;
+    reader->MakeReader()->GetSchemaChanges(hasSchemaChanges, ddlChanges);
+
+    UNUSED_VARIABLE(ddlChanges);
```
rschili: What is this? Copilot flagged GetSchemaChanges as an unnecessary call, and you added these macros to silence the warning, but WHY are we calling it we we don't use any of the results? The code also ignores the DbResult returned from GetSchemaChanges().
Is this for debugging purposes?

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
```diff
@@ -0,0 +1,519 @@
...
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+DbResult PreparedECChangesetReader::OpenGroup(T_Utf8StringVector const& files, bool invert, Mode mode) {
+    if(IsOpen()) {
+        LOG.errorv("Attempting to open a file on an already open PreparedECChangesetReader.");
+        return BE_SQLITE_ERROR;
+    }
+    m_changeGroup = std::make_unique<ChangeGroup>(m_ecdb);
+    DdlChanges ddlGroup;
```
rschili: ddlGroup is collected but never passed anywhere and just dies at the end of scope. Looks like unfinished or debug code?

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
```diff
@@ -0,0 +1,519 @@
...
+    ClearTableFilters();
+    ClearOpcodeFilters();
+    ClearECClassIdFilters();
+}
+
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+DbResult PreparedECChangesetReader::Step() {
```
rschili: I do not like the "hollow rows" concept when we use filters. Normally I'd expect filtered rows to not be returned from Step(). We should just internally skip over them. You documented this behavior properly, but it still seems counter intuitive to me. At least leave a comment that defends the design decision.

iModelCore/ECDb/ECDb/PreparedChangesetReader.cpp
```diff
@@ -0,0 +1,519 @@
...
+    }
+    if (!IsStepped()) {
+        LOG.errorv("Attempting to get column values from a PreparedECChangesetReader that has not been stepped or is on an invalid change.");
+        return ERROR;
+    }
+    Utf8String tableName;
+    if (GetTableName(tableName) != SUCCESS)
+        return ERROR;
+
+    CachedStatementPtr stmt = m_ecdb.GetCachedStatement("SELECT [name] FROM PRAGMA_TABLE_INFO(?) ORDER BY [cid]");
```
rschili, line 250: In a changeset with 10k rows touching the same table, that's 20k PRAGMA queries.
The mapping per table is immutable, maybe cache it?

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
```diff
@@ -0,0 +1,519 @@
...
+    if(!IsOpen()) {
+        LOG.errorv("Attempting to step a closed PreparedECChangesetReader.");
+        return BE_SQLITE_ERROR;
+    }
+    if (m_changes == nullptr) {
+        m_changes = std::make_unique<Changes>(*m_changeStream, m_invert);
+        m_currentChange = m_changes->begin();
+    } else {
+        if(m_currentChange.IsValid()) ++m_currentChange;
```
rschili: Calls sqlite3changeset_next. Runs before ClearFields() inside ReFetchValues(). The old ChangesetPrimitiveValue objects hold invalidated sqlite3_value* pointers during that window. Safe in practice since nothing reads them, but fragile - ClearFields() should come before the advance.

iModelCore/ECDb/ECDb/ECChangesetReaderImpl.cpp
```diff
@@ -0,0 +1,204 @@
...
+USING_NAMESPACE_BENTLEY_EC
+
+BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+DbResult ECChangesetReader::Impl::OpenFile(ECDbCR ecdb, Utf8StringCR file, bool invert, Mode mode) {
+    if (!IsPrepared())
+        m_prepared = std::make_unique<PreparedECChangesetReader>(ecdb);
```
rschili: Seems inconsistent, here we reuse a prepared reader while the other methods always create new.

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,954 @@
...
+        if (status != SUCCESS)
+            return status;
+    }
+    return SUCCESS;
+}
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+bool ChangesetValueFactory::isChildClassOfBisCore(ECClassId classId, ECDbCR conn) {
```
rschili: Better call this IsDerivedFromBisElement

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
```diff
@@ -0,0 +1,954 @@
...
+    // -----------------------------------------------------------------------
+    // Step 4: Build remaining property fields (slots [2+]).
+    // -----------------------------------------------------------------------
+    fields.emplace_back(std::move(instanceIdField));
+    fields.emplace_back(std::move(classIdField));
+
+    if (mode == ECChangesetReader::Mode::Instance_Key)
+        return SUCCESS; // caller only needs the instance key — skip user properties
+
+    if(mode == ECChangesetReader::Mode::Bis_Element_Properties && isChildClassOfBisCore(resolvedClassId, conn) && !tbl.GetName().EqualsIAscii("bis_Element"))
```
rschili: The Bis_Element_Properties mode emits ALL properties for classes that aren't BIS Element derivatives. Is that intended? If so, it needs a doc comment explaining the asymmetry.

iModelCore/ECDb/ECDb/PreparedChangesetReader.cpp
soham-bentley: There shouldn't be 10k rows touching the same table in a changeset. As far as I know there is a concept known as `coalescing` https://sqlite.org/sessionintro.html#:~:text=After%20a%20session%20object%20has,no%20new%20information%20is%20recorded.
image

@khanaffan please correct me if I am wrong

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
soham-bentley: Our api is designed to step rows one by one. And before I step I don't know whether to include it or filter it out so for rows which are not part of the filter I simply don't compute the fields(which is the heavy part of the changeset reader api)

soham-bentley: Comment added explaining it

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
soham-bentley: Removed

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
soham-bentley: Removed unnecessary stuff

iModelCore/ECDb/ECDb/ChangesetReaderImpl.cpp
soham-bentley: Not really on typescript side either we get the column values and the instance key and the changeFetchedPropNames or we just get undefined and not get any of these
It can never happen that  instance key will be an empty string
Because while computing instance key I finally check if both the parts of it are non empty then only return SUCCESS or else ERROR.

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
soham-bentley: Done

iModelCore/ECDb/ECDb/ECChangesetReaderImpl.cpp
soham-bentley: Made consistent

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
soham-bentley: Done

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
soham-bentley: I do have a clear documentation on the typescript side in the ECChangesetReader.md file

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
soham-bentley: I think currently we do put all the changesetFiles into a single Change Group, that actually causes `coalescing`, meaning an insert to a table in changeset 1 and then update to the same row in changeset 2 will just result in an insert overall.
Now if we do streaming of the files one by one I guess we will loose out on this `coalescing`. I don't think (logically) SQlite even has an api where we get both `coalescing` as well as file by file streaming. Thats understandable because to have `coalescing` we must have all the changesets in memory.

Even if there is an api from SQLITE which support both `coalescing` and `streaming` , incorporating that needs a change in our `BeSQLITE` layer preferably in the change group api and change stream api, which I guess is beyond the scope of this PR and should be handled separately(according to me).

@khanaffan can take the final call

rschili: The question is which matters more, streaming or coalescing. I'll leave the call to you. Have no strong opinion on this.

iModelCore/ECDb/ECDb/PreparedECChangesetReader.cpp
rschili: Yes that is how the method inside works, but why would it return SUCCESS for a row that was filtered, instead of immediately moving on to the next row? I don't think I underatand.

soham-bentley: Hmm you are actually right....made it true skipping of rows

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
soham-bentley: Yes I think @khanaffan should take the final call here. But even if we want streaming of changesets from a change group I think a proper implementation should be done on the BeSQLITE level instead of doing it on this API level, which according to me is beyond the scope of this PR. But I am open to doing what @khanaffan decides finally. 😄

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
soham-bentley: This rename is done @rschili I guess you can resolve this

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
soham-bentley: Just an idea, should we have some check on the number of changeset files that can be passed to be opened as a change group?
Like we can open maximum of 5 changeset files as a changeset group. Not more than that. Although this does not guarantee much as each changeset file can still be quite big, but still it is something.

rschili: That seems like magic number logic to me. How would we choose a good number? A single changeset may be 12KB or 1GB.

rschili (approved):
The Two PR titles are both different tense from the story title. I suggest we make this a bit more verbose.

iTwin:
Add NativeECChangesetReader API with lazy loading and filtering
native:
Add native ECChangesetReader

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
```diff
@@ -0,0 +1,83 @@
...
+    std::vector<Utf8String> m_tableFilters;
+    std::vector<DbOpcode> m_opcodeFilters;
+    std::vector<Utf8String> m_ecclassNameFilters;
+
+    PreparedECChangesetReader(PreparedECChangesetReader const&) = delete;
+    PreparedECChangesetReader& operator=(PreparedECChangesetReader const&) = delete;
+    void ClearFields();
+    BentleyStatus ReFetchValues(bool& isCurrentRowFilteredOut);
+    // Utf8String GetTableName() const { return m_currentChange.GetTableName(); };
+    // DbOpcode GetOpcode() const { return m_currentChange.GetOpcode(); };
```
khanaffan:
```suggestion
```

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
khanaffan: have a few thoughts. The main thing this PR seems to be missing is support for streaming `ChangeGroups`. Since `ChangeGroup` is backed by a hash table, it can consume a significant amount of memory. While it doesn’t allocate a single large chunk—so it’s less likely to immediately cause an OOM—systems can still have tight memory limits.

One possible approach would be to require enough memory to group the changeset initially, then write the grouped data to disk and stream it from there. This would cause a temporary peak in memory usage, but the memory would be released afterward.

@soham-bentley, I think this PR is already large enough. Could you create a separate issue to handle streaming for large change groups and investigate that independently?

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
khanaffan: ECInstanceId is a system property. We have SystemProperty map that has Column property map for each table. So ECInstanceId property map will have collection of column map but per table.

khanaffan (approved):

iModelCore/ECDb/ECDb/ChangesetValueFactory.cpp
soham-bentley: Yes when I try to find the ColumnPropertyMap of ECInstanceId system property in `Overflow` tables, I don't find it. Isn't that a bug? I should be able to find it because even in `Overflow` tables the primary key column is there. Should I log an issue for this?

Anyways, what I do now is I try to find the the ColumnPropertyMap of ECInstanceId system property in the parent table of the `Overflow` table, get the column and then as ECInstanceId is primary key I then find the value from the changeset. This logic works okay for me at least.  
@khanaffan

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
soham-bentley: Sure, Added new issue

iModelCore/ECDb/ECDb/PreparedECChangesetReader.h
soham-bentley: Done

PR #1369, opened 2026-03-29
Expose pragma for SchemaView for fast schema access

rschili:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9151

Please see the itwinjs-core PR for a comprehensive description.


iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,859 @@
...
+                size_t ccCountPos = m_output.size();
+                PutU8(0);
+                uint8_t ccCount = 0;
+                while (constrClassStmt.Step() == BE_SQLITE_ROW)
+                    {
+                    PutSRef(Safe(constrClassStmt.GetValueText(0)));
+                    PutSRef(Safe(constrClassStmt.GetValueText(1)));
+                    ccCount++;
+                    }
+                m_output[ccCountPos] = ccCount;
```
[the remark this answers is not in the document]
rschili: Fixed: added explicit UINT8_MAX overflow checks for both constrCount and ccCount that mirror the pattern used for baseCount and the property ref count. Overflow now returns BE_SQLITE_ERROR with a descriptive log line instead of silently wrapping.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,859 @@
...
+* See LICENSE.md in the repository root for full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+#include "ECDbPch.h"
+#include "SchemaViewWriter.h"
+#include <pugixml/src/pugixml.hpp>
+#include <cinttypes>
+
+BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
+
+static constexpr uint32_t MAGIC = 0x43534348; // "CSCH" - Indicates the file is an EC schema blob."
```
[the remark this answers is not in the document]
rschili: Fixed - the trailing quote is gone.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,869 @@
...
+//    SchemaUpgradeCustomAttributes. Their classes are only CustomAttribute and Struct
+//    types used for decoration. Since the blob doesn't include CA instances, including
+//    these schema definitions provides little value.
+// The list is checked case-insensitively via BeStringUtilities::StricmpAscii.
+static bool IsExcludedSchema(Utf8CP name)
+    {
+    struct CiLess {
+        bool operator()(std::string const& a, std::string const& b) const { return BeStringUtilities::StricmpAscii(a.c_str(), b.c_str()) < 0; }
+    };
+    static const std::set<std::string, CiLess> s_excluded = {
```
khanaffan: Wondering if this is super set of this
```cpp
bool ECSchema::IsStandardSchema () const
    {
    return IsStandardSchema(m_key.GetName());
    }

static Utf8CP s_originalStandardSchemaFullNames[] =
    {
    "Bentley_Standard_CustomAttributes.01.00.00",
    "Bentley_Standard_Classes.01.00.00",
    "EditorCustomAttributes.01.00.00",
    "Bentley_Common_Classes.01.00.00",
    "Dimension_Schema.01.00.00",
    "iip_mdb_customAttributes.01.00.00",
    "KindOfQuantity_Schema.01.00.00",
    "rdl_customAttributes.01.00.00",
    "SIUnitSystemDefaults.01.00.00",
    "Unit_Attributes.01.00.00",
    "Units_Schema.01.00.00",
    "USCustomaryUnitSystemDefaults.01.00.00"
    };
```

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,869 @@
...
+
+//---------------------------------------------------------------------------------------
+// Pre-collect IDs of schemas in the exclusion list. This avoids per-row name checks
+// and enables SQL-level filtering in the property query.
+//---------------------------------------------------------------------------------------
+DbResult SchemaViewWriter::CollectExcludedSchemaIds(DbCR db)
+    {
+    m_excludedSchemaIds.clear();
+    Statement stmt;
+    auto rc = PrepareStmt(stmt, db, "SELECT Id, Name FROM ec_Schema");
```
khanaffan: use '[main].[ec_Schema]'. If you reading anything out of db we have schema dispatcher for each attach db. I think SchemaViewWriter should take into account which attach db its querying schema out of.

khanaffan:
I am wondering this can work with older ECDb profile version 4.0.0.1 and onwards. It seems to reference ec_* tables where bunch of them were added in 4.0.0.2.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,869 @@
...
+    auto rc = PrepareStmt(stmt, db, sql.c_str());
+    if (rc != BE_SQLITE_OK)
+        return rc;
+
+    while (stmt.Step() == BE_SQLITE_ROW)
+        {
+        int64_t classId = stmt.GetValueInt64(0);
+        int64_t propertyId = stmt.GetValueInt64(1);
+
+        PropertyDefRecord def;
```
khanaffan, line 259: Does this binary have versioning? Given its binary and we can add new concept in future we should be able to work with older and newer schema versions. 

There is also backend and frontend. Where frontend talk to compatible backend. And so newer backend can send newer binary.

I think your pragma should specify version of binary it needs. 

We must have strong upgrade path here. These things once release with custom binary format is very hard to manage.

khanaffan:
I think this PR need to validate its implementation using compatibility test to ensure we can read schema off older profile version.  We have plenty of these file on hub.

rschili:
@khanaffan 
> I am wondering this can work with older ECDb profile version 4.0.0.1 and onwards. It seems to reference ec_* tables where bunch of them were added in 4.0.0.2.


Thanks for raising this - it sent me down a useful audit. Sharing what I found and what I did about it.

### Audit results

Profile version timeline for the `ec_*` tables that SchemaViewWriter touches:

| Profile | Date | Change relevant to `ec_*` tables |
|---|---|---|
| 4.0.0.1 | Oct 2017 | baseline |
| **4.0.0.2** | **2018** | EC3.2 Units/Formats migration: added `ec_Unit`, `ec_Format`, `ec_Phenomenon`, `ec_UnitSystem`, `ec_FormatCompositeUnit`; added `ec_Schema.OriginalECXmlVersionMajor/Minor`; migrated `ec_KindOfQuantity.PersistenceUnit/PresentationUnits` from FUS → EC3.2 format |
| 4.0.0.3 | Jan 2023 | no-op on `ec_*` tables |
| 4.0.0.4 | Oct 2023 | no-op on `ec_*` tables |
| 4.0.0.5 | Apr 2024 | no-op on `ec_*` tables |

I then cross-checked every `ec_*` table and column SchemaViewWriter queries against the 4.0.0.1 schema. **It does not reference any of the 4.0.0.2-added tables or columns.** The new units/formats schemas are in the writer's excluded list, and the new `OriginalECXmlVersion*` columns are not read.

The only real incompatibility on 4.0.0.1 is that `ec_KindOfQuantity.PersistenceUnit` and `PresentationUnits` still contain legacy FUS-format strings rather than the EC3.2 alias-qualified form (`"u:M"`, `"f:DefaultRealU"`). The writer passes those strings through as-is, so the pragma succeeds and produces a valid blob; only the KoQ unit strings are in the old format.

### What I changed

Rather than reject pre-4.0.0.2 files, I let the pragma continue to work and made the degradation visible:

1. **Compatibility test against the 4.0.0.1 benchmark file** - `ECSqlPragmasTestFixture.schema_view_works_on_old_profile_4001` in `ECSqlPragmasTests.cpp`. Opens `Data/fileformatbenchmark/4001/imodel2.ecdb` readonly (no upgrade), asserts the profile is still 4.0.0.1, runs `PRAGMA schema_view`, and verifies the blob header (`CSCH` magic, format version 1) and schema token. Build passes.
2. **One-time runtime warning** in `PragmaSchemaView::Read` when `ecdb.GetECDbProfileVersion() < ProfileVersion(4, 0, 0, 2)`, naming the FUS-format consequence and recommending a profile upgrade.
3. **Docs updates** in itwinjs-core:
   - `SchemaViewBinaryFormat.md` - new "ECDb Profile Compatibility" section as source of truth.
   - `Pragmas.md` and `SchemaView.md` - short notes linking to that section.
   - TSDoc on `SchemaView.KindOfQuantity.persistenceUnit` and `presentationFormatsRaw`.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
rschili: Yes, the binary has versioning. Frontend always provides the version it needs, and backend always uses the latest version (as it's bundled with the addon).
Currently there is only one version. The SchemaViewWriter will have to support all older versions, so it will grow over time, I hope we won't have to change it too often.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
rschili: Thanks - good catch.

**Done in this PR:** I've gone through `SchemaViewWriter.cpp` and qualified every `ec_*` table reference with `[main].[ec_*]`. That's defensive against any attached db that happens to register a colliding view, and makes the intent explicit.

**On the broader attach-db question:** I dug into how the rest of the ECDb stack handles this today, and I'd like to keep actual attach support out of this PR. Reasoning:

- `PRAGMA checksum(ecdb_schema)` is **already hardcoded to `"main"`** ([ECSqlPragmas.cpp:38](https://github.com/iTwin/imodel-native/blob/main/iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp#L38)). `PRAGMA schema_view` matches that precedent.
- ECDbMeta (`SELECT * FROM meta.ECClassDef`) also only returns classes from main today. The dispatcher routes the class lookup, but the underlying SQL still hits unqualified `ec_*` tables and SQLite resolves those to main.
- The only existing place that does branch on table space is `SchemaManagerDispatcher::IsSubClassOf`, which formats the alias into the SQL when `tableSpace != "main"`. That's the pattern a real fix would need to follow.
- `TableSpaceTests` covers ECSQL data queries / indexes / FKs across attaches, but does not exercise any meta or pragma path across the attach boundary - there's no existing contract to break.

So the gap is real but it's a coherent feature across pragmas + ECDbMeta, not a `schema_view`-specific issue. Doing it well needs: a `PRAGMA schema_view(version, tableSpace)` overload, plumbing `Utf8CP tableSpace` through every `SchemaViewWriter` method, qualifying every prepared statement with the resolved alias, computing `schemaToken` against the attached db rather than main, and a frontend-side decision about which attach to hydrate.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
rschili: I audited the gap and refactored so it now *is* a superset by composition. The function now delegates to `ECN::ECSchema::IsStandardSchema` as the base and adds a small ECDb-specific set on top:

```cpp
static bool IsExcludedSchema(Utf8CP name)
    {
    if (ECN::ECSchema::IsStandardSchema(Utf8String(name)))
        return true;
    static const std::set<std::string, CiLess> s_additional = {
        "BisCustomAttributes",
        "ECDbFileInfo",
        "ECDbSchemaPolicies",
        "ECDbSystem",
        "ECv3ConversionAttributes",
        "SchemaUpgradeCustomAttributes",
    };
    return s_additional.find(name) != s_additional.end();
    }
```

Before committing to that, I audited whether excluding the IsStandardSchema-only entries (the legacy EC2-era stuff: `Bentley_*`, `Dimension_Schema`, `Unit_Attributes`, `Units_Schema`, `KindOfQuantity_Schema`, `iip_mdb_*`, `rdl_*`, etc.) could break anything:

- All 12 of those schemas are pure-CA or pre-EC3.2 unit/quantity definitions. The class counts confirm it: `Dimension_Schema` (137 CA classes, 0 base-class refs), `Units_Schema` (486 CA classes, 0 base-class refs), `KindOfQuantity_Schema` (281 CA classes, 102 base-class refs but all intra-schema), `Unit_Attributes` (CA classes + 3 CA-only structs), the rest are 1-class CA schemas.
- Searched all of `bis-schemas` for structural refs (`typeName="alias:X"` or `<BaseClass>alias:X</BaseClass>`) into any of the 12 - **zero hits**. The only refs found were CA-instance usage of `Bentley_Standard_CustomAttributes` from a handful of Plant WIP schemas, and since we don't transport CA instances anyway, excluding the schema is a no-op for those consumers.
- The SchemaView reader is designed to handle unexpected dangling refs gracefully (drops structural-ref properties, skips dangling base classes, logs warnings), so even if some out-of-tree domain schema *did* have a struct ref into one of these, the failure mode is "property silently dropped" - same as for any other excluded schema today.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,870 @@
...
+    PatchU32(stOffsetPos, stOffset);
+    }
+
+//---------------------------------------------------------------------------------------
+// Main orchestration - pre-collects metadata, then writes flat tables top-down.
+//---------------------------------------------------------------------------------------
+DbResult SchemaViewWriter::WriteAllSchemas(DbCR db)
+    {
+    m_output.clear();
+    m_output.reserve(2 * 1024 * 1024);
```
RohitPtnkr1996, line 834: We can get a better/more accurate capacity value if we reserve space after the pre-pass has run.

RohitPtnkr1996 (approved):

RohitPtnkr1996:
There are two places which still reference the old RuntimeSchema name, one in each comment in the SchemaViewWriter header and source.

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
rschili: I don't think there is a good way to do this.
Even with the pre-pass, the values don't occupy a fixed size, so we can only guess using magic numbers.
This is just the initial vector size, reallocating is just a memcopy so the cost of a wrong initial guess should be low.

khanaffan (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.h
```diff
@@ -156,6 +156,22 @@ struct PragmaCheckECSqlWriteValues : PragmaManager::GlobalHandler {
...
+//        PRAGMA schema_view(1)        -- explicitly request format version 1
+// Result: single row with columns: format (string), formatVersion (int), data (binary), schemaToken (string)
+//+===============+===============+===============+===============+===============+======
+struct PragmaSchemaView : PragmaManager::GlobalHandler {
+    static constexpr uint8_t CURRENT_FORMAT_VERSION = 1;
+    PragmaSchemaView():GlobalHandler("schema_view", "schema_view [(version)] - returns all schema metadata as a binary blob. Optional integer argument selects format version (default: latest)."){}
+    ~PragmaSchemaView(){}
+    virtual DbResult Read(PragmaManager::RowSet&, ECDbCR, PragmaVal const&, PragmaManager::OptionsMap const&) override;
+    virtual DbResult Write(PragmaManager::RowSet&, ECDbCR, PragmaVal const&, PragmaManager::OptionsMap const&) override;
+    static std::unique_ptr<PragmaManager::Handler> Create () { return std::make_unique<PragmaSchemaView>(); }
```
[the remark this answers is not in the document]
aruniverse: @copilot fix the documentation, review it, make sure it is sound

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -0,0 +1,870 @@
...
+    str = Safe(str);
+    std::string key(str);
+    auto it = m_stringIndex.find(key);
+    if (it != m_stringIndex.end())
+        return it->second;
+    uint32_t idx = (uint32_t)m_stringTable.size();
+    auto result = m_stringIndex.emplace(std::move(key), idx);
+    m_stringTable.push_back(result.first->first.c_str()); // pointer into map key (stable)
+    return idx;
+    }
```
kring, line 115: Is this function a performance hotspot? It seems like it could be. If it is, here's a nice optimization opportunity: store the strings in `m_stringTable` and reference them in `m_stringIndex`, rather than the other way around. So the two fields become:

```cpp
bvector<std::string> m_stringTable;
std::unordered_map<std::string_view, uint32_t> m_stringIndex; // pointers into m_stringTable strings (stable across rehash)
```

And then write this function as:

```suggestion
uint32_t SchemaViewWriter::Intern(Utf8CP str)
    {
    str = Safe(str);
    auto it = m_stringIndex.find(str);
    if (it != m_stringIndex.end())
        return it->second;
    uint32_t idx = (uint32_t)m_stringTable.size();
    // m_stringTable stores a std::string
    m_stringTable.emplace_back(str);
    // m_stringIndex holds a std::string_view pointing into the string table (stable)
    m_stringIndex.emplace(std::string_view(m_stringTable.back()), idx);
    return idx;
    }
```

The main benefit is we don't need to construct a `std::string` (which involves a heap allocation) for every lookup.

`WriteStringTable` becomes a little simpler and faster, too (because `std::string::size` is O(1), `strlen` is O(n)):

```cpp
void SchemaViewWriter::WriteStringTable(size_t stOffsetPos)
    {
    uint32_t stOffset = (uint32_t)m_output.size();
    PutU32((uint32_t)m_stringTable.size());
    for (const std::string& s : m_stringTable)
        {
        uint32_t len = (uint32_t)s.size();
        PutU32(len);
        if (len)
            m_output.insert(m_output.end(), (Byte const*)s.data(), (Byte const*)s.data() + len);
        }
    PatchU32(stOffsetPos, stOffset);
    }
```

kring (approved):

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
rschili: I like it, thanks! Will apply.

rschili: I took your approach but with a std::deque, it's an odd choice, but like we discussed offline, std::vector would invalidate the string_views when it reallocates during grow, while deque doesn't do that.

kring: I just want to add a note about _why_ it invalidates the string_views, cause this was surprising to me and I learned something!

If std::string always heap-allocated the buffer in which it stores the string, and moved the buffer pointer when the std::string was moved to a new instance, there would be no problem. The std::string_view would refer to the buffer, which would not be invalidated unless the string itself were modified (which we don't do in this case).

Unfortunately, that first assumption is wrong. std::string implementations often have a "small string optimization" where small strings are stored in a small buffer embedded in the std::string instance itself, rather than allocated on the heap. This embedded buffer can't be moved to a new std::string, which means std::string_views to a small string are invalidated when the std::string is moved to a new instance. 🤦 

std::deque works because it, unlike std::vector, doesn't destroy the original std::string instances when it grows.

aruniverse (approved):

PR #1387, opened 2026-04-16
Fix zlib copy of zlib files that were not correctly updated

DanRod1999:
@jffmarker  noticed that because of this we we are accidentally publishing our lib as 1.3.1, which is incorrect. I tried to remove the redundancy but was getting build issues so I just decided to copy the updated files and add a readme for future reference


aruniverse (approved):

DanRod1999:
@mergifyio backport release/5.8.x

iModelCore/libsrc/compress/VendorAPI/zlib/README.txt
```diff
@@ -0,0 +1,6 @@
+When updating zlib package please note tht you will need to update the 2 .h files in this dir with the new ones located in compress/zlib.
+Also you will need to update iModelCore/libsrc/png/vendor/pnglibconf.h to have use the correct version so zlib
```
tcobbs-bentley:
```suggestion
When updating zlib package please note that you will need to update the 2 .h files in this dir with the new ones located in compress/zlib.
Also you will need to update iModelCore/libsrc/png/vendor/pnglibconf.h to use indicate the correct zlib version
```

I'm not sure where you were going with the "so zlib". If my suggestion isn't good, fix the comment however you like.

tcobbs-bentley (approved):

iModelCore/libsrc/compress/VendorAPI/zlib/README.txt
tcobbs-bentley: Side note: if the PNG header isn't updated, it will trigger a compiler error, correct?

DanRod1999: yeah this will fail during build if not updated

PR #1390, opened 2026-04-20
Update openssl 3.3.7

DanRod1999:
Resolve CVEs:
CVE-2026-28389
CVE-2026-28390

Note: `OPENSSL_NO_ENGINE` is now defined in configuration.h file, so files in engine dir are now no longer compiled. We did not need these files to begin with, but there was a mistake in a prior update which caused us to compile them. That configuration setting has been removed and now we can build without them.


aruniverse (approved):

tcobbs-bentley (approved):

iModelCore/libsrc/openssl/BeOpenSSL.mke
```diff
@@ -909,50 +909,6 @@ $(o)encoder_meth$(oext) : $(vendorSrc)crypto/encode_decode/encoder_meth.c ${Mult
...
-$(o)tb_dsa$(oext) : $(vendorSrc)crypto/engine/tb_dsa.c ${MultiCompileDepends}
-
-$(o)tb_eckey$(oext) : $(vendorSrc)crypto/engine/tb_eckey.c ${MultiCompileDepends}
-
-$(o)tb_pkmeth$(oext) : $(vendorSrc)crypto/engine/tb_pkmeth.c ${MultiCompileDepends}
-
-$(o)tb_rand$(oext) : $(vendorSrc)crypto/engine/tb_rand.c ${MultiCompileDepends}
-
-$(o)tb_rsa$(oext) : $(vendorSrc)crypto/engine/tb_rsa.c ${MultiCompileDepends}
-
```
tcobbs-bentley, line 955: Is `OPENSSL_NO_ENGINE` defined somewhere? It seems like it should be, but I can't see where. I thought maybe macros.h.

DanRod1999: I removed some files from our mke in the engine dir. I believe this is due to the configuration settings i used. I think in prior updates i configured in a way that made us compile files we didn't need. I should have been configuring with `perl Configure no-engine no-dso VC-WIN64A` like mentioned in line 6 of the mke file. Note, we used to also include no-asm, but since we use hardware acceleration I believe we need to configure with asm now.

tcobbs-bentley: I see that this is defined in `vendor/include/configuration.h`, and engine support was deprecated in OpenSSL 3.0.

DanRod1999:
@mergifyio backport release/5.8.x

PR #1393, opened 2026-04-23
Remove Jsoncpp continued

DanRod1999:
More removals of jsoncpp using copilot. This PR should fully clean up Geocoord, line styles, and a few other smaller areas.  

Attempted to remove usage out of Units, however, there are some spots where its still problematic to change like here:
https://github.com/iTwin/imodel-native/blob/main/iModelCore/iModelPlatform/DgnCore/DgnUnits.cpp#L475
Wanted to open another PR specifically working on replacing usage in units if possible.


DanRod1999:
Looks like I have some problems with tests still. Going to draft for now, and decide how best to proceed. If issues are related to floating point precision like with Units then may shelf those changes for now.

DanRod1999:
Looks like some GCS related tests might also have floating point percision issues, so I kept those tests the same for now, and add comments to find them later so I (copilot really) know not to try to update later.

khanaffan (approved):

bbastings (approved):

rschili (approved):

markschlosseratbentley (approved):

pmconne:
> some spots where its still problematic to change like here: https://github.com/iTwin/imodel-native/blob/main/iModelCore/iModelPlatform/DgnCore/DgnUnits.cpp#L475 Wanted to open another PR specifically working on replacing usage in units if possible.

@DanRod1999 By default (as of 2 years ago), we now [round each axis of the project extents to the nearest power of ten](https://github.com/iTwin/itwinjs-core/pull/6727) and encode that into the tile Id. The [expandProjectExtents](https://github.com/iTwin/itwinjs-core/blob/8bb22375c40d6b93a13096f337afb61385747f51/core/frontend/src/tile/TileAdmin.ts#L1105) flag that controls this behavior is entirely `@internal`. So, if you remove the flag and always do the rounding, you should be able to eliminate this dependency on jsoncpp. The main annoyance would be updating lots of existing tests that were written to expect the old encoding before the rounding was introduced.

PR #1399, opened 2026-04-25
Refactor ConcurrentQueryManager: SecondaryConnection abstraction and connection pool improvements

khanaffan:
## Summary

Refactors the ConcurrentQueryManager's connection management layer with several improvements:

### SecondaryConnection Abstraction
- New `SecondaryConnection` virtual interface and `TypedSecondaryConnection<T>` template in `ECDb.h`
- `ECDb::_CreateSecondaryConnection()` virtual method allows DgnDb to override and create `TypedSecondaryConnection<DgnDb>`, registering domain SQL functions through its normal open path
- **Eliminates the regex-based SQL function copying hack** (`GetPrimaryDbSqlFunctions` / `UpdateSqlFunctions`)

### Connection Pool Rewrite
- Replaces fragile `shared_ptr::use_count() == 1` availability check with an explicit `m_inUse` flag guarded by mutex
- Adds `condition_variable`-based wait/notify (`ReleaseConnection` / `GetConnection`) instead of spin-yield loops
- `StopWaiting()` enables clean shutdown with no stuck worker threads
- `Interrupt()` now waits via condition variable instead of busy-looping

### Shutdown Safety
- `ConcurrentQueryMgr::Shutdown()` is now called on db close for both `NativeECDb` and `NativeDgnDb`, preventing use-after-close
- `Shutdown()` gracefully returns instead of throwing when the db is already closed

### RAII & Cleanup
- `QueryResponse::Future`: raw `Impl*` → `unique_ptr<Impl>` with defaulted move operations
- Deduplicated response marshalling in `JsInterop.cpp` into a shared `marshalResponse` lambda
- Moved N-API argument parsing into `JsInterop` overloads for cleaner separation
- Fixed `Config::GetFromEnv()` logic error (`||` → `&&`)
- Removed double semicolons and duplicate forward declaration


iModelCore/ECDb/PublicAPI/ECDb/ECDb.h
```diff
@@ -154,6 +154,32 @@ struct AsciiCaseInsensitiveCompare {
...
+    virtual ECDb const& GetDb() const = 0;
+};
+
+//=======================================================================================
+//! Template wrapper that holds a secondary connection as a value member of the specified type.
+//! For ECDb: TypedSecondaryConnection<ECDb>.
+//! For DgnDb: TypedSecondaryConnection<DgnDb> — the DgnDb open path registers all domain SQL functions.
+// @bsiclass
+//+===============+===============+===============+===============+===============+======
+template<typename T>
```
soham-bentley, line 176: Shouldn't we have something like this?
```C++
template <typename T>
concept IsECDb = std::is_base_of_v<ECDb, T>;
template <IsECDb T>
```

iModelCore/ECDb/PublicAPI/ECDb/ECDb.h
```diff
@@ -154,6 +154,32 @@ struct AsciiCaseInsensitiveCompare {
...
+struct TypedSecondaryConnection : SecondaryConnection {
+    T m_db;
```
soham-bentley, line 178: I maybe missing something. But is there any reason we are not holding `T&` reference here ?
And then while creating a secondary connection we can have a constructor like the following:
```C++
TypedSecondaryConnection(const T& db): m_db(db) {}
```
Then in `_CreateSecondaryConnection` we can do something like
```C++
auto conn = std::make_unique<TypedSecondaryConnection<DgnDb>>(this);
```

soham-bentley: Okay got it....we fill in newConnection from `BeSQLite.cpp`

khanaffan:
This is not require right now. And if redone will probably do it differently than this my moving out concurrent query from ECDB in to iModelPlatform or add-on.

PR #1400, opened 2026-04-26
Add InstanceGraph API and relations() virtual table for fast relationship traversal

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9647

## Summary

Adds the experimental `InstanceGraph` C++ API and `ECVLib.Relations()` ECSQL table-valued function for relationship traversal. Traversal SQL is generated directly from class/property maps rather than preparing an ECSQL query for each relationship. The outer `Relations()` query still goes through ECSQL preparation.

`InstanceGraph` materializes a graph with depth-limited BFS and set operations. `Relations()` streams one-hop results. Both share an ECDb-level cache of relationship discovery and generated SQL; prepared statements use the existing SQLite statement cache.

**Experimental:** `Relations()` is disabled by default. Enable it with `PRAGMA experimental_features_enabled=true` or append `ECSQLOPTIONS ENABLE_EXPERIMENTAL_FEATURES` to a query. The C++ API is also experimental, but does not require that ECSQL option.

Multiple virtual-table modules can contribute classes to the same virtual schema. `Relations` shares `ECVLib` with `IdSet`, and can be called as either `ECVLib.Relations(...)` or `Relations(...)`.

ECSQL version: `2.0.4.0` -> `2.0.4.1`.

## Relations API

```text
Relations(ECInstanceId, ECClassId [, direction])
```

Both IDs are required. Direction is case-insensitive: `'forward'` (source to target), `'backward'` (target to source), or `'both'` (default, also used for `NULL`). Other direction values are rejected.

### Virtual-table schema

| Column | Type | Role | Description |
| --- | --- | --- | --- |
| `RelatedECInstanceId` | `long (Id)` | Output | Related entity instance ID |
| `RelatedECClassId` | `long (Id)` | Output | Related entity class ID |
| `Direction` | `string` | Output | `'forward'` or `'backward'` for this edge |
| `RelationshipECClassId` | `long (Id)` | Output | Relationship class ID |
| `RelationshipECInstanceId` | `long (Id)` | Output | Relationship instance ID; for navigation relationships, the ID of the row holding the navigation property |
| `NavPropertyName` | `string` | Output | Navigation property name; `NULL` for link-table relationships |
| `ECInstanceId` | Integer ID | Hidden input | Required first argument: seed instance ID |
| `ECClassId` | Integer ID | Hidden input | Required second argument: seed class ID |
| `TraversalDirection` | Text | Hidden input | Optional third argument: `'forward'`, `'backward'`, or `'both'` |

Output types are defined by the ECSQL virtual class. Hidden input types describe the values consumed by the SQLite module; these inputs are not ECSQL properties.

Distinct relationship instances between the same entities are preserved. A self-loop can produce both a forward and a backward row.

## ECSQL examples

Enable the feature before running these examples. Replace `0x1A` and `0x38` with an instance ID and its class ID from the database; the `bis.Element` examples require a BIS schema.

```sql
PRAGMA experimental_features_enabled=true;
```

### One hop, including relationship identity and navigation property

```sql
SELECT RelatedECInstanceId, RelatedECClassId, Direction,
       RelationshipECClassId, RelationshipECInstanceId, NavPropertyName
FROM Relations(0x1A, 0x38);
```

### Directional traversal

```sql
SELECT RelatedECInstanceId FROM Relations(0x1A, 0x38, 'forward');
SELECT RelatedECInstanceId FROM Relations(0x1A, 0x38, 'backward');
```

### Correlated traversal from elements

```sql
SELECT r.RelatedECInstanceId, r.RelatedECClassId, r.Direction
FROM bis.Element a,
     Relations(a.ECInstanceId, a.ECClassId) r
WHERE a.CodeValue = 'P1';
```

### Two forward hops

```sql
SELECT r1.RelatedECInstanceId, r1.RelatedECClassId,
       r2.RelatedECInstanceId, r2.RelatedECClassId
FROM bis.Element a,
     Relations(a.ECInstanceId, a.ECClassId, 'forward') r1,
     Relations(r1.RelatedECInstanceId, r1.RelatedECClassId, 'forward') r2
WHERE a.CodeValue = 'P1';
```

### Recursive walk bounded to three hops

```sql
WITH RECURSIVE graph(ECInstanceId, ECClassId, Depth) AS (
    SELECT 0x1A, 0x38, 0
    UNION
    SELECT r.RelatedECInstanceId, r.RelatedECClassId, g.Depth + 1
    FROM graph g
    CROSS JOIN Relations(g.ECInstanceId, g.ECClassId) r
    WHERE g.Depth < 3
)
SELECT ECInstanceId, ECClassId, Depth FROM graph ORDER BY Depth;
```

This is a bounded walk, not the C++ graph's visited-node BFS: `UNION` compares the entire row, including `Depth`, so a node can recur at different depths. The depth condition bounds cycles.

## C++ usage

Within a function returning `BentleyStatus`, with `ecdb`, `classId` and `instanceId` supplied:

```cpp
InstanceGraph graph(ecdb);
ECInstanceKey seedKey(classId, instanceId);
graph.AddSeed(seedKey);

if (SUCCESS != graph.ExpandAll(3))
    return ERROR;

auto const* related = graph.GetRelated(seedKey);
```

The `ECInstanceKey` constructor takes **class ID first**, unlike the SQL function. `ExpandNode(key, direction)` expands one node and replaces its cached edges on re-expansion. `ExpandAll(maxDepth)` traverses from added seeds; zero performs no expansion, and the default `UINT8_MAX` is a limit of 255 hops, not unlimited traversal.

`InstanceGraph::Overlaps`, `Intersection` and `Union` compare instance keys and require both graphs to belong to the same ECDb. Set-operation results have no traversal seeds; add a seed before calling `ExpandAll()` on them.

Only the primary (`main`) table space is supported. Do not traverse concurrently with schema import or `ClearECDbCache()`. Clearing the ECDb cache invalidates shared discovery/SQL caches, not edges already materialized in an `InstanceGraph`.

## Review follow-up

This PR also includes:

- Track successful expansion direction so `ExpandAll()` completes a previous one-direction expansion; publish replacement edges only after successful traversal.
- Make virtual-schema lookup/merging case-insensitive and remove the unused `ApplicableRelationship::m_mapStrategy`.
- Add directional, set-operation, schema-name-case and recursive-walk regression coverage.
- Add opt-in `InstanceGraphTests.DISABLED_CompareECInstanceFinder`: equivalent one-hop entity results for 128-neighbor link-table and navigation-property stars, with separate first-traversal-after-reopen and warm-cache timings.

The benchmark has not been run; no quantitative speedup is claimed. Its cold case does not claim a cold OS/file cache. It can be selected with:

```text
--gtest_also_run_disabled_tests --gtest_filter=InstanceGraphTests.DISABLED_CompareECInstanceFinder
```


iModelCore/ECDb/ECDb/InstanceGraph.cpp
```diff
@@ -0,0 +1,1201 @@
...
+BentleyStatus InstanceGraph::ExpandAll(uint8_t maxDepth)
+    {
+    if (maxDepth == 0)
+        {
+        // Seed only: the seeds are part of the graph and have no (known) edges.
+        for (auto const& seed : m_seeds)
+            m_adjacency.emplace(seed, bvector<RelatedInstance>());
+
+        return SUCCESS;
+        }
```
RohitPtnkr1996: This special case will mark all seeds as expanded for future calls.
Every subsequent ExpandAll(x) call after ExpandAll(0) will end up with an empty graph right?

khanaffan: Fixed in 013f1f86cf. `ExpandAll(0)` now leaves seeds unexpanded, and the test verifies that a later expansion traverses normally.

iModelCore/ECDb/ECDb/InstanceGraph.cpp
```diff
@@ -0,0 +1,1192 @@
...
+        m_visited.insert(rel.GetKey());
+        }
+
+    return SUCCESS;
+    }
+
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+BentleyStatus InstanceGraph::ExpandAll(uint8_t maxDepth)
```
rschili, line 1047: `ExpandAll()` mistakes any existing adjacency entry for a complete bidirectional expansion. After `ExpandNode(seed, TraversalDirection::Forward)`, a later `ExpandAll()` skips that seed and never discovers its backward edges, so the supposedly full graph depends on call order; track expansion state per direction or re-expand nodes whose cached direction is narrower than `Both`.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -81,6 +81,45 @@ BentleyStatus VirtualSchemaManager::AddAndValidateVirtualSchema(Utf8StringCR sch
             return ERROR;
         }
     }
+
+    // If a schema with the same name already exists, merge the classes into it. This allows
+    // several virtual table modules to contribute classes to one shared virtual schema.
+    auto existingIt = m_schemas.find(schema->GetName());
```
rschili, line 93: Schema names are case-insensitive, but this seems a case-sensitive lookup

iModelCore/ECDb/Tests/NonPublished/InstanceGraphTests.cpp
```diff
@@ -0,0 +1,2644 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the repository root for full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+#include "ECDbPublishedTests.h"
```
rschili, line 5: The PR introduces substantial SQL-generation and caching machinery specifically for faster traversal, and its history claims a 3–10× improvement, but it adds no comparison benchmark against `ECInstanceFinder`. A focused benchmark would document the workload for which this implementation is worthwhile and catch regressions in discovery/cache behavior.

iModelCore/ECDb/ECDb/InstanceGraphImpl.h
```diff
@@ -0,0 +1,219 @@
...
+//=======================================================================================
+//! Describes a relationship that can be traversed from a given entity class
+// @bsiclass
+//+===============+===============+===============+===============+===============+======
+struct ApplicableRelationship final
+    {
+    ECN::ECRelationshipClassCP  m_relClass;         //!< Base relationship class
+    ECN::ECRelationshipEnd      m_thisEnd;           //!< Which end "our" class sits on (Source or Target)
+    ClassMap::Type              m_mapType;            //!< LinkTable or EndTable
+    MapStrategy                 m_mapStrategy;        //!< For EndTable: ForeignKeyRelationshipInSourceTable or TargetTable
```
rschili: `ApplicableRelationship::m_mapStrategy` is initialized but never read.

rschili (approved):

iModelCore/ECDb/ECDb/InstanceGraphImpl.h
khanaffan: Removed the unused member and constructor argument in 08e3c4d53a; kept the NotMapped filter.

iModelCore/ECDb/ECDb/InstanceGraph.cpp
khanaffan: Fixed in 08e3c4d53a. ExpandAll now completes directional expansions and only reuses successful bidirectional expansions.

iModelCore/ECDb/Tests/NonPublished/InstanceGraphTests.cpp
khanaffan: Added an opt-in ECInstanceFinder comparison benchmark with cold/warm timings in 08e3c4d53a. Not run yet; no quantitative speedup claimed.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
khanaffan: Fixed in 08e3c4d53a. The schema map now uses CompareIUtf8Ascii for case-insensitive lookup and merging.

PR #1405, opened 2026-04-27
Update TypeScript dev dependencies and clean up redundant types in NativeLibrary.ts

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9663

Housekeeping for the `iModelJsNodeAddon/api_package/ts` package: **dependency updates** and **`NativeLibrary.ts` type cleanup**.

Fixes iTwin/itwinjs-backlog#2334

### 1. Dependency updates

13 dev dependency updates to the latest release **within their currently adopted major** (no major jumps, to keep risk low). `engines.node` → `^20.11.0 || ^22.11.0 || ^24.11.0`.

`@itwin/core-*` and `@itwin/build-tools` are intentionally untouched — managed by the `sync-itwinjs-core-deps` workflow.

<details><summary>Version table, and what was deliberately skipped</summary>

| Package | Before | After |
| --- | --- | --- |
| `@itwin/eslint-plugin` | `^5.0.0-dev.2` | `^6.1.0` |
| `@types/chai` | `^4.3.6` | `^4.3.20` |
| `@types/chai-as-promised` | `^7.1.6` | `^7.1.8` |
| `@types/fs-extra` | `^4.0.13` | `^8.1.5` |
| `@types/node` | `~20.17.0` | `~20.17.58` |
| `@types/sinon` | `^9.0.11` | `^17.0.4` |
| `chai` | `^4.3.10` | `^4.5.0` |
| `chai-as-promised` | `^7.1.1` | `^7.1.2` |
| `eslint` | `~9.14.0` | `^9.39.5` |
| `mocha` | `^11.7.5` | `^11.8.0` |
| `sinon` | `^9.2.4` | `^17.0.2` |
| `typescript` | `~5.6.2` | `~5.6.3` |

Skipped — each needs a major jump deserving its own PR: `chai` 4→6 (ESM-only), `eslint` 9→10, `typescript` 5.6→7, `fs-extra` 8→11, `rimraf` 2→6, `cpx2` 3→9, `@types/node` 20→26, `@types/sinon` 17→22, `@types/chai` 4→5.

One exception to the "no major jumps" rule: `@types/fs-extra` moves 4→8, because it was describing a library four majors newer than itself (`fs-extra` is `^8.1.0`). This aligns the typings with what is actually installed rather than adopting anything new.

</details>

### 2. Removed redundant type duplicates

Local declarations replaced by their `@itwin/core-common` equivalents, re-exported as aliases so the public `IModelJsNative.*` names keep working: `TxnProps`, `PerStatementHealthStats`, `FontEncodingProps` → `RscFontEncodingProps`, `ModelExtentsResponseProps` → `ModelExtentsProps`, `FontType`.

Also:
- **Dropped deprecated `IDisposable`** from 8 classes (`ECDb`, `ECSqlStatement`, `SQLiteDb`, …). Deprecated in core-bentley 5.0 with a removal date now passed; every class still declares `dispose()`, so consumers are unaffected.
- **Fixed `perStatementStats`**: was a 1-element tuple `[PerStatementHealthStats]`, now `PerStatementHealthStats[]`, matching what native returns.

> `FontType` goes from ambient `enum` to type alias. Safe: the addon never exported it as a runtime value (no `"FontType"` in any `.cpp`), and the values are byte-identical.

### 3. Replaced `any` with real types

`createBRepGeometry`, `elementGeometryCacheOperation`, `processGeometryStream`, `updateElementGeometryCache` and `doSnap` now use their actual prop types. These were derived from core-backend's **call sites** in `IModelDb.ts`, not the inline comments — e.g. `processGeometryStream` claimed `ElementGeometryOptions` but core passes `ElementGeometryRequest`.

Also added two fields native emits but the typings omitted: `ChangesetHealthStats.changesetIndex` and `ChangesetFileProps.uncompressedSize`.

`ExportGraphicsOptions`, `ExportPartGraphicsOptions`, `SimplifyElementGeometryArgs` and `IModelDb` **must stay `any`** — they live in core-backend, which can't be a dependency here (it depends on this addon). Their comments were corrected to the real names.

### 4. Flagged members not used by itwinjs-core

43 members annotated `@not-used-by-itwinjs-core` (comment-only; no signatures changed), from scanning every `.ts`/`.tsx` in itwinjs-core against all 541 declared members. Three groups: no caller at all, addon-test-only, and re-declared in core-backend but never invoked.

⚠️ **Informational, not removal candidates** — imodel-transformer, presentation and internal tooling also bind to this surface.

### 5. Removed dead APIs

- **`DgnDb.inBulkOperation()`** — declared in the public addon typings since the initial open-source commit, but never registered in N-API, so calling it threw `TypeError: dgndb.inBulkOperation is not a function`. Removed the declaration. Fixes iTwin/itwinjs-backlog#2334.
- **`RevisionUtility`** — unused debugging/testing helper. Removed the TypeScript class and the `NativeRevisionUtility` N-API binding. The underlying `BeSQLite::RevisionUtility` C++ API is unchanged.

### 6. Test files

`DgnDb.test.ts` / `ECDb.test.ts` only drop `as any` / `as DbQueryRequest` assertions that became unnecessary once the types came from core-common — required, since `no-unnecessary-type-assertion` otherwise fails lint.

The previously included test suites (`DgnDbCrud.test.ts`, `ECDb.standalone.test.ts`, `SqliteDb.test.ts`) and the `makeTs.mke` change were **removed to keep this PR focused**; they can be revisited separately.

### Review feedback

All addressed: `TxnProps` exported, `PerStatementHealthStats` aliased rather than deleted, `perStatementStats` tuple→array, `GeometryOutputFormat` spacing.

On `TxnType` gaining `"Schema"` — that's a **correctness fix**, not a regression: `TxnManager.cpp` already emits `"Schema"` when the txn type column is NULL, which the old union (`"Data" | "ECSchema" | "Ddl"`) failed to declare. No exhaustive switch over this type exists in the repo.

### Verification

`npm run buildAndLint` passes cleanly (0 problems); affected core-backend call sites type-checked against the installed `@itwin/core-common`.


iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -560,19 +550,9 @@ export declare namespace IModelJsNative {
...
-    sessionId: number;
-    nextId?: TxnIdString;
-    prevId?: TxnIdString;
-    props: { description?: string; source?: string, appData: { [key: string]: any } };
-    type: "Data" | "ECSchema" | "Ddl";
-    reversed: boolean;
-    grouped: boolean;
-    timestamp: string; // ISO 8601 format
-  }
+  /** @see `TxnProps` from `@itwin/core-common` */
```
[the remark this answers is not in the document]
khanaffan: @copilot apply changes based on [this feedback](https://github.com/iTwin/imodel-native/pull/1405#discussion_r3149892412)

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -402,13 +402,7 @@ export declare namespace IModelJsNative {
...
 
-  interface PerStatementHealthStats {
-    sqlStatement: string;
-    dbOperation: string;
-    rowCount: number;
-    elapsedMs: number;
-    fullTableScans: number;
-  }
+  /** @see `PerStatementHealthStats` from `@itwin/core-common` */
   interface ChangesetHealthStats {
```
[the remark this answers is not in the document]
khanaffan: @copilot apply changes based on [this feedback](https://github.com/iTwin/imodel-native/pull/1405#discussion_r3149892399)

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -402,13 +402,7 @@ export declare namespace IModelJsNative {
...
     changesetId: string;
     uncompressedSizeBytes: number;
```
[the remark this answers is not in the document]
khanaffan: @copilot apply changes based on [this feedback](https://github.com/iTwin/imodel-native/pull/1405#discussion_r3149892379)

iModelJsNodeAddon/api_package/ts/src/test/ChangesetReader.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        category: categoryId,
+        code: Code.createEmpty(),
+      } as any);
+
+      const txnId = db.getCurrentTxnId();
+      db.saveChanges("insert for txn reader");
+
+      const reader = new iModelJsNative.ChangesetReader();
+      reader.openTxn(db, txnId, false, 0);
+
```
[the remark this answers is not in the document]
khanaffan: @copilot apply changes based on [this feedback](https://github.com/iTwin/imodel-native/pull/1405#discussion_r3149892312)

khanaffan:
@copilot Note undo all changes realted to lzma savetxn and read txn

khanaffan:

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -560,18 +552,8 @@ export declare namespace IModelJsNative {
...
-    prevId?: TxnIdString;
-    props: { description?: string; source?: string, appData: { [key: string]: any } };
-    type: "Data" | "ECSchema" | "Ddl";
-    reversed: boolean;
-    grouped: boolean;
-    timestamp: string; // ISO 8601 format
-  }
-
+  /** @see `TxnProps` from `@itwin/core-common` */
+  type TxnProps = CoreTxnProps;
```
aruniverse: **Two issues on this line:**

**1. Missing `export` (breaking API change)**  
`main` declares `export interface TxnProps`; this PR drops to `type TxnProps = CoreTxnProps` without `export`.  
Inside a `declare namespace`, un-exported declarations are not accessible as `IModelJsNative.TxnProps`.  
No direct usages were found in this repo, but it is a public API removal.

```ts
// fix:
export type TxnProps = CoreTxnProps;
```

**2. `TxnType` silently gains `"Schema"` (exhaustive-check risk)**  
The removed local type had `type: "Data" | "ECSchema" | "Ddl"` (3 values).  
`CoreTxnProps.type` is `TxnType = "Data" | "ECSchema" | "Schema" | "Ddl"` (4 values).  
Any downstream switch or if/else over the original 3-value union silently falls through on `"Schema"` transactions.

Options:
- Narrow the alias until downstream is migrated: `export type TxnProps = Omit<CoreTxnProps, "type"> & { type: "Data" | "ECSchema" | "Ddl" };`
- Or document the new value in the PR description / CHANGELOG so callers can handle it.

aruniverse (commented):
## Adversarial review — PR #1405 (`affank/housekeeping-cleanup`)

Three reviewer angles: **type-safety**, **test quality**, **dependency compatibility**.  
See inline comments for specifics. Summary below.

---

### Fixes worth doing now

| # | File | Issue |
|---|------|-------|
| 1 | `NativeLibrary.ts:556` | `export` keyword dropped from `TxnProps` — breaks `IModelJsNative.TxnProps` as a qualified name |
| 2 | `NativeLibrary.ts:556` | `CoreTxnProps.type` widens to `"Data" | "ECSchema" | "Schema" | "Ddl"` — exhaustive switch handlers silently miss `"Schema"` |
| 3 | `DgnDbCrud.test.ts:268` | `assert.fail()` inside `catch` swallows the failure — test can't detect a missing throw |
| 4 | `DgnDbCrud.test.ts:193,212,229` | Shared-DB `abandonChanges()` with no `try/finally` — assertion failure leaves state dirty for subsequent tests |

### Optional improvements

- `TxnProps` and `RscFontEncodingProps` are `@alpha` in core-common — add a re-verify comment to the upgrade runbook
- `ChangesetHealthStats.perStatementStats` is still typed as a tuple `[PerStatementHealthStats]` in core-common itself; file an upstream fix
- ECDb.standalone guid test binds a GUID against an `IGeometry` property — should use a proper `Guid`-typed property
- Async callback in ECDb.standalone leaks `ins` statement if an assertion throws before `ins.dispose()`
- `@types/node ~20.17.0` lags behind the new `engines.node ^24.11.0`; no CI matrix entry for Node 24

### Not an issue

- `@types/mocha 5→10` / `useColors` — **already fixed** in the PR (`color: true` in `index.ts`) ✓
- sinon 9→17: all APIs in use (`stub`, `restore`, `callsFake`, `withArgs`, `match`) are present in v17 ✓
- `PerStatementHealthStats`, `FontEncodingProps`, `ModelExtentsProps` aliases — exact field-by-field matches with core-common ✓
- `perStatementStats: PerStatementHealthStats[]` (array) is semantically correct; the old `[PerStatementHealthStats]` (tuple) was a typo ✓

iModelJsNodeAddon/api_package/ts/src/test/DgnDbCrud.test.ts
```diff
@@ -0,0 +1,460 @@
...
+        modeledElement: { id: partitionId },
+        isPrivate: false,
+      };
+      const modelId = db.insertModel(modelProps);
+      db.saveChanges();
+
+      // Empty model should throw NoGeometry
+      try {
+        db.queryModelExtents({ id: modelId });
+        assert.fail("should have thrown");
```
aruniverse: **`assert.fail()` is swallowed by the surrounding `catch` — test can't detect a missing throw.**

If `queryModelExtents` does NOT throw, `assert.fail("should have thrown")` raises a Chai `AssertionError`. That `AssertionError` is caught here. `AssertionError` has no `errorNumber`, so `err.errorNumber` is `undefined`, and the test fails with a confusing `"expected undefined to equal 65570"` instead of `"should have thrown"`.

Replace the try/catch pattern with a flag:
```ts
let caught: any;
try {
  db.queryModelExtents({ id: modelId });
} catch (err: any) {
  caught = err;
}
assert.isDefined(caught, "queryModelExtents should have thrown for empty model");
assert.equal(caught.errorNumber, IModelStatus.NoGeometry);
```

iModelJsNodeAddon/api_package/ts/src/test/DgnDbCrud.test.ts
```diff
@@ -0,0 +1,460 @@
...
+
+      const elementId = db.insertElement(elemProps);
+      assert.isString(elementId);
+      assert.isTrue(elementId.length > 0);
+
+      const retrieved = db.getElement({ id: elementId });
+      assert.equal(retrieved.classFullName, "CrudTest:TestPhysical");
+      assert.equal(retrieved.model, physicalModelId);
+
+      db.abandonChanges();
```
aruniverse: **Shared-DB isolation: `abandonChanges()` is not protected by `try/finally`.**

This pattern repeats on lines 193, 212, and 229 (the three element-operation tests that share `db` via `before()`/`after()`).  
If any `assert.*` throws before reaching `abandonChanges()`, the inserted/updated elements remain in the open transaction. The next test in the describe block then operates on a dirty database — inserts accumulate, IDs shift, and failures produce misleading messages.

Wrap each test body:
```ts
it("insert element and retrieve it", () => {
  try {
    const elementId = db.insertElement(elemProps);
    // ... assertions ...
  } finally {
    db.abandonChanges();
  }
});
```
Same fix needed for "update element properties" (line 212) and "delete element" (line 229).

iModelJsNodeAddon/api_package/ts/src/test/ECDb.standalone.test.ts
```diff
@@ -0,0 +1,749 @@
...
+      const sel = new iModelJsNative.ECSqlStatement();
+      sel.prepare(db, "SELECT Bin FROM test.AllTypes WHERE ECInstanceId=?");
+      sel.getBinder(1).bindId(id);
+      assert.equal(sel.step(), DbResult.BE_SQLITE_ROW);
+      const retrieved = sel.getValue(0).getBlob();
+      assert.deepEqual(new Uint8Array(retrieved), blob);
+      sel.dispose();
+    });
+
+    it("bind and retrieve guid values", () => {
```
aruniverse, line 294: **GUID test uses an `IGeometry`-typed property — type mismatch may silently pass for wrong reasons.**

The `AllTypes` schema (defined earlier in this file) declares `G` as:
```xml
<ECProperty propertyName="G" typeName="Bentley.Geometry.Common.IGeometry"/>
```
But this test binds a GUID string via `stmt.getBinder(1).bindGuid(guid)` and reads it back with `sel.getValue(0).getGuid()`.  
If the native layer silently stores the GUID bytes as a blob and `getGuid()` reinterprets them, the test may pass while exercising unintended behavior. If it coerces or rejects, it may fail spuriously.

Suggested fix: add a proper `Guid`-typed property to `AllTypes`:
```xml
<ECProperty propertyName="Gu" typeName="Bentley.Common.Guid"/>
```
Then bind/read against `Gu` instead of `G`.

iModelJsNodeAddon/api_package/ts/src/test/ECDb.standalone.test.ts
```diff
@@ -0,0 +1,749 @@
...
+    });
+
+    after(() => {
+      db.dispose();
+    });
+
+    it("stepForInsertAsync returns correct result", (done) => {
+      const ins = new iModelJsNative.ECSqlStatement();
+      ins.prepare(db, "INSERT INTO test.AllTypes(S) VALUES('async-test')");
+      ins.stepForInsertAsync((r) => {
```
aruniverse, line 526: **Statement leak if an assertion throws inside the async callback.**

If `assert.equal(r.status, ...)` or `assert.isString(r.id)` throws, `ins.dispose()` is never reached and `done()` is never called. Mocha eventually reports a timeout rather than the real assertion failure, making the failure hard to diagnose.

Wrap the callback body:
```ts
ins.stepForInsertAsync((r) => {
  try {
    assert.equal(r.status, DbResult.BE_SQLITE_DONE);
    assert.isString(r.id);
    const insertedId = r.id;
    ins.dispose();
    // ... sel setup and stepAsync ...
    done();
  } catch (err) {
    ins.dispose();
    done(err);
  }
});
```

khanaffan:
@copilot in MD images not loading

aruniverse (approved):

khanaffan:
Re-checked all 12 review threads against the current head — all are resolved and genuinely addressed:

| Thread | Status |
| --- | --- |
| `TxnProps` missing `export` (@aruniverse, Copilot) | Fixed — `export type TxnProps = CoreTxnProps;`, verified `IModelJsNative.TxnProps` still resolves as a qualified name |
| `PerStatementHealthStats` removed without replacement | Fixed — re-exported as an alias |
| `perStatementStats` typed as a tuple | Fixed — now `PerStatementHealthStats[]` |
| `GeometryOutputFormat=` spacing | Fixed |
| `DgnDb.test.ts` missing-`value` cast | Not an issue — `CodeProps.value` is optional (`value?: string`), so the object compiles without a cast; `buildAndLint` passes |
| 7 threads on `Transactions.test.ts`, `ChangesetReader.test.ts`, `DgnDbCrud.test.ts`, `ECDb.standalone.test.ts` | Moot — those files are no longer part of this PR |

On the `TxnType` widening concern: including `"Schema"` is a correctness fix rather than a regression. `TxnManager.cpp` already emits `"Schema"` when the txn type column is NULL, which the old local union (`"Data" | "ECSchema" | "Ddl"`) failed to declare. No exhaustive switch over this type exists in the repo.

**Heads-up @aruniverse** — one commit landed after your approval, so the PR flipped back to `REVIEW_REQUIRED`:

- `9434a10` tags 43 members with `@not-used-by-itwinjs-core` (comments only, no signature changes), based on a scan of every `.ts`/`.tsx` in itwinjs-core. Split into 3 groups: no caller at all, addon-test-only, and re-declared in core-backend but never called (deliberate passthroughs, *not* removal candidates).

That scan turned up one real defect worth tracking separately: `DgnDb.inBulkOperation()` is declared here but has **no N-API registration**, so calling it throws at runtime — `IsBulkOperation` exists in DgnPlatform but was never bridged. I left it declared and annotated rather than changing behaviour in this PR.

Also worth an upstream fix: `ChangesetHealthStats.perStatementStats` in `@itwin/core-common` still has the `[PerStatementHealthStats]` tuple typo that this PR corrects locally.

khanaffan:
@pmconne can you review this PR.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -732,6 +722,9 @@ export declare namespace IModelJsNative {
     public importSchemasDuringSemanticRebase(schemaFileNames: string[], options?: SchemaImportOptions): void;
     public importSchemas(schemaFileNames: string[], options?: SchemaImportOptions): DbResult;
     public importXmlSchemas(serializedXmlSchemas: string[], options?: SchemaImportOptions): DbResult;
+    /** @not-used-by-itwinjs-core No native implementation is registered for this method, so calling it fails at
```
pmconne: I thought I saw a PR that fixed this, but I can't find it now. Did I hallucinate it? If no one's using it, and it throws at run-time, why not just delete it?
EDIT: it's not obvious from snippet above - I'm referring to `inBulkOperation`.

pmconne (approved):

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
pmconne: nvm, found the issue filed for this. Could save a bunch of time+effort by just deleting it in this PR.

khanaffan: I removed it in this PR along with another old class call RevisionUtility which I use to use to debug changeset but its not longer used as we have now first class concept of sqlitechangeset reader in the itwinjs.

PR #1412, opened 2026-05-05
update itwin packages

DanRod1999:
update itwin packages in dev deps to 5.10 && npm audit

Reason for updating is because we are currently on 5.0.0-dev version in the native code and that creates conflicting types with itwinjs-core latest version, so users that try to link in Windows will get compilation errors. 

Note: users on mac and linux do not see these problems due to how the packages are linked, and CI did not catch this issue because its links differently than how users typically link native to core locally using our scipts


aruniverse (approved):

bbastings (approved):

khanaffan (approved):

markschlosseratbentley (approved):

iModelJsNodeAddon/api_package/ts/src/test/ECDb.test.ts
```diff
@@ -1,7 +1,8 @@
 import { expect } from "chai";
 import * as os from "os";
+import { DbResult } from "@itwin/core-bentley";
```
aruniverse, line 3: what happened here?

DanRod1999: DbResult used to be re-exported in core-common, but that is not the case anymore in 5.0. Tbh I don't really understand why we just started getting this error on 5.10 version of itwin and not 5.0.0-dev.x. But this change should've been made a while ago

iModelJsNodeAddon/api_package/ts/package.json
```diff
@@ -17,10 +17,10 @@
     "test": "node ./lib/test/index.js"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^5.0.0-dev",
-    "@itwin/core-bentley": "^5.0.0-dev",
-    "@itwin/core-common": "^5.0.0-dev",
-    "@itwin/core-geometry": "^5.0.0-dev",
+    "@itwin/build-tools": "^5.10.0-dev",
```
hl662, line 20: would we have to do this bump manually every time we switch to a new nightly minor? Can this not be covered by our automated bump script?

aruniverse: agreed, but lets not hold this pr up, create separate issue and tackle independently

hl662 (approved):

iModelJsNodeAddon/api_package/ts/package.json
DanRod1999: This can be automated during our "imodeljsnodeaddon version bump" pipeline I think. When we create the new release branches, we can point main branch to these new dev versions.

DanRod1999: then technically we'd have to update the release branch to no longer be on dev versions as well.

Also this can get messy during major bumps cause they can cause breaking changes. Not sure if there's a good solution for that yet other than manually fixing as the breaks happen, and try to be proactive about it

PR #1418, opened 2026-05-11
WIP Helper function to maintain percision from jsoncpp in RapidJson

DanRod1999:
Don't think this pr is needed as [Paul pointed out here ](https://github.com/iTwin/itwinjs-backlog/issues/444#issuecomment-4415501546)so this may no longer be a problem. Wanted to keep the branch up for reference though just in case anyone is curious, or thinks it may be of value for other areas. some GCS tests also fail, and it may be due to precision too.


DanRod1999:
Closing as no longer needed. Please see linked comment in description for reference

PR #1421, opened 2026-05-12
Bulk element deletion API fixes

RohitPtnkr1996:
Issue:
The bulk element delete API fails with a BE_SQLITE_LOCKED error when called a second time.

Investigation:
The API uses a "DROP TABLE IF EXISTS..." DDL to re-create the temp.ElementsToDelete at the start.
This ddl needs to update the temp.sqlite_master table when the temp table is dropped which requires an exclusive lock.
If during this, an unrelated select statement is referencing an unrelated temp table, it will hold a shared lock.
The first call to deleteElements succeeds since the temp table does not yet exist.
But subsequent calls now throw the locked error.

Fix: 
Removed the redundant "drop table if exists" ddl entirely.
We are already doing a "create table if not exists - delete from" with the temp table.

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9285


iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -487,10 +487,7 @@ DgnElementId ElementAspectIteratorEntry::GetElementId() const {return m_statemen
 DbResult BulkElementDeletion::CreateTempTables() const
     {
     const auto tempTableCreationStatements = {
-        // Drop the temp table if it already exists
-        "DROP TABLE IF EXISTS " TEMP_TABLE(TEMP_ELEMENT_DELETION) "",
-
-        // Create the temp table from scratch
+        // Create the temp table if needed; subsequent calls reuse it after clearing stale entries.
         "CREATE TABLE IF NOT EXISTS " TEMP_TABLE(TEMP_ELEMENT_DELETION) " ("
```
khanaffan, line 491: You might also want to truncate the table if it already exist. `DELETE FROM TEMP_TABLE(TEMP_ELEMENT_DELETION)`

aruniverse:
Also, add some unit tests for this (in native and/or core)

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
kring: The `DELETE FROM` already exists a little lower down.

kring (approved):
I've been using these changes locally with OS+ and can confirm it solves the problem.

rschili (approved):

RohitPtnkr1996:
> Also, add some unit tests for this (in native and/or core)

Done!
Test added for the failing scenario in the iTwinjs PR.

aruniverse (approved):

aruniverse:
@mergifyio backport release/5.9.x

PR #1428, opened 2026-05-14
Allow element to change its parent or model or both

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9303

fixes: https://github.com/iTwin/itwinjs-backlog/issues/2070

This pull request introduces a new capability to move elements between models and/or parents within the DgnDb, along with robust validation of element-model compatibility. It adds a new `MoveElement` method to the API, implements model-specific validation logic, and exposes the new functionality to the Node.js native addon. These changes ensure that only compatible elements can be moved to a target model, preventing invalid state and improving API usability.


iModelJsNodeAddon/JsInteropDgnDb.cpp
```diff
@@ -546,6 +546,61 @@ void JsInterop::UpdateElement(DgnDbR dgndb, Napi::Object obj) {
...
+    if (!elementId.IsValid())
+        throwInvalidId();
+
+    DgnModelId newModelId = props["targetModelId"].GetId64<DgnModelId>();
+    DgnElementId targetElementId = props["targetElementId"].GetId64<DgnElementId>();
+    DgnElementId newParentId;
+
+    if (!newModelId.IsValid() && !targetElementId.IsValid())
+        throwInvalidId(); // at least one of targetModelId or targetElementId must be specified
+
```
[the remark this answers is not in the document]
khanaffan: We do check on typescript side. This is just safeguard.

rschili (approved):

soham-bentley (approved):

khanaffan:

naveedkhan8067 (approved):

rschili (changes requested):
I'm re-reviewing this, withdrawing my earlier approve until I'm finished, a lot has changed.

rschili (approved):

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -1500,6 +1500,261 @@ DgnDbStatus DgnElements::UpdateElement(DgnElementR replacement)
...
     }
 
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+DgnDbStatus DgnElements::ChangeElementParent(DgnElementId elementId, DgnElementId newParentId)
+    {
+    DgnDb::VerifyClientThread();
+
+    if (!newParentId.IsValid())
```
aruniverse, line 1510: This guards `!newParentId.IsValid()` but not `elementId == newParentId`. When they're equal, all subsequent checks pass: the parent element loads fine, it's in the same model, no code constraint fires — so a self-referential `ParentId` gets written to the DB.

That row is now corrupted. Any future call to `ValidateSubtreeForModelChange` on it calls `element.QueryChildren()`, which returns the element itself, recurses back in, and overflows the stack. Any parent-chain traversal also loops forever.

Needs a guard immediately after `VerifyClientThread()`:
```cpp
if (elementId == newParentId)
    return DgnDbStatus::InvalidId;
```

A full cycle check (walking up from `newParentId` to verify it isn't already a descendant of `elementId`) would be the complete fix, but at minimum self-reparent must be blocked.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -1500,6 +1500,261 @@ DgnDbStatus DgnElements::UpdateElement(DgnElementR replacement)
...
+        return DgnDbStatus::InvalidId;
+
+    // 4. BIS requires a parent and all of its children to reside in the same model, so changing the model of a
+    // root element relocates its entire subtree. Validate the whole subtree first (no mutation) so a rejected
+    // move leaves the iModel untouched, then perform the relocation.
+    DgnDbStatus stat = ValidateSubtreeForModelChange(elementId, *newModel, true);
+    if (DgnDbStatus::Success != stat)
+        return stat;
+
+    return MoveSubtreeToNewModel(elementId, newModelId, *newModel);
```
aruniverse, line 1598: `ValidateSubtreeForModelChange` only calls `_ValidateElementForModel` — it doesn't invoke `_OnUpdate`. `MoveSubtreeToNewModel` does call `_OnUpdate` (via `_UpdateInDb`). So an `_OnUpdate` override that rejects a model transition can fail on, say, the second child after the first child and its entire subtree have already been relocated. The per-element rollback inside `MoveSubtreeToNewModel` only reverts the single element that failed — it doesn't undo completed sibling moves.

This needs a savepoint wrapping the entire mutation so any mid-subtree failure is atomic:
```cpp
DgnDbStatus stat = ValidateSubtreeForModelChange(elementId, *newModel, true);
if (DgnDbStatus::Success != stat)
    return stat;

auto sp = m_dgndb.GetSavepoint("ChangeElementModel");  // or whatever DgnDb's savepoint API is
stat = MoveSubtreeToNewModel(elementId, newModelId, *newModel);
if (DgnDbStatus::Success != stat)
    sp.Cancel();
return stat;
```

iModelCore/iModelPlatform/DgnCore/DgnModel.cpp
```diff
@@ -564,6 +564,56 @@ DgnDbStatus InformationModel::_OnInsertElement(DgnElementR element)
...
+    return (nullptr != dynamic_cast<GroupInformationElement const*>(&element)) ? DgnDbStatus::Success : DgnDbStatus::WrongModel;
+    }
+
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+DgnDbStatus InformationRecordModel::_ValidateElementForModel(DgnElementCR element) const
+    {
+    return (nullptr != dynamic_cast<InformationRecordElement const*>(&element)) ? DgnDbStatus::Success : DgnDbStatus::WrongModel;
+    }
```
aruniverse, line 615: Three models have `_OnInsertElement` restrictions that aren't covered by the new `_ValidateElementForModel` overrides. Since `ChangeElementModel` uses `_ValidateElementForModel` (not `_OnInsertElement`), incompatible element types can be moved into these models without error:

| Model | `_OnInsertElement` restriction | Gap |
|---|---|---|
| `DocumentListModel` | `IsDocumentElement()` only (L683) | inherits `InformationModel` — accepts all `IsInformationContentElement()` |
| `RepositoryModel` | excludes `IsDefinitionElement()` (L776) | inherits `InformationModel` — `DefinitionElement`s accepted |
| `RoleModel` | rejects `IsGeometricElement()` (L784) | no override — `DgnDbStatus::Success` for everything |

Suggested additions (plus matching declarations in `DgnModel.h`):
```cpp
DgnDbStatus DocumentListModel::_ValidateElementForModel(DgnElementCR element) const
    { return element.IsDocumentElement() ? DgnDbStatus::Success : DgnDbStatus::WrongModel; }

DgnDbStatus RepositoryModel::_ValidateElementForModel(DgnElementCR element) const
    { return (element.IsInformationContentElement() && !element.IsDefinitionElement()) ? DgnDbStatus::Success : DgnDbStatus::WrongModel; }

DgnDbStatus RoleModel::_ValidateElementForModel(DgnElementCR element) const
    { return element.IsGeometricElement() ? DgnDbStatus::WrongModel : DgnDbStatus::Success; }
```

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -1500,6 +1500,261 @@ DgnDbStatus DgnElements::UpdateElement(DgnElementR replacement)
...
+                {
+                GeometrySourceCP geom = finalElem->ToGeometrySource();
+                if (nullptr != geom && geom->HasGeometry())
+                    newGeomModel->AddElementRange(elementId, geom->CalculateRange3d());
+                }
+            }
+        }
+
+    // Recurse into the children so the entire subtree moves with the element (parent moved first).
+    for (DgnElementId childId : children)
```
aruniverse: Standard `UpdateElement` calls `parent->_OnChildUpdate(element, replacement)` before writing, giving the parent a chance to veto. This recursion loop doesn't. `ViewDefinition::_OnChildUpdate` and `Sheet::_OnChildUpdate` both unconditionally return `DgnDbStatus::InvalidParent`, meaning they block all child updates. When the subtree recursion reaches children of those parent types, those children are moved without the parent's veto being consulted — behaviour that `UpdateElement` would never allow.

Before calling `_UpdateInDb` for each element in `MoveSubtreeToNewModel`, look up the element's current parent (if it exists and isn't the root being relocated in this same call) and call `parent->_OnChildUpdate(element, *editCopy)`, returning the error if it vetoes.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
khanaffan: Fixed in `645d78eed8`. `ChangeElementParent` now checks the return status from `SetParentId(...)` before calling `UpdateElement`. `_SetParentId` already rejects `elementId == newParentId` with `DgnDbStatus::InvalidParent`, so self-parenting is blocked before any DB write.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
khanaffan: Fixed in `645d78eed8`. The mutation phase no longer recurses element-by-element or calls `_OnUpdate` / `_UpdateInDb` / `_OnUpdated`. After validation succeeds, `MoveSubtreeToNewModel` performs one recursive-CTE `UPDATE` for the whole subtree, then updates caches/range indexes/source-model bookkeeping/geometric-model tracking explicitly. That removes the partial-mutation failure mode this comment was pointing out.

iModelCore/iModelPlatform/DgnCore/DgnModel.cpp
khanaffan: I think this is already covered by the TypeScript wrapper for the exposed API path. `EditTxn.changeElementModel` requires the source and target model classes to match exactly before native is called (`sourceModel.classFullName !== targetModel.classFullName` throws `ElementError("model-type-mismatch", ...)`).

So callers cannot move arbitrary elements into `DocumentListModel`, `RepositoryModel`, or `RoleModel` through this API unless the source model is already that same model class. In that same-class case, the element already satisfied that model class's content restriction in its source model. Also, consumers cannot realistically inherit and create new model types beyond the model classes we already have, so I think the additional native `_ValidateElementForModel` overrides are redundant for this PR's public API path.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
khanaffan: This path changed in `645d78eed8`: `MoveSubtreeToNewModel` no longer calls `_UpdateInDb` per element. It validates the subtree first, then performs one recursive-CTE `UPDATE` for the model change and handles cache/range/model-change bookkeeping explicitly. The model change is now atomic at the SQL statement level and intentionally does not fire per-element update callbacks/events.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
aruniverse: Verified — `_SetParentId` at L890–891 catches direct self-reparent: `if (parentId.IsValid() && parentId == GetElementId()) return DgnDbStatus::InvalidParent`. The function comment even says *"Check for direct cycle...will check indirect cycles on update"*, and `_OnUpdate` does exactly that via `parentCycleExists(parentId, GetElementId(), GetDgnDb())` which walks the full ancestor chain before any write is committed. Both the direct and transitive cycle cases are covered. My original comment was wrong to call this unfixed — the protection was already there through the `UpdateElement` path. ✅ Can close.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
aruniverse: Understood — the redesign removes the per-element `_UpdateInDb` loop entirely, so `_OnChildUpdate` is no longer in the picture. The design tradeoff (bulk SQL, no per-element callbacks) is documented in the code comment at ~L1671. Can close.

iModelCore/iModelPlatform/DgnCore/DgnModel.cpp
aruniverse: Fair point for callers going through the TypeScript API — the `sourceModel.classFullName !== targetModel.classFullName` guard in `EditTxn.changeElementModel` blocks cross-class model moves before native is ever reached, which covers the three model types in question. The remaining gap is direct native C++ callers bypassing the TS layer, where `_ValidateElementForModel` would be the only gatekeeper. If `DgnElements::ChangeElementModel` is treated as an internal/trusted API, the TS-layer guard is sufficient. Worth a comment in the header noting the precondition for native callers if so.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
aruniverse: ✅ The single bulk `UPDATE … WHERE Id IN (SELECT Id FROM subtree)` is atomic within the SQLite write transaction — no partial state is possible. Cleaner than a savepoint. Can close.

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
```diff
@@ -1500,6 +1500,253 @@ DgnDbStatus DgnElements::UpdateElement(DgnElementR replacement)
...
+    updateStmt->BindId(1, elementId);
+    updateStmt->BindId(2, newModelId);
+    updateStmt->BindId(3, newModelId);
+    if (BE_SQLITE_DONE != updateStmt->Step())
+        return DgnDbStatus::WriteError;
+
+    DropFromPool(subtreeIds);
+
+    for (DgnModelId sourceModelId : sourceModelIds)
+        m_dgndb.Txns().m_modelChanges.AddModel(sourceModelId);
```
aruniverse, line 1735: The destination model (`newModelId`) is never passed to `AddModel`. Only the source models are tracked here:

```cpp
for (DgnModelId sourceModelId : sourceModelIds)
    m_dgndb.Txns().m_modelChanges.AddModel(sourceModelId);  // ← newModelId missing
```

For geometric models the destination gets implicit change-tracking through the `AddGeometricElementChange(newModelId, …, Insert)` call below (~L1743), but for non-geometric information models (e.g. moving elements between two `InformationRecordModel`s) there are no `AddGeometricElementChange` calls for the destination, and `AddModel(newModelId)` is never called. Any consumer that walks `m_modelChanges` to discover which models were modified (change summary, briefcase merge, push-pull diffs) will miss the destination model for non-geometric moves.

Fix: add `m_dgndb.Txns().m_modelChanges.AddModel(newModelId);` in the same block, either inside or just after the `sourceModelIds` loop.

aruniverse (approved):

iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
khanaffan: fixed.

aruniverse: ✅ Confirmed fixed in `ec447c58`. `AddModel(newModelId)` is now called after the source-model loop, guarded by `!sourceModelIds.empty()` (correct — if nothing moved there is nothing to track), with a comment that accurately explains the non-geometric case. Can close.

bbastings (approved):
No objections, hopefully whoever requested this will make sure it suits their requirements.

PR #1431, opened 2026-05-15
Update to use vcpkg for zlib/minizip

### LINKED ISSUE #1416: Investigate using vcpkg for third-party libraries: zlib
Investigate using the [vcpkg](https://vcpkg.io/) dependency manager to manage third-party libraries. Cesium already uses vcpkg.

Update zlib to build using vcpkg instead of bmake.

tcobbs-bentley:
Use vcpkg to build zlib/minizip instead of manually importing new source code every time there is a new release.

Note: new software was installed on CI Macs.


iModelCore/libsrc/compress/vcpkg-configuration.json
```diff
@@ -0,0 +1,7 @@
+{
+  "default-registry": {
+    "kind": "git",
+    "baseline": "81de6771512413aaf89ea77add5ad1fda126b9d0",
+    "repository": "https://github.com/microsoft/vcpkg"
```
naveedkhan8067, line 5: Is this repository going to be cloned into our source? If yes we should consider following:

-  Shouldn't we add it to our bbconfig? it contains all our repo and package dependencies.
- Is it going to be scanned by mend? If yes, consider the case if there are any vulnerabilities reported from it wouldn't our build be blocked until they fix?

iModelCore/libsrc/compress/vcpkg.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "imodel-native-compress",
+  "version": "1.0.0",
+  "dependencies": [
+    {
+      "name": "zlib",
+      "version>=": "1.3.2"
```
naveedkhan8067, line 7: Are we also allowing auto major version update? What if it has breaking changes?

iModelCore/libsrc/git_helper.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
+"""Small git helper used by vcpkg.mke for cross-platform command execution."""
```
naveedkhan8067: Why we need a separate script to support the clone? Can't we reuse how other repo dependencies are added?

iModelCore/libsrc/compress/vcpkg.json
naveedkhan8067: ok `overrides` property should handle this.

iModelCore/libsrc/compress/triplets/x64-osx.cmake
```diff
@@ -0,0 +1,10 @@
+set(VCPKG_TARGET_ARCHITECTURE x64)
```
naveedkhan8067: I think we only support ARM64 now?

iModelCore/libsrc/vcpkg.mke
```diff
@@ -0,0 +1,49 @@
...
+
+# To update vcpkg, set VCPKG_TAG to the new tag from https://github.com/microsoft/vcpkg/tags
+# and set VCPKG_COMMIT to the full commit hash shown on that tag's page.
+# You can also get the commit hash by running:
+#   git ls-remote --tags https://github.com/microsoft/vcpkg.git <tag>^{}
+
+# What version of vcpkg to use
+VCPKG_TAG=2026.05.25
+# The commit hash for the above version.
+VCPKG_COMMIT=d015e31e90838a4c9dfa3eed45979bc70d9357fc
```
naveedkhan8067: Rather than hardcoding it in the mke file, shouldn't we define it in the config file?

iModelCore/libsrc/git_helper.py
tcobbs-bentley: On Windows, `bmake` was converting all forward slashes to backslashes (including in the URL). The reason I did it this way instead of adding vcpkg to bbconfig.json was so that I could do the hash verification after checkout to make sure the specific version of vcpkg hadn't been tampered with (since this is an external source, unlike all the others in bbconfig.json). I guess the hash validation could be done even using bb for the checkout. Thoughts?

iModelCore/libsrc/compress/triplets/x64-osx.cmake
tcobbs-bentley: Fixed. It now errors out if MacOSX64 is selected as the platform, and this triplet file has been deleted.

iModelCore/libsrc/vcpkg.mke
tcobbs-bentley: As noted for the Python script, I can do this, but will need to still do the hash check since this is a third-party repository.

iModelCore/libsrc/compress/vcpkg-configuration.json
tcobbs-bentley: Updated to pull vcpkg as part of `bb pull`.

I'm not sure about the mend scan part. I will investigate.

tcobbs-bentley: @DanRod1999 Does mend scan everything or just specific things? Should it scan the vcpkg build tool?

iModelCore/libsrc/git_helper.py
tcobbs-bentley: Script is now gone.

iModelCore/libsrc/vcpkg.mke
tcobbs-bentley: I updated so that vcpkg is pulled as part of `bb pull`.

iModelCore/libsrc/compress/vcpkg-configuration.json
DanRod1999: mend scans should only scan what is released and delivered publicly. I don't think we would need to scan this. For example we don't dev deps, which I think we can consider this as.

DanRod1999 (approved):

aruniverse (approved):

iModelCore/libsrc/compress/vcpkg-configuration.json
naveedkhan8067: So have we already excluded it from the paths needs to be scanned? Otherwise it will a blocker same as BentleyBuild.

tcobbs-bentley: Since vcpkg goes directly into SrcRoot, I _think_ that it won't be scanned, but I'm not sure. @DanRod1999 what do you think?

DanRod1999: in pipline.yml this is where we decide what gets scanned or not
 ```
  - name: WS_EXCLUDES
    value: '**/BentleyBuild/ **/iModelJsNodeAddon/api_package/ **/iModelCore/Visualization/TilesetPublisher/test/ **/BisSchemas/'
```

and we scan starting in this dir
```
- name: BB_SRC_ROOT
  value: $(Agent.BuildDirectory)/bbsrc
```

tcobbs-bentley: Thanks. I'll create a new imodel-native-internal PR and link it.

tcobbs-bentley: I created an imodel-native-internal PR, but did not link it to this one, because that would complicate the build for no gain.

PR #1435, opened 2026-05-19
Marked cascading changes as indirect

RohitPtnkr1996:
In the bulk delete api implementation, to avoid per row cascade delete triggers which caused a drop in performance, the api was deferring foreign keys and bulk deleting the cascading rows before the element deletion.
As a result, the cascade deletes were also being tracked as direct changes, whereas sqlite tracks these cascades as indirect changes.
This caused a problem on OS+ which needs the changes to be marked as direct/indirect.

This PR sets the ChangeTracker's mode to indirect for the scope of the cascade delete changes and resets the mode to the previous value at the end.

Verified the fix locally on OS+.

### DIFF
--- a/iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
+++ b/iModelCore/iModelPlatform/DgnCore/DgnElements.cpp
@@ -1100,6 +1100,10 @@ DbResult BulkElementDeletion::ExecuteDeletion()
     {
     // Explicitly execute ON DELETE CASCADE and ON DELETE SET NULL FK actions for all tables that reference bis_Element.
     {
+    const auto changeMode = m_dgndb.Txns().GetMode();
+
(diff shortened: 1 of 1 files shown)


kring:
This is needed for iTwin/itwinjs-backlog#2054, and I've confirmed that it works as intended. It preserves the previous behavior (before the bulk delete API) where, e.g., element aspect deletions (which were previously triggered by a CASCADE) are considered indirect changes.

kring (approved):

aruniverse:
@RohitPtnkr1996 @kring @wgoehrig we need this backported for 5.9.x for the june release right?

wgoehrig:
@aruniverse yes, we will want to backport this

aruniverse:
@mergifyio backport release/5.9.x

PR #1438, opened 2026-05-20
Add file-based transaction support and enhance revert functionality to support large size and number of changesets

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9322

### Issue

RevertTimelineChanges could not handle large amounts of changes or very large changesets. The limitation is SQLite's inability to store blobs greater than 2 GB in the dgn_Txn table. Reverting anything exceeding 2 GB resulted in a failure when SQLite could not allocate the blob.

### Solution

Add support for file-based transaction storage. When enabled, changeset data is written to individual .txn files on disk (LZMA-compressed) rather than stored as blobs in dgn_Txns. Only a small header (signature + uncompressed size + SHA1 checksum) is stored in the Change column.

This is activated automatically when calling RevertTimelineChanges, which enables the fileBasedTxns briefcase-local property and saves each reverted changeset individually with SaveChanges("reverted changeset <index>").


iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
```diff
@@ -182,6 +182,15 @@ DbResult DgnDb::_OnDbOpened(Db::OpenParams const& params)
         return rc;
         }
 
+    // Validate file-based txn files if enabled
+    if (m_txnManager.IsValid() && Txns().IsFileBasedTxnEnabled()) {
```
rschili, line 186: Should there be a recovery path for this?
Sounds like the .bim is no longer self-contained, and if something happens to the txns/ directory, the user will be stuck unable to open the db anymore.

iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
```diff
@@ -3741,65 +4142,135 @@ DbResult TxnManager::PullMergeUpdateTxn(ChangeSetCR changeSet, TxnId id) {
...
-    rc = stmt->Step();
-    if (BE_SQLITE_DONE != rc) {
-        LOG.errorv("SaveChangeSet failed: %s", BeSQLiteLib::GetErrorName(rc));
-        BeAssert(false);
-        return rc;
-    }
+        // DB update succeeded — now safely replace the old file
+        if (finalPath.DoesPathExist())
+            finalPath.BeDeleteFile();
+        BeFileName::BeMoveFile(tempPath, finalPath);
```
rschili: BeMoveFile's return value is ignored here, and BeDeleteFile returns a status too. If the delete-then-move sequence fails halfway (likeliest on Windows, where MoveFileExW is called without MOVEFILE_REPLACE_EXISTING), the DB row references a checksum with no corresponding .txn file on disk and the caller gets back BE_SQLITE_OK.

Suggested change: replace the delete-then-move with a backup-then-move-then-delete pattern, check every BeMoveFile / BeDeleteFile return, and return BE_SQLITE_ERROR (with best-effort restore of the backup) on the install failure so the caller's transaction can roll the DB row back. Cleanup-delete failures should at least be logged, not swallowed.

Same unchecked tempPath.BeDeleteFile() and the new DeleteTxnFile / DeleteAllTxns / DeleteFromStartTo call sites would benefit from the same treatment - orphaned .txn files won't be fatal but should not pile up silently.

rschili (approved):

iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
khanaffan: This is currently only limited to revertTimelineChanges(). But there is no recovery if txn file is missing, corrupt etc. We have untracked local changes and so we cannot push or pull. Txn is way to track what's local vs remote. if you do not have that briefcase is as good as corrupt. Any changes push from it can currupt the change history where someone pull can end up with conflicts.

iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
khanaffan: fixed

soham-bentley (approved):

rschili:
@pmconne @bbastings One of you needs to approve this. Affan asked me to babysit this PR while he is travelling. Only the three of you can approve changes to iModelPlatform, and since Affan created it...

PR #1439, opened 2026-05-20
Code swap does not take effect when second user pulls changes

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9289

SQLite update is from https://github.com/iTwin/imodel-native/releases/tag/itwin-sqlite-v3.53.2-r1

Following commits is part of 3.53.2

[Update the session module so that it can apply changesets containing two or more UPDATE changes that form a dependency loop - so that no single UPDATE can be applied independently without violating a constraint.](https://github.com/iTwin/sqlite/commit/3eb6a5ee33cf6ec66ef41b25222cf823cb185a78)

[Add the SQLITE_CHANGESETAPPLY_NOUPDATELOOP option to the sqlite3_changeset_apply_v2/3() method. To disable the extra processing to commit changesets that swap two or more values subject to a UNIQUE constraint between rows.
](https://github.com/iTwin/sqlite/commit/f380599476317bacfb234926916028a69567614e)


khanaffan:

soham-bentley:
The `noUpdateLoop` boolean flag which we are now carrying for `MergeChangeset` and `ApplyChangeset` methods, as far as I understand that flag is just to preserve old behavior where in if we swap two values in a column with UNIQUE_CONSTRAINT, it would be considered CONSTRAINT conflict. But will we ever need this old behavior going forward?
@khanaffan 
If not we can simple not carry the boolean flag

soham-bentley (approved):

iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
```diff
@@ -2524,9 +2524,6 @@ DbResult TxnManager::ApplyChanges(ChangeStreamCR changeset, TxnAction action, bo
     const bool ignoreNoop = pmConf.IsRebasingLocalChanges();
     bool fastForwardEncounteredMergeConflict = false;
     auto fastForwardConflictHandler = [&fastForwardEncounteredMergeConflict](ChangeStream::ConflictCause _, Changes::Change change) {
-        if(change.IsIndirect())
```
rschili, line 2527: Why drop this check? I don't see an explanation, it changes behavior - do we want this?

iModelCore/BeSQLite/SQLite/sqlite3.c
```diff
@@ -238491,41 +238814,119 @@ static int sessionRetryConstraints(
...
+
+    sqlite3_free(cons.aBuf);
+    if( rc!=SQLITE_OK ) break;
+
+    /* If no progress has been made this round, break out of the loop. */
+    if( pApply->constraints.nBuf>=cons.nBuf ) break;
+  }
+
+  /* Step (2) */
+  while( rc==SQLITE_OK && pApply->constraints.nBuf && !pApply->bNoUpdateLoop ){
```
rschili, line 238985: Re-raising the two points Copilot flagged here, since I don't think the green test covers them.

This !bNoUpdateLoop-guarded loop (238835-238851) never does real work: cons stays {0,0,0}, so the progress check pApply->constraints.nBuf >= cons.nBuf is >= 0 and always breaks on the first iteration. It also allocates pUp via sessionRetryIterInit and breaks without finalizing it, so it leaks a changeset iterator on the default path every time constraints are deferred.

As a result the bNoUpdateLoop guard is on the wrong loop. The real Step (2) loop just below (238853+) runs unconditionally, so SQLITE_CHANGESETAPPLY_NOUPDATELOOP has no effect and the whole flag plumbing (ChangeSet -> TxnManager -> addon -> NativeLibrary.ts) is currently dead. I confirmed against origin/main that this Step (1)/(2)/(3) + delete/insert resolution is a new custom patch (sessionUpdateToDeleteInsert / sessionRetryIterInit don't exist in the prior amalgamation), so this reads like a half-finished edit rather than upstream SQLite.

The existing test exercises neither problem: the leak needs ASAN to surface, and nothing sets noUpdateLoop=true to prove the flag works. Could you confirm the intent? If the flag is meant as an escape hatch to the old give-up behavior, the guard should move to the real loop and the duplicate skeleton removed. If we don't need the old behavior, dropping the flag entirely seems cleaner.

rschili: This is the test that claude proposed to hit this problem:

```cpp
// @bsimethod
// Demonstrates that the SQLITE_CHANGESETAPPLY_NOUPDATELOOP flag (SetNoUpdateLoop) is not
// honored. The flag is meant to be an escape hatch back to the old "give up" behavior:
// when set, the new delete/insert update-loop resolution should be skipped, so a unique
// index swap can no longer be auto-resolved and instead surfaces to the conflict handler
// (the original silent-data-loss path - here the handler SKIPs, so the swap is dropped).
//
// CURRENT BEHAVIOR (FAILS): the !bNoUpdateLoop guard sits on a dead duplicate loop in
// sessionRetryConstraints, so the real update loop always runs. The swap is resolved, the
// conflict handler is never called (conflictCount == 0) and db2 ends up swapped (goo/foo).
// Once the guard is moved to the real Step (2) loop, this test passes: the handler is
// invoked and db2 retains its original values (foo/goo).
TEST_F(BeSQliteTestFixture, no_update_loop_flag_skips_swap_resolution)
    {
    auto db1 = Create("first.db");
    ASSERT_EQ(BE_SQLITE_OK, db1->ExecuteSql("CREATE TABLE test_swap(id INTEGER PRIMARY KEY, name TEXT)"));
    ASSERT_EQ(BE_SQLITE_OK, db1->ExecuteSql("CREATE UNIQUE INDEX uidx_test_swap_name ON test_swap(name)"));
    ASSERT_EQ(BE_SQLITE_OK, db1->ExecuteSql("INSERT INTO test_swap(id, name) VALUES(1, 'foo')"));
    ASSERT_EQ(BE_SQLITE_OK, db1->ExecuteSql("INSERT INTO test_swap(id, name) VALUES(2, 'goo')"));
    db1->SaveChanges();
    db1->CloseDb();

    ASSERT_EQ(BeFileNameStatus::Success, Clone("first.db", "second.db"));

    db1 = OpenReadWrite("first.db");
    auto db2 = OpenReadWrite("second.db");
    ASSERT_TRUE(db1 != nullptr);
    ASSERT_TRUE(db2 != nullptr);

    // Capture the same net swap changeset as apply_changeset_swap_unique_index_values:
    // row 1 foo->goo, row 2 goo->foo.
    std::unique_ptr<BeSQLite::ChangeSet> cs = Capture(*db1, [](DbR db, void*) {
        if (BE_SQLITE_OK != db.ExecuteSql("UPDATE test_swap SET name='__temp__' WHERE id=1"))
            return false;
        if (BE_SQLITE_OK != db.ExecuteSql("UPDATE test_swap SET name='foo' WHERE id=2"))
            return false;
        if (BE_SQLITE_OK != db.ExecuteSql("UPDATE test_swap SET name='goo' WHERE id=1"))
            return false;
        return true;
        }, nullptr);

    ASSERT_TRUE(cs != nullptr);
    db1->SaveChanges();

    // Apply with NoUpdateLoop enabled. The handler counts CONSTRAINT conflicts and SKIPs
    // them, mirroring the original conflict-handler behavior that silently loses swaps.
    int conflictCount = 0;
    BeSQLite::ApplyChangesArgs args;
    args.SetNoUpdateLoop(true);
    args.SetConflictHandler([&conflictCount](BeSQLite::ChangeStream::ConflictCause cause, BeSQLite::Changes::Change change) {
        conflictCount++;
        return BeSQLite::ChangeStream::ConflictResolution::Skip;
        });

    ASSERT_EQ(BE_SQLITE_OK, cs->ApplyChanges(*db2, args));
    db2->SaveChanges();

    // With the flag honored, the update loop is skipped and the swap conflicts surface.
    ASSERT_GT(conflictCount, 0) << "NoUpdateLoop did not skip the update-loop resolution; the flag has no effect";

    // The skipped swap leaves db2 with its original (pre-swap) values.
    {
    auto stmt = db2->GetCachedStatement("SELECT name FROM test_swap WHERE id=1");
    ASSERT_EQ(BE_SQLITE_ROW, stmt->Step());
    ASSERT_STREQ("foo", stmt->GetValueText(0)) << "swap was resolved despite NoUpdateLoop=true";
    }
    {
    auto stmt = db2->GetCachedStatement("SELECT name FROM test_swap WHERE id=2");
    ASSERT_EQ(BE_SQLITE_ROW, stmt->Step());
    ASSERT_STREQ("goo", stmt->GetValueText(0)) << "swap was resolved despite NoUpdateLoop=true";
    }
    }
```

Sorry I am right now 3 layers deep in discussions and PR reviews, this here seems to deserve some more serious thinking before we can confirm.

khanaffan: Oh you meant this issue. I thought it was about memory error. I know about this one. It was reported by copilot the first time I assign it PR and I verify it by a test as well from itwinjs side. 

I waiting for fix for this.

khanaffan: So it turn out there was issue cherry picking branch changes into release version. Dan said I can choose to pick 3.53.2 which will be release in few days.

I will hold on and merge upgrade SQLite in separate PR in few days.

iModelCore/BeSQLite/Tests/NonPublished/BeSQLite_Test.cpp
```diff
@@ -455,6 +455,103 @@ TEST_F(BeSQliteTestFixture, sqlite_stat1)
...
+        if (BE_SQLITE_OK != db.ExecuteSql("UPDATE test_swap SET name='goo' WHERE id=1"))
+            return false;
+        return true;
+        }, nullptr);
+
+    ASSERT_TRUE(cs != nullptr);
+    db1->SaveChanges();
+
+    // Verify the changeset contains exactly 2 UPDATE operations with the net swap values.
+    // The session extension collapses the 3 intermediate UPDATEs into 2 net changes per row.
```
kring: Nitpick:

```suggestion
    // The session extension collapses the 3 intermediate UPDATEs into 2 net changes, one per row.
```

kring (approved):
I also have the same question ~Rob~ Soham did about whether we really need the "noUpdateLoop" option. But other than that and a super minor nit in a comment (below), this looks good to me!

khanaffan:
@kring @soham-bentley 

The main reason I believe the `noUpdateLoop` flag is necessary is due to how our checkpoint agent applies existing changesets. We’re encountering issues in this process, and I wanted to provide a way to disable this behavior for all existing changesets when needed.

In earlier scenarios, when code swaps were ignored, connectors (such as the application) could start with an inconsistent view of the database. This mismatch sometimes caused them to generate data that resulted in swap codes being reapplied unnecessarily.

To address this, I wanted to achieve two key goals:

1. Enable this new behavior for all newly generated changesets.
2. Provide the ability to fall back to the old behavior for existing changesets in the hub, in case issues are discovered after adopting the new approach.

I initially attempted to add a property directly to the changeset to control this behavior. However, that approach quickly became overly complex. As a result, I decided to leave the responsibility to the application, allowing it to determine whether it needs to fall back to the previous behavior.

I do not hope we hit any case where we fail to apply historical changeset but if we do reverting to previous behavior should not require us to release a new version of itwinjs on urgent bases and downstream application has to release there versions to consume the fix.

iModelCore/iModelPlatform/DgnCore/TxnManager.cpp
khanaffan: reverted

rschili (approved):

bbastings (approved):

aruniverse:
## Adversarial Code Review — PR #1439

Three parallel reviewers examined this PR across **correctness wiring**, **memory/resource safety**, and **test/API completeness**. Findings are ranked by severity.

---

### 🔴 Fixes Worth Doing Now

#### 1. `noUpdateLoop` silently dropped on the reverse-changeset path
**File:** `iModelJsNodeAddon/IModelJsNative.cpp` (~line 2856)

The PR correctly threads `noUpdateLoop` into the merge path but not the reverse path:

```cpp
if (revision->GetParentId() == currentId)  // merge
    stat = db.Txns().MergeChangeset(*revision, fastForward, noUpdateLoop);  // ✓ flag passed
else if (revision->GetChangesetId() == currentId) //reverse
    db.Txns().ReverseChangeset(*revision);                                   // ✗ noUpdateLoop DROPPED
```

`ReverseChangeset` signature is unchanged and does not accept `noUpdateLoop`. A TypeScript caller that passes `noUpdateLoop=true` gets the flag honoured on merge but **silently ignored on reverse**, producing asymmetric behaviour at the same API entry point.

**Fix:** Either propagate `noUpdateLoop` into `ReverseChangeset` (add the param, thread it into its `ApplyChanges` call), or explicitly document that the flag has no effect on the reverse path.

---

#### 2. Bind loop in `sessionUpdateToDeleteInsert` swallows bind failures
**File:** `iModelCore/BeSQLite/SQLite/sqlite3.c` — new `sessionUpdateToDeleteInsert` function

```c
for(iCol=0; iCol<pApply->nCol; iCol++){
    ...
    rc = sqlite3_bind_value(pRet, iCol+1, pVal);   // overwrites rc every iteration — no early exit
}
```

No `if( rc!=SQLITE_OK ) break;` guard. If column *k* fails (e.g., `SQLITE_NOMEM`) but column *k+1* succeeds, the error is silently discarded. The subsequent `if( rc!=SQLITE_OK ){ sqlite3_finalize(pRet); pRet=0; }` guard is never triggered, and a partially-bound INSERT statement runs with a silently wrong value for the failed column.

**Fix:** Add `if( rc!=SQLITE_OK ) break;` inside the loop (or fold into the loop condition: `for(iCol=0; rc==SQLITE_OK && iCol<pApply->nCol; iCol++)`).

---

### 🟡 Optional Improvements

#### 3. `SAVEPOINT update_op` left dangling when `sessionUpdateToDeleteInsert` fails
**File:** `sqlite3.c` — new `sessionRetryConstraints` Step 2

When `sessionUpdateToDeleteInsert` returns non-OK, neither `RELEASE update_op` nor `ROLLBACK TO update_op` is executed. The savepoint is left open.

In the iTwin context, `ChangeStream::ApplyChanges` always sets `SQLITE_CHANGESETAPPLY_NOSAVEPOINT` (confirmed at `ChangeSet.cpp:894, 912`), meaning there is no outer `changeset_apply` savepoint to provide implicit cleanup. The dangling savepoint is eventually cleaned up by `AbandonChanges()` → full SQLite `ROLLBACK` in the error handling path, so **data integrity is maintained**. Low urgency in iTwin's usage pattern, but worth fixing in the SQLite fork to avoid surprises if `NOSAVEPOINT` assumptions change.

#### 4. No test exercises `noUpdateLoop=true`
**File:** `iModelCore/BeSQLite/Tests/NonPublished/BeSQLite_Test.cpp`

The new test `apply_changeset_swap_unique_index_values` validates the **default behaviour** (update loop enabled, swap resolves with zero conflicts). It would pass identically even if the `NOUPDATELOOP` flag were never implemented — `SetNoUpdateLoop` is never called in the test, and `conflictCount == 0` is asserted but the handler is never invoked.

**Suggested addition:** A companion test that:
1. Sets `SetNoUpdateLoop(true)`
2. Applies the same swap changeset
3. Expects `ApplyChanges` to return non-`BE_SQLITE_OK`
4. Asserts `conflictCount >= 2` with `ConflictCause::Constraint`

Without this, the entire new feature path (step-2 skip → step-3 CONSTRAINT delivery) is exercised only by code review.

#### 5. `fastForward=true` + `noUpdateLoop=true` is a dangerous undocumented combination
**File:** `TxnManager.cpp` ~line 2526

The `fastForwardConflictHandler` ignores `ConflictCause` (parameter is named `_`). With `noUpdateLoop=true`, swap-UPDATE cycles surface as `CONSTRAINT` conflicts. The fast-forward handler returns `Abort` for any non-indirect change regardless of cause — a changeset that would apply cleanly with the retry loop silently fails as "conflict detected" when both flags are set together. The error message gives no indication that the flag combination caused the failure. A guard, comment, or log warning at the call site would help callers avoid this.

#### 6. `sessionVarintGet` masking should be documented
**File:** `sqlite3.c` ~line 233859

```c
static int sessionVarintGet(const u8 *aBuf, int *piVal){
  int ret = getVarint32(aBuf, *piVal);
  *piVal = (*piVal & 0x7FFFFFFF);   // custom iTwin addition — not in upstream SQLite 3.53.2
  return ret;
}
```

This is a custom divergence from upstream 3.53.2, which uses `SQLITE_CORRUPT_BKPT` guards in `sessionReadRecord` instead. The mask is safe (SQLite max value < 2^30), but silently truncates malformed values rather than returning `SQLITE_CORRUPT`. A comment explaining the defensive intent would help future maintainers, especially when pulling future SQLite upgrades.

---

### ✅ Confirmed Correct — No Action Needed

| Item | Verdict |
|---|---|
| Both `ApplyChangesArgs` instances set `SetNoUpdateLoop` (schema-only + data paths) | ✓ Correct |
| `invert + noUpdateLoop` flag combination semantics | ✓ Valid |
| TypeScript `undefined` as third argument handled by `IsBoolean()` check | ✓ Correct |
| Deleted `sqlite3session.h` — no remaining `#include` anywhere in the repo | ✓ Safe |
| `sessionChangesetBufferRecord` uses `>=` (not `>`) for initial type-byte check | ✓ Correct — `>` would be the bug |
| `sessionApplyOneOp` PK-bind condition reorder (`pOld &&` moved first) | ✓ Matches upstream; practical behaviour unchanged for well-formed changesets |
| Step (3) of `sessionRetryConstraints` is unconditional (not gated on `bNoUpdateLoop`) | ✓ Correct by design — this is the CONSTRAINT conflict delivery mechanism |

---

### Summary

Two runtime bugs worth addressing before merge: the **reverse-path `noUpdateLoop` drop** (PR-specific, straightforward fix in `IModelJsNative.cpp`) and the **bind loop failure-swallowing** in the new `sessionUpdateToDeleteInsert` function (report to the iTwin SQLite fork). Adding a `noUpdateLoop=true` test is the highest-value optional improvement — the entire new feature path currently has zero test coverage.

khanaffan:
@aruniverse the reverse-changeset path and divergent in sqlite session has been fixed.

PR #1442, opened 2026-05-27
Skip the automated core-dep bump in begin-addon-release

hl662:
## TL;DR

The publish-driven advance commits `package.json`/`package-lock.json` back to `imodel-native/main` with a fixed message. Without a guard, that commit would trip `begin-addon-release` and queue a pointless patch release. This skips the release only for that exact automation commit, by message — so real dependency PRs (including non-`@itwin` bumps like `minimatch`/`mocha`) still cut releases as before.

## What

| Asset | Why it exists |
|---|---|
| `build/begin_addon_release.yml` | Matches the exact automation commit message and skips just that, instead of blanket path-excluding `package.json`/`package-lock.json` — which would also drop legitimate dependency-bump PRs that touch only those files (e.g. #71 "Bump minimatch and mocha"). |

The guard sits alongside the existing rule that skips any commit without a `(#num)` PR suffix; the automation commit has no PR number, so it's caught either way — the explicit message check makes the intent clear in the logs.

---
*Nambot 🤖 (powered by claude-opus-4.8)*


build/begin_addon_release.yml
aruniverse, line 1: plz dont pollute pr diff w/ whitespace changes

build/begin_addon_release.yml
```diff
@@ -36,39 +36,56 @@ stages:
         displayName: Setup Azure CLI
 
       - bash: |
+            set -euo pipefail
+
             commitMsg="$(Build.SourceVersionMessage)"
+            safeCommitMsg="${commitMsg//'##vso'/'## vso'}"
+            echo "Commit message: $safeCommitMsg"
```
DanRod1999: Do this special line endings really belong after the `"`? also whats the point of this, just for logging purposes?

DanRod1999: i now see copilots review comment, nvm

DanRod1999:
All my comments in the other PR are also valid for here

hl662:
> All my comments in the other are also valid for here
Thanks, I've addressed these by cleaning up the unnecessarry `echo` of the commit msg, renamed the PR variable, and the QUEUE_PATCH variable was kept because in this pipeline an itwinjs_pr_num=0 could be a valid scenario.

We got:
 - SHOULD_QUEUE_PATCH=false → skip queue entirely
 - SHOULD_QUEUE_PATCH=true + ITWINJS_PR_NUM=0 → queue, just without a linked itwinjs-core PR
 - SHOULD_QUEUE_PATCH=true + ITWINJS_PR_NUM=12345 → queue with linked PR 12345

build/begin_addon_release.yml
```diff
@@ -36,39 +36,52 @@ stages:
...
         displayName: get itwinjs pr
         name: getItwinjsPr
 
       - bash: |
-          echo $(getItwinjsPr.ITWINJS_BRANCH)
-          az pipelines run \
-            --branch $(Build.SourceBranch) \
-            --id  8467 \
-            --parameters "BumpType=patch" "iTwinjsPrNum=$(getItwinjsPr.ITWINJS_PR_NUM)"
+          az pipelines run             --branch $(Build.SourceBranch)             --id 8467             --parameters "BumpType=patch" "iTwinjsPrNum=$(getItwinjsPr.ITWINJS_PR_NUM)"
```
DanRod1999: i think formatting got busted up here

hl662: ty that was a boo boo. fixed

DanRod1999 (approved):

PR #1443, opened 2026-05-27
Revert "bump @itwin/* dev dependencies to next minor version"

hl662:
## TL;DR

The branch-cut flow moved `imodel-native/main` to `^5.11.0-dev` before that dev line existed on npm, while the 5.10 patch lane was still active. That left native `main` pointing at packages it could not install.

This PR rolls the TypeScript API dev dependencies back to `^5.10.0-dev` so native `main` is usable again while the release automation fixes land.

## What

| Asset | Why it exists |
|---|---|
| `iModelJsNodeAddon/api_package/ts/package.json` | `main` was pointed at unpublished `5.11.0-dev` packages and started failing installs, so this restores the `5.10.0-dev` line that actually exists while `release/5.10.x` owns the active patch lane. |

---
*Nambot 🤖*

### DIFF
--- a/iModelJsNodeAddon/api_package/ts/package.json
+++ b/iModelJsNodeAddon/api_package/ts/package.json
@@ -17,10 +17,10 @@
     "test": "node ./lib/test/index.js"
   },
   "devDependencies": {
-    "@itwin/build-tools": "^5.11.0-dev",
-    "@itwin/core-bentley": "^5.11.0-dev",
-    "@itwin/core-common": "^5.11.0-dev",
-    "@itwin/core-geometry": "^5.11.0-dev",
+    "@itwin/build-tools": "^5.10.0-dev",
+    "@itwin/core-bentley": "^5.10.0-dev",
+    "@itwin/core-common": "^5.10.0-dev",
+    "@itwin/core-geometry": "^5.10.0-dev",
     "@itwin/eslint-plugin": "^5.0.0-dev.2",
     "@types/chai": "^4.3.6",
     "@types/chai-as-promised": "^7.1.6",


aruniverse:
why?

hl662:
The bot went ahead before I got to rein it in, but I updated the PR description.
Basically, we got a race condition caused from this: https://github.com/iTwin/imodel-native-internal/pull/1069
That meant the automation bumped main up to the next minor nightly, before it was even released.
Im working on the long term fix, the short term is to pull this back to 5.10 for now

aruniverse:
There will be a 5.11 nightly published tonight so this should be moot then right?

hl662:
welp, no 5.11 nightly published last night cuz there were no changes since last nightly. I'll just queue a manual nightly release and close this out -_-

aruniverse:
when we do the minor version bump, it should create a dev.0, did that not publish? might got stuck in the pipeline and need to reuque

hl662:
yeah thats the first immediate thing I tried 
image

Seems like the trigger for this build does not publish ADO artifacts of itjs to upload onto npm, so imma do it manually

rschili (approved):

hl662:
@aruniverse @khanaffan  the pipeline in my image does not contain artifacts, and to create a build run that produces an artifact we need a nightly version bump, then itwin.js CI build, then this pipeline is auto run.

I'm in favor of just approving this and  unblock our colleagues now instead of doing the above, im working on the long term fix

DanRod1999 (approved):

naveedkhan8067 (approved):

khanaffan (approved):

DanRod1999:
To clarify the exact problem that occurred:
This [PR](https://github.com/iTwin/imodel-native-internal/pull/1091) attempted to automatically update the pkg json in native to the next minor/major version of itwinjs-core. However, this takes place right after cutting the release branches of the 2 native repos.

There is no dependency on itwinjs-core and its releases. Because of this, it accidentally caused a race condition between native and core. The correct solution should be to update these deps after an itwinjs-core release happens instead of at the version bump of imodel-native.

I also have a [PR](https://github.com/iTwin/imodel-native-internal/pull/1091) reverting this pipeline automation, and we can correct it properly in the core pipelines

PR #1452, opened 2026-06-05
Presentation: Fix "actual source classes" not being set of nested content fields

grigasp:
https://github.com/iTwin/imodel-native/pull/788 made `GetActualSourceClasses` return a nullable set and only set the classes on direct related content fields, saying that the set is invalid for nested related content fields. 

That's not totally correct - with the fields structure `F1[A -> B] -> F2[B -> C]`, the "actual source classes" for both fields should be either A or its subclasses. That attribute is only invalid for F2 in a special context, where we create a descriptor from F1, and F2 becomes the root field. So instead of never setting "actual source classes" for F2, we should not set it only in that context.


grigasp:
@mergifyio backport release/5.10.x

saskliutas (approved):

PR #1454, opened 2026-06-08
Proposal - No-Remap POC

RohitPtnkr1996:
## 1. Background

When an ECSchema is upgraded in ECDb - for example, promoting a property from a subclass to a base class, or inserting a new class into the middle of an existing hierarchy - the existing "remap pipeline" detects that the new canonical column for a property differs from where data was previously stored, and **physically moves the data** into the new column.

This pipeline (`RemapManager`) is:
- Correct and well-tested.
- Required to hold a **write lock** for the duration of the upgrade.
- Expensive for large files: it performs full-table scans and row-by-row UPDATE operations.
- But more importantly from our perspective, it is internally complex. It must detect and handle circular column swaps, cross-table remaps between primary/overflow table pairs, and transitive inheritance chains.

> Note: Everything stems for a single architectural decision baked into the system which unfortunately, we cannot skip: ec_PropertyMap is a canonical, single-truth mapping where each (property, class-in-hierarchy) pair maps to exactly one column.
All the changes to the ClassMap in-memory model, SavePropertyMapVisitor, ClassMapColumnFactory, DbMapValidator, and the entire ViewGenerator SQL pipeline done in this POC all rely on this invariant.

The No-Remap POC investigates a fundamentally different approach: **skip all data movement entirely**, and instead:
1. Record where data *actually lives* (the original physical columns) in a new sidecar table (`ec_PropertyMap_Overrides`).
2. At schema-load time, remap the in-memory `ClassMap` to reflect the real storage layout.
3. At query time, detect when sibling classes in the same table hierarchy store the same logical property in *different* physical columns, and generate a `UNION ALL` query that reads from the correct column for each class partition.

### 2.1 New Schema Import Flag `SkipRemapping`

A new flag `SkipRemapping` was added to `SchemaManager::SchemaImportOptions`.  
All remap-sensitive code paths gate on this flag during schema import/upgrade.

> Note: This flag was added just to get the POC up, a different method must be decided if we are to go ahead with this approach in production. The reason being, once we switch to the no-remapping approach and import/upgrade a schema, there is no going back. The user cannot be allowed to do incremental schema changes while using this flag as an option as it will lead to data corruption.

---

### 2.2 New Sidecar Table `ec_PropertyMap_Overrides`

```sql
CREATE TABLE ec_PropertyMap_Overrides (
    ClassId       INTEGER NOT NULL REFERENCES ec_Class(Id),
    AccessString  TEXT    NOT NULL COLLATE NOCASE,
    ColumnId      INTEGER NOT NULL REFERENCES ec_Column(Id),
    PRIMARY KEY (ClassId, AccessString)
)
```

A new system table is added which stores **per-class column deviations**: cases where a class's property data physically resides in a column that differs from what `ec_PropertyMap` canonically records after the upgrade.

---

### 2.3 Remap Pipeline Bypassed

When `SkipRemapping` is active, the three core remap steps return `SUCCESS` immediately without doing any work:

---

### 2.4 Divergent-Column Query Generation

This is the **read-side counterpart** to the no-remap strategy.

#### Detection: `HasDivergentColumnAssignments()`
Walks all concrete class maps in the TPH table, compares their columns against the table-root class's assignments, and returns `true` if any mismatch is found, signaling a divergent column.
Only triggered for `TablePerHierarchy + ShareColumns` hierarchies (the only ones that use shared columns).

#### Rendering: `RenderDivergentEntityPartition()`
When divergence is detected, generates a `UNION ALL` where:
- Each **arm** is a full `SELECT` rendered for a single concrete class map (using its own column-remapped `PropertyMap`).
- Each arm is filtered with `WHERE [ECClassId] = <id>` (or `IN (...)` when multiple concrete classes share an identical SELECT body).
- Arms are de-duplicated: concrete classes whose rendered SQL is identical are merged into a single arm with an `IN` list.
- For `ONLY` queries (non-polymorphic), only the arm matching the queried class is emitted.

---

### 2.5 Tests - `SchemaRemapTests.cpp`

- **~30 existing tests** updated to use `SkipRemapping` so the test suite exercises the new no-remap approach.
- All are passing.

---

## 3. Pros

| # | Benefit |
|---|---|
| 1 | **No data movement during upgrade** - avoids full-table scans and row-by-row UPDATEs that can be slow for large files. |
| 2 | **No data-write lock required** - since rows are never touched, the upgrade footprint is limited to DDL + metadata writes. |
| 3 | **Entire remap pipeline complexity bypassed** - circular-remap sorting, cross-table overflow/primary pair detection, and the `RemappedColumnInfo` graph are no longer executed. |
| 4 | **Existing data is immutable** - no risk of data corruption from a failed mid-upgrade data migration. |
| 5 | **Correct read-back via UNION ALL** - pre-upgrade and post-upgrade rows are both served correctly by the divergent-partition query path. |
| 6 | **Smaller `ec_PropertyMap` churn** - the canonical system table is updated only once (new canonical column assignment), not cleaned + re-populated for every derived class. |
| 7 | **`ec_PropertyMap_Overrides` is an explicit contract** - the sidecar table makes the per-class column deviation a first-class, inspectable piece of metadata rather than implicit state in the remap pipeline. |

---

## 4. Cons

| # | Drawback |
|---|---|
| 1 | **Query performance regression for divergent hierarchies** - every polymorphic `SELECT` on a class whose hierarchy contains divergent column assignments becomes a `UNION ALL` instead of a single `SELECT`. For large hierarchies with many concrete subclasses this can multiply the query plan significantly. |
| 2 | **Divergence accumulates over repeated upgrades** - each no-remap upgrade adds more `ec_PropertyMap_Overrides` rows and potentially more UNION ALL arms. Without a consolidation step, the performance impact compounds. |
| 3 | **Two sources of truth for column assignments** - `ec_PropertyMap` and `ec_PropertyMap_Overrides` must be kept consistent. Any tool or code path that only reads one table gets an incomplete picture. |
| 4 | **Every column allocation decision now queries `ec_PropertyMap_Overrides`** - three new SQL queries per shared-column allocation add I/O overhead at schema-import time. |
| 5 | **In-memory `Remap()` is applied on every database open** - the `_Load()` phase re-reads `ec_PropertyMap_Overrides` and re-wires `PropertyMap` pointers each time the schema is loaded, adding startup cost proportional to the number of overrides. |
| 6 | **`SkipRemapping` is an opt-in flag** - it is the caller's responsibility to request it; the default path still uses the full remap pipeline. Meant to be used as a one-off call, forever switching to the no-remap after the initial use, unless a revert is made. |
| 7 | Debugging the data - since there are no remaps, after a few schema upgrades, the data is going to end up scattered. Refer to the example in section 4.1 below. |
---

### 4.1 Data readability problem example:

- Consider a very simple scenario where:
1. An initial schema state has a common base TPH class with shared columns with a set of child sibling classes all having an identical property "Prop1" kept in the shared column ps1.
2. A schema upgrade moves this property to a common base class but keeps all the child properties intact as overrides.
- With a remap, the properties get a new shared column "Prop1", and the data is moved over and consolidated.
- But without a remap, the new column "Prop1" will be created but the data will not be moved.
- The child class values for Prop1 will continue to exist in ps1, new inserts will also be in ps1, and Prop1 will continue to remain empty for the child classes.
- This will create difficulties when trying to reason out the data in the underlying db table/columns directly.
- When looking at raw data, we will always have to either use an ECSql which supports the no-remap code or look at the schema directly to reason out why the data is in this state.
- And this will only compound with time and further upgrades.

--- 
### 5.1 ECDb Profile Version Upgrade needed
The `ec_PropertyMap_Overrides` table is created in `CreateProfileTables()` (new databases only). There is **no profile version increment** and **no upgrader** to create this table in existing databases opened after the code change. Any attempt to run `SkipRemapping` against an older database that lacks the table will either fail silently or assert.

### 5.3 Multi-Column (Struct) Properties
Multi-column properties (e.g., `Point3d`, `struct`) generate multiple access strings (`Color.red`, `Color.green`, `Color.blue`). The override mechanism handles each sub-column individually, but the conflict-check and override-population SQL paths become proportionally more expensive and have been tested only through a small number of struct scenarios. A lot more testing is required.

### 5.4 ECSql UPDATE Semantics for Pre-Upgrade Rows
After a no-remap upgrade, a row inserted *before* the upgrade stores its property in the old physical column, while a row inserted *after* uses the new canonical column. The behavior of an ECSql `UPDATE` has not been explored in this POC. The update logic will also have to be updated to make it aware of possible divergent columns.

### 5.5 No Automatic Override Consolidation
There is no mechanism to "consolidate" overrides back into canonical `ec_PropertyMap` entries (i.e., perform a lazy, deferred remap). Divergence can only grow; it never shrinks without manual intervention.

### 5.6 Changeset / Briefcase / SchemaSync / Semantic Rebase Integration Not Verified
`ec_PropertyMap_Overrides` is a new system table. Its interaction with changeset push/pull, `SchemaSync`, semantic rebase has not been analyzed or tested. It is possible that overrides are lost or duplicated during schema-changeset apply.

### 5.7 Lots of edge cases still unexplored
Edge cases involving mixin queries and abstract-class queries against a divergent hierarchy have not been tested.

---

## 6. POC Validity Assessment

- The core concept is **viable**: data can survive a schema upgrade without being physically moved.
- A `UNION ALL` query strategy can **correctly reconstruct** property values for both pre-upgrade (old column) and post-upgrade (new column) rows from the same polymorphic query.
- The `ec_PropertyMap_Overrides` table provides a clean, inspectable mechanism for recording per-class column deviations.
- The majority of the existing `SchemaRemapTests` suite (30+ scenarios) passes under `SkipRemapping`, covering property promotion, base-class injection, sibling merging, mixin-to-base conversion, struct properties, overflow tables, relationship classes as well as some real-world Revit scenarios.

## 7. To-Do:

- Benchmark UNION ALL vs. single SELECT on realistic iModel workloads
- Verify and test ECSql `UPDATE` / `DELETE` semantics for pre-upgrade rows in divergent hierarchies
- Test `ec_PropertyMap_Overrides` behavior during changeset push/pull and rebase
- Extend `SearchPropertyMapVisitor` and other metadata tools to include override columns
- Test deep hierarchies and multiple sequential no-remap upgrades
- A lot of edge case testing

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -90,8 +90,7 @@
             "request": "launch",
             "program": "${env:OutRoot}Winx64/Product/ECDb-Gtest/ECDbTest.exe",
             "args": [
-                "--gtest_break_on_failure",
-                "--gtest_filter=InstanceRepositoryFixture.basic"
+                "--gtest_filter=SchemaRemapTestFixture.*"
             ],
             "stopAtEntry": false,
             "cwd": "${workspaceFolder}",
--- a/iModelCore/ECDb/ECDb/ClassMap.cpp
+++ b/iModelCore/ECDb/ECDb/ClassMap.cpp
@@ -459,7 +459,55 @@ BentleyStatus ClassMap::_Load(ClassMapLoadContext& ctx, DbClassMapLoadContext co
     if (m_propertyMaps.Insert(ecClassIdPropertyMap, 1) != SUCCESS)
         return ERROR;
 
-    return LoadPropertyMaps(ctx, dbLoadCtx);
+    if (SUCCESS != LoadPropertyMaps(ctx, dbLoadCtx))
+        return ERROR;
+
+    /*
+     * POC-No-Remap
+     * Apply per-class column overrides from ec_PropertyMap_Overrides when a property is promoted to a base class without physically moving data.
+     * The canonical ec_PropertyMap now points to a fresh base-class column, but instances of this class still have their data at the original column.
+     * Remapping each property map to the override column makes the in-memory ClassMap reflect the actual storage layout, so HasDivergentColumnAssignments can detect the divergence at query time.
+     */
+    if (m_ecdb.TableExists(TABLE_PropertyMap_Overrides))
+        {
+        const ECClassId classId = GetSchemaManager().GetClassId(m_ecClass);
+        CachedStatementPtr overrideStmt = m_ecdb.GetCachedStatement(
+            "SELECT o.AccessString, col.Name, tab.Name "
+            "FROM " TABLE_PropertyMap_Overrides " o "
+            "JOIN ec_Column col ON col.Id = o.ColumnId "
+            "JOIN ec_Table  tab ON tab.Id = col.TableId "
+            "WHERE o.ClassId = ?");
+
+        if (overrideStmt != nullptr && overrideStmt->BindId(1, classId) == BE_SQLITE_OK)
+            {
+            std::map<Utf8String, DbColumn*, CompareIUtf8Ascii> overrides;
+            while (overrideStmt->Step() == BE_SQLITE_ROW)
+                {
+                Utf8CP accessString = overrideStmt->GetValueText(0);
+                Utf8CP colName      = overrideStmt->GetValueText(1);
+                Utf8CP tableName    = overrideStmt->GetValueText(2);
+                if (const DbTable* table = GetSchemaManager().GetDbSchema().FindTable(tableName))
+                    {
+                    if (DbColumn* col = table->FindColumnP(colName))
+                        overrides.emplace(Utf8String(accessString), col);
+                    }
+                }
+
+            if (!overrides.empty())
+                {
+                SearchPropertyMapVisitor visitor(PropertyMap::Type::SingleColumnData);
+                m_propertyMaps.AcceptVisitor(visitor);
+                for (const PropertyMap* pm : visitor.Results())
+                    {
+                    auto it = overrides.find(pm->GetAccessString().c_str());
+                    if (it != overrides.end())
+                        pm->GetAs<SingleColumnDataPropertyMap>().Remap(*it->second);
+                    }
+                }
+            }
+        }
+
+    return SUCCESS;
     }
 
 //---------------------------------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


RohitPtnkr1996:
In this PR I have reverted a change done recently to support circular column remaps.
The no-remap approach handles it already.

rschili:
Whoa lots of info. My claude got totally thrown off when reading this together with my internal plan document and declared it like "this has major issues" when actually all of those were design choices that we knew would come with this.

One thing I wonder about: Given the new ec_ table and the profile bump, older code will both, not be able to write AND read with this, true? I was hoping we'd just arrive at "older code cannot import schemas anymore".

But apparently the old code uses "isMixin" to determine whether it should use the unions, and we cannot easily change that. So it would be possible to implement this without the new ec_ table, just using ec_PropertyMap, but old code still would not handle it correctly.

That is the new piece of information I learn from this. Everything else looks like it's what we expected. I have to think about this for a bit - right now I'm undecided if this changes plans or not.

RohitPtnkr1996:
I ran a bunch of performance tests to see the impact of this on polymorphics ECSqls.
The performance worsens significantly as the child class count increases.

1. ECSql Prepare time performance:

| Child class count | With remapped data (in us) | Without remapping data (in us) | Performance impact |
| --- | --- | --- | --- |
| 2 | 27303 | 29204 | 6% drop |
| 4 | 32743 | 37451 | 14% drop |
| 8 | 37281 | 75089 | 101% drop |
| 16 | 95668 | 179240 | 87% drop |

2. ECSql for a polymorphic `select all` across different row count:

| Total rows | With remapped data (in us) | Without remapping data | Performance impact |
| --- | --- | --- | --- |
| 800 | 9450 | 20141 | 113% drop |
| 1600 | 17383 | 47098 | 170% drop |
| 3200 | 37060 | 60893 | 64% drop |


@rschili

khanaffan:
If this is a significant change, I would prefer introducing JSON property mapping capabilities. Specifically, a class, a group of properties, or individual properties should be able to be marked for storage within a JSON column. I believe this can be implemented through feature gating rather than requiring a profile upgrade. This approach would allow us to isolate changes vertically across all layers instead of tying them to profile versioning.

Additionally, we should avoid using `CustomAttributes`. Instead, we can provide the necessary metadata through enhancements to the import schema to describe the mappings.

One important improvement we should prioritize is making the property map extensible. Currently, we lack a mechanism to safely ignore property mappings that the runtime does not recognize. Furthermore, many parts of our query generation logic assume a limited, predefined set of property mappings. This creates tight coupling between property mapping and query generation, which will become a challenge if we want to introduce extensibility.

For example, adding support for mapping a class, a set of properties, or individual properties to a JSON column would likely expose existing limitations in our codebase and highlight all the areas that need to support extensibility.

This change should also incorporate feature gating more broadly. The presence of a feature should determine whether a file is readable by older versions of the runtime. In some cases, files may remain backward compatible, while in others they may not be readable at all. Importantly, breaking changes should not be tied to profile version increments but instead to whether a feature is enabled.

Rather than incrementing profile versions, each feature should follow semantic versioning and provide a clear upgrade path. Eventually when feature matures it can be included in profile by default. Feature set control should be down in BeSQlite not just EC level but it could exist on SQLite or iModelPlatform level as well. e.g. new geometry stream change.

rschili:
@RohitPtnkr1996 oh those numbers are worse than I thought. Does that mean Mixins are like this, too? That sounds pretty bad...

@khanaffan agreed on both features, we have stories for that on the backlog, but I'd say they do not solve this problem here - they just make it happen less often. But they add complexity, like one additional map type and schema concepts that we have to support going forwards.

RohitPtnkr1996:
> One thing I wonder about: Given the new ec_ table and the profile bump, older code will both, not be able to write AND read with this, true? I was hoping we'd just arrive at "older code cannot import schemas anymore".

Well, this will be breaking when older code will try to open a newer imodel where remapping was skipped, as it will expect the data to have been moved and consolidated, and it also needs the updated ECSql query mechanism to handle divergent columns.

As for the import, we have used the ImportRequiresVersion CA as an import blocker in the past, but the issue is not with anything in the schemas that would require an import failure.

All ECDBs (old and new) will and should be able to import schemas, the change is in the underlying ECDb mappings.
The scary bit is the older runtimes having no clue about this with a newer iModel.

But it all eventually comes back to the columns sharing conflicts.
The minute we leave data spread across columns that was marked shared, we need some kind of UNION ALL logic to query it all, and this the older code does not have.

rschili:
@RohitPtnkr1996 Good research. The poor performance numbers along with the "older code cannot work with this" makes this unattractive. We'll talk about it with the others once you are back, but I believe we can mark the experiment as finished for now.

RohitPtnkr1996:
The main motivation for this exploration was to see if the remap surface can be reduced in such a way that maintenance would become easy for us.
This POC was a very straighforward, minimalist approach to achieve this without moving data around.
However, the ECSql performance impact is unacceptable.

Closing this until we find a better alternative to remapping.

PR #1455, opened 2026-06-08
Update crashpad build to use vcpkg

### LINKED ISSUE #1451: Update to latest crashpad and switch it to vcpkg
The current instructions for updating crashpad start out with building it on Windows using vcpkg. Update to the latest crashpad, but instead of adapting it to build using bmake, create a new mke file that uses the vcpkg build directly.

tcobbs-bentley:
This pull request refactors how the `crashpad` third-party library is built and integrated in the project, migrating its build process to use `vcpkg` for improved consistency and maintainability. It also updates documentation to provide clearer instructions for using and updating vcpkg-managed libraries. The changes simplify the build process, reduce custom scripts, and ensure all dependencies are handled in a unified way.

**Migration to vcpkg for Crashpad:**
- Refactored `crashpad` build to use vcpkg, replacing custom compilation rules and manual header management with standardized vcpkg scripts in `client.mke` and `handler.mke`. This includes automatic triplet selection, header delivery, and static library merging for downstream consumption.
- Removed the platform-specific, manual build scripts `client_linux.mki`, `client_windows.mki`, `handler_linux.mki`, and `handler_windows.mki`, as all logic is now centralized in the new vcpkg-based approach. Note: The windows scripts were never used in production.

**Build System Improvements:**
- Updated `Zlib.mke` to install vcpkg libraries in a subdirectory (`vcpkg_installed/compress/`) to avoid conflicts.
- Adjusted existing vcpkg scripts used with Zlib to default to using a local binary cache of builds.

**Documentation Updates:**
- Expanded `VCPKG.md` with detailed instructions on how vcpkg is used, how to update libraries, and best practices for integration, including environment variable usage and version pinning.Note: right now, crashpad is only built and deployed on Linux. That is expected change soon, but _this_ PR just deletes our local bmake source tree and uses vcpkg to build a version that we specify.

**Note:** This PR updates crashpad to use vcpkg, but does **not** expand crashpad use into platforms other than Linux. A Future PR will do that.


tcobbs-bentley:
@jffmarker You are the only person (other than Chuck) that I could find who modified the file in [this commit](https://github.com/iTwin/imodel-native/pull/1455/changes/f344a10432624a04ffa5b1473928b8a5252f3507). This is still WIP, but I wanted to make sure you didn't see any problems with the change. The change was recommended by GPT-5.5 after my update to use a vcpkg build of crashpad caused a test failure.

jffmarker:
I do not recall vcpkg integration with our build system. The primary concerns with using other people's build system are:

1. Do they use the same compile and link options as the rest of our build?
2. Do they support the myriad of platforms that we need to?
3. Do they build for debug and release as determined by our variables?
4. How are the binaries signed (Windows)?
5. How are the binaries versioned (Windows)?

A. Do you plan to test and support both the vcpkg and "legacy" builds? Why have both?
B. If you're only really asking about changes in CrashReportingLinuxCrashpad.cpp... honestly I don't have a very informed opinion. What was the test failure? I presume `CrashpadInfo::GetCrashpadInfo` returns non-null (as used in `SetCrashReportProperty`) even when called earlier in the initialization function?

aruniverse:
> I do not recall vcpkg integration with our build system. The primary concerns with using other people's build system are:
> 
> 1. Do they use the same compile and link options as the rest of our build?
> 2. Do they support the myriad of platforms that we need to?
> 3. Do they build for debug and release as determined by our variables?
> 4. How are the binaries signed (Windows)?
> 5. How are the binaries versioned (Windows)?
> 
> A. Do you plan to test and support both the vcpkg and "legacy" builds? Why have both? B. If you're only really asking about changes in CrashReportingLinuxCrashpad.cpp... honestly I don't have a very informed opinion. What was the test failure? I presume `CrashpadInfo::GetCrashpadInfo` returns non-null (as used in `SetCrashReportProperty`) even when called earlier in the initialization function?

cc @naveedkhan8067 @rolandas-rimkus  would you guys be able to help chime in here?

Whatever the answers are, we need to document this somewhere in this repo or in native-internal

tcobbs-bentley:
> I do not recall vcpkg integration with our build system. The primary concerns with using other people's build system are:
> 
> 1. Do they use the same compile and link options as the rest of our build?
> 2. Do they support the myriad of platforms that we need to?
> 3. Do they build for debug and release as determined by our variables?
> 4. How are the binaries signed (Windows)?
> 5. How are the binaries versioned (Windows)?
> 
> A. Do you plan to test and support both the vcpkg and "legacy" builds? Why have both? B. If you're only really asking about changes in CrashReportingLinuxCrashpad.cpp... honestly I don't have a very informed opinion. What was the test failure? I presume `CrashpadInfo::GetCrashpadInfo` returns non-null (as used in `SetCrashReportProperty`) even when called earlier in the initialization function?

Some notes.

Assuming that the vcpkg build of crashpad works out, the legacy build will be removed.

Regarding your numbered questions:

1. We have local triplet definitions that reproduce the options from before.
2. I believe that all platforms are supported, although I don't really know what all of those might be outside the add-on.
3. I'll have to investigate. *
4. I'll have to investigate. *
5. I'll have to investigate. *

\* So far, we have only submitted (WIP) PRs to update zlib/minizip and crashpad to use vcpkg. The former is statically built (both before and after), and the latter is only deployed on Linux (although there is infrastructure to build on Windows since that is where the last update was created). I don't think a debug build of crashpad has any value. I will investigate to see if the BeZlib build using vcpkg honors the debug setting. Because BeZlib is static, and crashpad is Linux-only, the binaries don't include Windows versioning.

aruniverse:
> The former is statically built (both before and after), and the latter is only deployed on Linux (although there is infrastructure to build on Windows since that is where the last update was created).

We should look to make sure it works on Windows and MacOS because we would like Sentry reporting for Studio based apps on those platforms. Previously we only had crsh reports for linux because thats what we deploy our remote cloud backends but we will be running on more platforms

https://docs.sentry.io/platforms/native/

tcobbs-bentley:
> > The former is statically built (both before and after), and the latter is only deployed on Linux (although there is infrastructure to build on Windows since that is where the last update was created).
> 
> We should look to make sure it works on Windows and MacOS because we would like Sentry reporting for Studio based apps on those platforms. Previously we only had crsh reports for linux because thats what we deploy our remote cloud backends but we will be running on more platforms
> 
> https://docs.sentry.io/platforms/native/

As far as I can tell, it builds fine on Windows (although there is a `%error` in the legacy mke file to prevent people from doing so without manually commenting out a line in the mke file, and I included that in my vcpkg mke file). I have not tried building on macOS, but I could do that.

Having said that, simply building crashpad doesn't do any good, since it also has to be connected to the executable. I will investigate how that is done on Linux currently, and then attempt to do the same on Windows and macOS. I feel that actually doing that should be a separate PR, though.

tcobbs-bentley:
> 3. Do they build for debug and release as determined by our variables?

I am updating the parent PR to pay attention to the `DEBUG` variable used by Bentley Build and build a debug build if it is set, and a release build otherwise. I will then update this PR to implement the same logic.

iModelCore/libsrc/crashpad/vcpkg.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "imodel-native-crashpad",
+  "version": "1.0.0",
+  "dependencies": [
+    {
+      "name": "crashpad",
+      "version>=": "2024-04-11#12"
```
naveedkhan8067: So will always move to the latest version by default? and not lock to a specific version?

naveedkhan8067:
> > 2. Do they support the myriad of platforms that we need to?

> 2. I believe that all platforms are supported, although I don't really know what all of those might be outside the add-on.

I agree with @jffmarker that before migration we should make sure that the required platforms are supported. Specifically looking at `crashpad`, their supported triplets seems to be `android | linux | osx | (windows & !uwp)`.

iModelCore/libsrc/crashpad/vcpkg-configuration.json
```diff
@@ -0,0 +1,7 @@
+{
+  "default-registry": {
+    "kind": "git",
+    "baseline": "81de6771512413aaf89ea77add5ad1fda126b9d0",
+    "repository": "https://github.com/microsoft/vcpkg"
```
naveedkhan8067, line 5: Will this repository be first cloned into our source and then be used further?

naveedkhan8067:
> > 4. How are the binaries signed (Windows)?
> > 5. How are the binaries versioned (Windows)?

> 4. I'll have to investigate. *
> 5. I'll have to investigate. *

We are currently using the signin tool for signing and I think the version we setup is constant accross all the dlls that we generate. It will be interesting to see that how vcpkg handle these.

iModelCore/libsrc/crashpad/client.mke
```diff
@@ -2,56 +2,90 @@
...
 #  See LICENSE.md in the repository root for full copyright notice.
 #---------------------------------------------------------------------------------------------
+# client.mke - Build crashpad client via vcpkg, deliver as CrashpadClient.
+#---------------------------------------------------------------------------------------------
 
-# Ignore warnings in third-party code.
-IS_THIRD_PARTY_LIBRARY = 1
-BUILD_WITH_C17 = 1
+# Toggle for rollout/testing:
+#   - default: build crashpad client via vcpkg (this file)
```
naveedkhan8067: Are you going to document this behaviour somewhere?

iModelCore/libsrc/crashpad/vcpkg.json
naveedkhan8067: Seems like overrides property handle this thats fine. I see `2024-04-11#13` is the latest one. Did we try that?

tcobbs-bentley: The override does indeed lock the version. I don't know why the `version>=` line is needed, but apparently it is, even with the override. Regarding the version, 13 was released after I did the initial work. I'll check updating to that version.

tcobbs-bentley: Updated to `#13`.

iModelCore/libsrc/crashpad/vcpkg-configuration.json
tcobbs-bentley: This controls the git registry for vcpkg ports. The corresponding vcpkg.json file in the same directory controls what port(s) to check out of that registry. When vcpkg is asked to install a port, it will download the port if it hasn't already done so, then build if it doesn't have the exact binary configuration already built in the local vcpkg install's binary cache. If the required build is already in the binary cache, it will use that instead of rebuilding, although this behavior can be disabled if it causes problems in the future.

iModelCore/libsrc/crashpad/client.mke
tcobbs-bentley: This PR is still in draft, but I suspect that the legacy fallback support will be removed before it leaves draft. The parent PR's zlib/minizip build is vcpkg-only, with no legacy fallback.

tcobbs-bentley:
> We are currently using the signin tool for signing and I think the version we setup is constant accross all the dlls that we generate. It will be interesting to see that how vcpkg handle these.

Even when using vcpkg, we are still building locally, not downloading builds that other people have created. So we should be able to sign and version the binaries just like we already do. (vcpkg is not integrated into bb, but is being made use of via `.sh`/`.bat` files callled from `.mke`/`.mki` files.)

iModelCore/libsrc/crashpad/vcpkg-configuration.json
naveedkhan8067: ok but this file is now already in the source code present in `main` branch: https://github.com/iTwin/imodel-native/blob/main/iModelCore/libsrc/compress/vcpkg-configuration.json Why we have to add it again? I would suggest to keep one file but move it from `/libsrc/compress` to `/libsrc/` and reference it form here for the ones where it is required.

tcobbs-bentley: Unfortunately, vcpkg requires this file to be in the same directory as the vcpkg.json file. I tried to configure it to share onc vcpkg-configuration.json file and learned that that is not possible. In theory we could use symlinks, but I think that causes problems on Windows.

tcobbs-bentley: I will investigate if it is better to have a single vcpkg.json with all of our libraries in it instead of keeping the libraries separated like they are now.

tcobbs-bentley: Unfortunately, combining everything won't work, because it would also require any custom settings for the separate libraries to be combined. We already have different custom settings for both zlib and crashpad.

aruniverse (commented):
🟡 MEDIUM — Test now passes even when StartHandler() fails; root cause of vcpkg test failure unaddressed (DgnDb.test.ts)

 The PR author noted the annotation move was "recommended by GPT-5.5 after a test failure." The annotation loop was moved before all guards, so testCrashReportingConfig passes even when the crashpad handler binary is missing or StartHandler()
 fails. The actual root cause (likely: CrashpadHandler binary path changed with vcpkg layout) was never fixed. The test is now weaker — it validates that properties are set in memory, not that the handler is running.

 Suggested fix: Identify why StartHandler() was failing (check handler binary path, vcpkgHandlerExe expansion), fix it, then consider whether the annotation loop needs to stay early or can return to its original position.

tcobbs-bentley:
> 🟡 MEDIUM — Test now passes even when StartHandler() fails; root cause of vcpkg test failure unaddressed (DgnDb.test.ts)
> 
> The PR author noted the annotation move was "recommended by GPT-5.5 after a test failure." The annotation loop was moved before all guards, so testCrashReportingConfig passes even when the crashpad handler binary is missing or StartHandler() fails. The actual root cause (likely: CrashpadHandler binary path changed with vcpkg layout) was never fixed. The test is now weaker — it validates that properties are set in memory, not that the handler is running.
> 
> Suggested fix: Identify why StartHandler() was failing (check handler binary path, vcpkgHandlerExe expansion), fix it, then consider whether the annotation loop needs to stay early or can return to its original position.

This is fixed in three ways by [this commit](https://github.com/iTwin/imodel-native/pull/1455/changes/a7dd2504a781490e2ff030bc630f23b03618379f):
* The moved initialization code was moved back where it started.
* The mke was updated to use chmod on the crash handler executable.
* InitializeCrashReporting now returns a bool, and if that is false, a JS exception is thrown.

aruniverse:
Picking up the review thread — summarizing what needs to be resolved before this can merge, including direct answers to @jffmarker's Jun 9 concerns.

---

### Build blockers (will fail as-is)

**1. Wrong library filenames in `client.mke`**

The four library variable assignments use a `vcpkg_` prefix that doesn't match what vcpkg installs:

```makefile
# current (wrong)
vcpkgCrashpadClient = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_client$(stlibext)

# should be (verify exact names against actual vcpkg output)
vcpkgCrashpadClient = $(vcpkgLibDir)$(stlibprefix)crashpad_client$(stlibext)
```

This applies to all four names in both the `%if defined (DEBUG)` and `%else` blocks. The Linux `merge_static_libs.sh` call and the Windows `lib.exe` call will both fail with "file not found." Worth also confirming that `crashpad_client_common` is actually a separate archive produced by vcpkg `2024-04-11#13`, and whether any additional libraries (`crashpad_minidump`, `crashpad_snapshot`, etc.) need to be included in the merge.

**2. `%error` gate in `handler.mke` still fires on Windows**

The "Windows fixes" commit added the correct `vcpkgHandlerExe` path and platform guards below lines 11–13, but the `%error` itself was never removed. It runs at parse time, before any of the code below it, so all of the Windows handler logic added in the latest commit is currently unreachable.

**3. `crashpad.PartFile.xml` — `CrashpadHandler` lists `OnlyPlatforms="linux*,x*"` but Windows builds still hard-error**

Until item 2 is resolved, any Windows build that hits this Part gets a fatal preprocessor error. The fix depends on the Windows scope decision (see below).

---

### Questions that need a decision

**Q1: Is Windows crashpad delivery in scope for this PR?**

`CrashReportingWindows.cpp` still uses Breakpad — the new Windows vcpkg triplets and handler paths don't connect to any Windows runtime code yet. Two options:

- **In scope:** Remove the `%error`, wire up the Windows crashpad client, and keep `OnlyPlatforms="linux*,x*"`.
- **Deferred:** Change `CrashpadHandler` to `OnlyPlatforms="linux*"`, drop or stub the Windows paths from `client.mke`/`handler.mke`, and leave a clear TODO.

Either is fine, but the current state is inconsistent — Windows build infrastructure exists but hard-errors, and the runtime never calls into crashpad on Windows.

**Q2: Linux release handler path — `tools/crashpad/crashpad_handler` vs `tools/crashpad_handler`**

The debug path is `debug/tools/crashpad_handler` (flat) and the release path is `tools/crashpad/crashpad_handler` (extra `crashpad/` subdirectory). Can you confirm this matches the actual vcpkg install layout for `2024-04-11#13`? If the release binary lands at the flat path the `~linkfile` step silently fails.

---

### Answers to @jffmarker's Jun 9 concerns

The triplet files address these — spelling out the answers explicitly so they're on record:

**1. Same compile and link options as the rest of the build?**

The custom triplets forward the flags that matter for ABI compatibility:
- Linux: `-D_FILE_OFFSET_BITS=64` to match the previous build's large-file ABI mode (documented in `x64-linux.cmake`)
- Windows: `VCPKG_CRT_LINKAGE dynamic` to match the rest of the build's runtime linkage
- `vcpkg_run_install.sh` propagates `CC`, `CXX`, `AR`, `RANLIB`, `NM`, `STRIP` so vcpkg uses the project's own toolchain on Linux

General optimization/warning flags are not forwarded — this is the standard pattern for third-party libs in this repo (the same approach used for zlib/minizip via vcpkg).

**2. Platform support?**

Same scope as the legacy build: Linux x64 and Windows x64, via custom triplets in `iModelCore/libsrc/crashpad/triplets/`. macOS, iOS, and Android are excluded via `OnlyPlatforms="linux*,x*"` in `crashpad.PartFile.xml` — no change from before.

**3. Debug and release builds?**

- Linux: vcpkg builds both debug and release. `client.mke` and `handler.mke` both have `%if defined (DEBUG)` branches selecting `debug/lib/` vs `lib/` artifacts.
- Windows: forced to release-only via `VCPKG_BUILD_TYPE release` in the triplet. The bmake Windows pipeline links against the release CRT — mixing CRT versions (or iterator-debug-level mismatch) causes linker errors. `NODEFAULTLIB` could theoretically work around this, but crashpad calls debug-only CRT functions when compiled debug, so it doesn't in practice. The rationale is documented in `x64-windows-static.cmake`.

**4 & 5. Signing and versioning (Windows)?**

Neither applies yet since the Windows handler isn't being delivered from this PR. When it is: vcpkg builds locally (it doesn't download pre-built binaries), so the output `.exe` goes through the same BentleyBuild signing and versioning pipeline as other binaries. The version resource and `.rc` file will need to be added at that point — it's deferred work with no impact on the current Linux-only scope.

**Question A (legacy vs. vcpkg):** The legacy `.mki` files were fully deleted in commit `54ae6e2`. There is no dual-path.

**Question B (scope beyond `CrashReportingLinuxCrashpad.cpp`):** Correct, the PR is much broader — it's a complete replacement of the crashpad build system (4 `.mki` files and ~970 lines of hand-built GN-style build logic replaced by a vcpkg manifest and two rewritten `.mke` files).

---

### Documentation

`iModelCore/libsrc/crashpad/readme.md` currently describes the 2019 depot_tools workflow and a 2023 vcpkg experiment — neither reflects how the build works now. Before merging, this should be replaced (or substantially updated) to cover:

- How crashpad is now built: vcpkg manifest at `vcpkg.json`, pinned to a specific version via `overrides`, baseline locked in `vcpkg-configuration.json`
- How to update the version: bump `version>=` and `overrides` in `vcpkg.json`, update the baseline hash in `vcpkg-configuration.json`
- How custom triplets work and where to add one for a new platform (`iModelCore/libsrc/crashpad/triplets/`)
- How binary caching works and what `VCPKG_BINARY_SOURCES` / `VCPKG_DEFAULT_BINARY_CACHE` do
- The general vcpkg integration pattern shared with `iModelCore/libsrc/compress/` (both use `vcpkg.mki`, `vcpkg_run_install.sh/bat`, and separate `vcpkg_installed/<name>/` directories)
- Runtime prerequisites: `LINUX_MINIDUMP_ENABLED`, `MINIDUMP_UPLOAD_URL`

The `iModelCore/libsrc/vcpkg.mki` shared helper is also undocumented — a short comment block at the top explaining the triplet-selection logic and how to add a new platform would help future contributors.

tcobbs-bentley:
> ### Build blockers (will fail as-is)
> **1. Wrong library filenames in `client.mke`**
> 
> The four library variable assignments use a `vcpkg_` prefix that doesn't match what vcpkg installs:
> 
> ```makefile
> # current (wrong)
> vcpkgCrashpadClient = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_client$(stlibext)
> 
> # should be (verify exact names against actual vcpkg output)
> vcpkgCrashpadClient = $(vcpkgLibDir)$(stlibprefix)crashpad_client$(stlibext)
> ```
> 
> This applies to all four names in both the `%if defined (DEBUG)` and `%else` blocks. The Linux `merge_static_libs.sh` call and the Windows `lib.exe` call will both fail with "file not found." Worth also confirming that `crashpad_client_common` is actually a separate archive produced by vcpkg `2024-04-11#13`, and whether any additional libraries (`crashpad_minidump`, `crashpad_snapshot`, etc.) need to be included in the merge.

This is wrong. I just checked my Linux build output, and it contains `libvcpkg_crashpad_base.a`, `libvcpkg_crashpad_client.a`, `libvcpkg_crashpad_client_common.a`, and `libvcpkg_crashpad_util.a`. In Windows, the files are `vcpkg_crashpad_base.lib`, etc. None of those libraries exist without the `vcpkg` prefix. Furthermore, the build succeeds, and (after uncommenting a line in one of the tests), crash dumps are successfully created.

tcobbs-bentley:
> ### Build blockers (will fail as-is)
> **2. `%error` gate in `handler.mke` still fires on Windows**
> 
> The "Windows fixes" commit added the correct `vcpkgHandlerExe` path and platform guards below lines 11–13, but the `%error` itself was never removed. It runs at parse time, before any of the code below it, so all of the Windows handler logic added in the latest commit is currently unreachable.
> 
> **3. `crashpad.PartFile.xml` — `CrashpadHandler` lists `OnlyPlatforms="linux*,x*"` but Windows builds still hard-error**
> 
> Until item 2 is resolved, any Windows build that hits this Part gets a fatal preprocessor error. The fix depends on the Windows scope decision (see below).

This is intentional. While Windows builds were used before to do initial work on compiling, it was not built in any standard build strategies, and **not allowed** to be built without commenting out the error line. I maintained that behavior, while also locally verifying that the Windows part still builds with the error line commented out. Also, even though Windows builds of CrashPadHandler are allowed in crashpad.PartFile.xml (as before), that part doesn't get built as part of any standard build strategies.

Note that a followup PR will add Windows and macOS support.

tcobbs-bentley:
> ### Questions that need a decision
> **Q1: Is Windows crashpad delivery in scope for this PR?**
> 
> `CrashReportingWindows.cpp` still uses Breakpad — the new Windows vcpkg triplets and handler paths don't connect to any Windows runtime code yet. Two options:
> 
> * **In scope:** Remove the `%error`, wire up the Windows crashpad client, and keep `OnlyPlatforms="linux*,x*"`.
> * **Deferred:** Change `CrashpadHandler` to `OnlyPlatforms="linux*"`, drop or stub the Windows paths from `client.mke`/`handler.mke`, and leave a clear TODO.
> 
> Either is fine, but the current state is inconsistent — Windows build infrastructure exists but hard-errors, and the runtime never calls into crashpad on Windows.

Windows Crashpad is not in the scope of this PR. There is a [followup PR](https://github.com/iTwin/imodel-native/pull/1466) (still in draft) that will take care of that, as well as macOS support.

> 
> **Q2: Linux release handler path — `tools/crashpad/crashpad_handler` vs `tools/crashpad_handler`**
> 
> The debug path is `debug/tools/crashpad_handler` (flat) and the release path is `tools/crashpad/crashpad_handler` (extra `crashpad/` subdirectory). Can you confirm this matches the actual vcpkg install layout for `2024-04-11#13`? If the release binary lands at the flat path the `~linkfile` step silently fails.

Both files exist in the release build, and they seemingly only differ with some relative paths inside the executables. The debug build does not include the `crashpad/crashpad_handler` executable. According to GPT-5.5's examination of the two executables, the one in directly the `tools` directory is the intended one, so I will update the `mke`.

> ### Documentation
> `iModelCore/libsrc/crashpad/readme.md` currently describes the 2019 depot_tools workflow and a 2023 vcpkg experiment — neither reflects how the build works now. Before merging, this should be replaced (or substantially updated) to cover:
> 
> * How crashpad is now built: vcpkg manifest at `vcpkg.json`, pinned to a specific version via `overrides`, baseline locked in `vcpkg-configuration.json`
> * How to update the version: bump `version>=` and `overrides` in `vcpkg.json`, update the baseline hash in `vcpkg-configuration.json`
> * How custom triplets work and where to add one for a new platform (`iModelCore/libsrc/crashpad/triplets/`)
> * How binary caching works and what `VCPKG_BINARY_SOURCES` / `VCPKG_DEFAULT_BINARY_CACHE` do
> * The general vcpkg integration pattern shared with `iModelCore/libsrc/compress/` (both use `vcpkg.mki`, `vcpkg_run_install.sh/bat`, and separate `vcpkg_installed/<name>/` directories)
> * Runtime prerequisites: `LINUX_MINIDUMP_ENABLED`, `MINIDUMP_UPLOAD_URL`

The environment variables are documented at the top of the file. ~~I **did** add a new section on how one would go about updating the crashpad version now (which is very small, since the whole point was to make that easy). I did not document how everything works, since the old documentation was focused on letting people know how to update in the future.~~ I accidentally added the new section in the wrong branch. I will fix that.

I will add a link to `../VCPKG.md`, and make sure that file contains instructions that span all vcpkg libraries (like `VCPKG_BINARY_SOURCES`.

> The `iModelCore/libsrc/vcpkg.mki` shared helper is also undocumented — a short comment block at the top explaining the triplet-selection logic and how to add a new platform would help future contributors.

I will add documentation for that in VCPKG.md.

tcobbs-bentley:
@aruniverse I believe that I have addressed the issues raised in your last comment.

aruniverse:
Two small items for the final review pass:

---

**@jffmarker** — all five of your Jun 9 concerns have now been answered in this thread (see the Jun 23 comment above). Summarising so everything is in one place for your re-read:

1. **Same compile/link options?** The custom triplets in `iModelCore/libsrc/crashpad/triplets/` forward the ABI-critical flags: `-D_FILE_OFFSET_BITS=64` on Linux to match the previous build's large-file ABI mode, `VCPKG_CRT_LINKAGE dynamic` on Windows, and `vcpkg_run_install.sh` propagates `CC`, `CXX`, `AR`, `RANLIB` etc. from the project's own toolchain.

2. **Platform support?** Linux x64 in this PR via a custom triplet. Windows and macOS are in the follow-up PR #1466.

3. **Debug and release builds?** Both `client.mke` and `handler.mke` have `%if defined (DEBUG)` branches that select `debug/lib/` vs `lib/` artifacts on Linux. Windows is release-only by design (documented in the triplet file — mixing CRT versions causes linker errors).

4. **Binary signing (Windows)?** Deferred to PR #1466 when the Windows handler is actually delivered. vcpkg builds locally, so the output goes through the existing BentleyBuild signing pipeline unchanged.

5. **Binary versioning (Windows)?** Same — deferred to PR #1466.

Would appreciate your sign-off if you're comfortable with those answers, or please raise any follow-up questions here.

---

**Separate note for PR #1466:** The `%error` message in `handler.mke` currently reads:

```
%error Need to comment out this line to make it compile
```

This reads like an oversight rather than intentional deferral. When that gate is removed in #1466, please update it (or any replacement) to something like:

```
%error Windows crashpad handler delivery is deferred — see PR #1466
```

Not a blocker for this PR — just worth capturing so it doesn't confuse a future reviewer looking at the file in isolation.

aruniverse (approved):

naveedkhan8067 (approved):

jffmarker:
> Two small items for the final review pass:
> 
> **@jffmarker** — all five of your Jun 9 concerns have now been answered in this thread (see the Jun 23 comment above). Summarising so everything is in one place for your re-read:
> 
> 1. **Same compile/link options?** The custom triplets in `iModelCore/libsrc/crashpad/triplets/` forward the ABI-critical flags: `-D_FILE_OFFSET_BITS=64` on Linux to match the previous build's large-file ABI mode, `VCPKG_CRT_LINKAGE dynamic` on Windows, and `vcpkg_run_install.sh` propagates `CC`, `CXX`, `AR`, `RANLIB` etc. from the project's own toolchain.
> 2. **Platform support?** Linux x64 in this PR via a custom triplet. Windows and macOS are in the follow-up PR [WIP: Enable crashpad on macOS and Windows #1466](https://github.com/iTwin/imodel-native/pull/1466).
> 3. **Debug and release builds?** Both `client.mke` and `handler.mke` have `%if defined (DEBUG)` branches that select `debug/lib/` vs `lib/` artifacts on Linux. Windows is release-only by design (documented in the triplet file — mixing CRT versions causes linker errors).
> 4. **Binary signing (Windows)?** Deferred to PR [WIP: Enable crashpad on macOS and Windows #1466](https://github.com/iTwin/imodel-native/pull/1466) when the Windows handler is actually delivered. vcpkg builds locally, so the output goes through the existing BentleyBuild signing pipeline unchanged.
> 5. **Binary versioning (Windows)?** Same — deferred to PR [WIP: Enable crashpad on macOS and Windows #1466](https://github.com/iTwin/imodel-native/pull/1466).
> 
> Would appreciate your sign-off if you're comfortable with those answers, or please raise any follow-up questions here.

Honestly I think the answer to (3) means that (1) is inaccurate... when we compile for "debug", we don't use the debug CRT... so if that's what debug builds of vcpkg are doing, then we are not accurately reflecting compile and link options. When we build for debug, we (at least) disable optimizations and enable assertions. I'm not sure which other flags differences there are off the top of my head.

With (2), (4), and (5) largely deferred to another PR, the remaining focus is ensuring we compile with the same options so that when linked together we get compatible results. I suspect we can't guarantee we're using the same options because we're mixing build systems. Perhaps the best we can hope for is your testing that ensures the builds work and are compatible. Have you tried using the debug and release outputs with our code to ensure that they work?

tcobbs-bentley:
> Honestly I think the answer to (3) means that (1) is inaccurate... when we compile for "debug", we don't use the debug CRT... so if that's what debug builds of vcpkg are doing, then we are not accurately reflecting compile and link options. When we build for debug, we (at least) disable optimizations and enable assertions. I'm not sure which other flags differences there are off the top of my head.
> 
> With (2), (4), and (5) largely deferred to another PR, the remaining focus is ensuring we compile with the same options so that when linked together we get compatible results. I suspect we can't guarantee we're using the same options because we're mixing build systems. Perhaps the best we can hope for is your testing that ensures the builds work and are compatible. Have you tried using the debug and release outputs with our code to ensure that they work?

The debug build of the native addon does unit tests, which pass. One of the unit tests (testCrashReportingConfig) initializes crash reporting, sets some config settings, and verifies that those settings are set. That passes for both release and debug builds.

Additionally, there is a commented out line of code in that test that triggers a crash. Uncommenting that line in this branch, the crash dump is successfully generated by the vcpkg-built crashpad in both release and debug builds. I'd be happy to perform more tests if it is felt that they would be valuable.

jffmarker (approved):

DanRod1999 (approved):

tcobbs-bentley:
@khanaffan @bbastings Even though I have approvals from some codeowners, it's still insisting that I need codeowner review. Can you take a look?

Note: if you collapse the crashpad/vendor deletion, there aren't very many files to review.

PR #1463, opened 2026-06-15
Fix deadlock between concurrent query worker and main thread

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9436

## Summary

Fixes a deadlock between the main thread (running `ECSqlStatement::Step`) and a concurrent query worker thread (`QueryAdaptorCache::TryGet` / `ECSqlStatement::Prepare`). This is a classic AB-BA lock ordering violation between the **ECDb impl mutex** and **SQLite's internal mutex**.

Fixes https://github.com/iTwin/itwinjs-backlog/issues/2113

## Root Cause

In `QueryAdaptorCache::TryGet`, when `usePrimaryConn == false`:
1. Worker acquires ECDb mutex (spin-lock)
2. Calls `ECSqlStatement::Prepare(primaryDb.Schemas(), workerDb, ecsql)` — schema resolution via `SchemaPersistenceHelper::GetClassId` calls `sqlite3_bind_text` on the **primary** database → needs primary SQLite mutex

Meanwhile on the main thread:
1. `sqlite3_step()` holds the primary SQLite mutex
2. A custom scalar function (`ExtractInstFunc`) fires during the step
3. It calls `InstanceReader::GetOrAddClass` → `SchemaManager::Dispatcher::GetIterable` → needs ECDb mutex

**Lock ordering violation:**
- Thread A: Primary SQLite mutex → ECDb mutex
- Thread B: ECDb mutex → Primary SQLite mutex

## Fix

For worker connections (`usePrimaryConn == false`), prepare using the worker's own ECDb instance instead of the primary's SchemaManager. The worker already has its own SchemaManager and SQLite connection (synced via `SyncAttachDbs`), so it can resolve schemas independently without touching the primary's locks.

## ECSqlReader concurrent-query performance: A vs B (final)

**A** = original addon (`9cd010a6`) · **B** = new addon with serialization fix (`4742fee6`).
Throughput in queries/sec (higher = better), mean of 2 runs/phase; range spans maxConc 16 and 64.
All runs: 0 errors, `rows=847828`.

| workers | A | B (fixed) | Δ |
|--:|--:|--:|--:|
| 1 | 182–183 | 182–184 | ~0% |
| 2 | 307–313 | 306–309 | ≤1.4% |
| 4 | 410–427 | 415–426 | ≤1.1% |
| 8 | 228–239 | 230–238 | ≤1.1% |


iModelCore/ECDb/PublicAPI/ECDb/ConcurrentQueryManager.h
```diff
@@ -530,6 +532,9 @@ struct ConcurrentQueryMgr final {
         Config& SetWorkerThreadCount(uint32_t workerThreadCount) { m_workerThreadCount = workerThreadCount; return *this;}
         Config& SetRequestQueueSize(uint32_t requestQueueSize) { m_requestQueueSize = requestQueueSize; return *this;}
         Config& SetIgnorePriority(bool ignorePriority) { m_ignorePriority = ignorePriority; return *this;}
+        //! @deprecated No longer consulted. Worker connections always prepare on their own ECDb
+        //! to avoid an AB-BA lock ordering deadlock with the primary connection. Kept for
+        //! backward-compatible config serialization only; calling it has no effect on behavior.
         Config& SetDoNotUsePrimaryConnToPrepare(bool doNotUsePrimaryConnToPrepare) { m_doNotUsePrimaryConnToPrepare = doNotUsePrimaryConnToPrepare; return *this;}
```
rschili, line 541: So we are flipping the default here, and have no way to use the old default.
Is there a scenario where callers would want to prepare on the primary connection? Pending Txns, in-flight modifications? I guess for that we still have usePrimaryConnection, so it is probably fine?

iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
```diff
@@ -1352,4 +1352,142 @@ TEST_F(ConcurrentQueryFixture, ImportSchemaShouldClearQueryCache) {
...
+
+    // Start the concurrent query manager
+    ConcurrentQueryMgr::WithInstance(m_ecdb, [&](auto& mgr) {
+        // Launch a thread that simulates "Thread A" - stepping through extract_inst on main connection
+        auto mainStepThread = std::thread([&]() {
+            ECSqlStatement stmt;
+            // SELECT $ triggers extract_inst for each row (polymorphic - reads full instance)
+            auto status = stmt.Prepare(m_ecdb, "SELECT $ FROM ts.Base");
+            if (status != ECSqlStatus::Success) {
+                workerError = true;
```
rschili: suggested correct name `mainThreadPrepareFailed`
This has nothing to do with a workerError, I think.

iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
```diff
@@ -1352,4 +1352,142 @@ TEST_F(ConcurrentQueryFixture, ImportSchemaShouldClearQueryCache) {
...
+        for (int i = 0; i < numConcurrentQueries; ++i) {
+            auto ecsql = SqlPrintfString("SELECT $ FROM ts.Base WHERE [Value] >= %d", i);
+            auto req = ECSqlRequest::MakeRequest(ecsql.GetUtf8CP());
+            futures.push_back(mgr.Enqueue(std::move(req)));
+        }
+
+        // Wait for the main step thread to finish. If the deadlock regresses, this join
+        // blocks forever and the test harness kills the run as a timeout (which is the
+        // intended failure signal for this reproduction).
+        mainStepThread.join();
```
rschili, line 1522: If this deadlock comes back, the worker threads get stuck holding locks. That means this join(), and the manager cleanup afterwards, both hang forever. The only sign of failure is the whole test run timing out, which is slow and easy to blame on the wrong thing. A watchdog makes it fail fast with a clear message instead. abort() is heavy-handed, but stuck threads can't be joined, so it's the only way out.

 ```cpp
 // Before mainStepThread.join():
 std::atomic<bool> testCompleted{false};
 auto watchdog = std::thread([&]() {
     const auto deadline = std::chrono::steady_clock::now() + 60s;
     while (!testCompleted.load()) {
         if (std::chrono::steady_clock::now() > deadline) {
             fprintf(stderr, "DEADLOCK DETECTED: main step vs worker prepare did not "
                             "complete within 60s (itwinjs-backlog#2113 regression).\n");
             fflush(stderr);
             std::abort();
         }
         std::this_thread::sleep_for(100ms);
     }
 });
watchdog.detach();
```

Then set `testCompleted = true;` as the last line inside the `WithInstance` lambda (after the result-collection loop) so the watchdog exits cleanly on success.

File: iModelCore/ECDb/PublicAPI/ECDb/ConcurrentQueryManager.h (hunk 1 above)
Comment (khanaffan) [reply iModelCore/ECDb/PublicAPI/ECDb/ConcurrentQueryManager.h:541]: This is a flag we had for testing we never used it. Now its default. Though it has nothing to do with primary connection queries.
The primary connection prepare and execute happen in same call and it never put in queue. The flag here effect none primary connections. Previously they use primary to prepare the statement but use a second sqlite connection to executed it. Now its prepare and execute by second sqlite connection. 

```
      if (runnableReq->GetRequest().UsePrimaryConnection()) {
                ExecuteSynchronously(conns, std::move(runnableReq));
            } else {
```

It tried find other solution but the issue is pretty complicated. One way is two do this complex refactor where we share SchemaManager with secondary connection in two phase locking. That is a big refactor. It will atleast let us shared schema/map cache between primary and secondary connection.

File: iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp (hunk 2 above)
Comment (khanaffan) [reply iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp]: fixed

File: iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp (hunk 3 above)
Comment (khanaffan) [reply iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp:1522]: Done in `dac3e0a05`. Added the watchdog just before `mainStepThread.join()`; it also guards the result-collection loop since `testCompleted` is set at the end of the `WithInstance` lambda.

One change from your snippet: instead of `watchdog.detach()` I set `testCompleted = true` and `watchdog.join()` at the end of the block. `testCompleted` is captured by reference, so a detached watchdog could outlive it and read a destroyed atomic — joining guarantees it stops before the local goes out of scope. Success path exits immediately; a regression still aborts with the diagnostic.

**Comment by aruniverse** [comment]
## Adversarial Review — PR #1463

Three parallel reviewers (Correctness & Deadlock Safety · Safety/Memory/Lifetime · Tests & API Completeness) each read the diff and source files independently. Findings synthesized below.

---

### 🔴 Fixes worth doing now

**1. New latent deadlock: `ConnectionCache::Interrupt()` ↔ `GetSchemaSourceConnection()`**

`Interrupt()` holds `m_mutex` for its **entire** drain spin-loop (`cpp:531–546`). The new `GetSchemaSourceConnection()` also acquires `m_mutex` (`cpp:512`). A worker blocked in `GetSchemaSourceConnection()` can't release its `shared_ptr<CachedConnection>`, so `use_count` never drops to 1, so `Interrupt()` never exits. Classic AB-BA with the very same mutex.

`ConnectionCache::Interrupt()` has no caller in this repo (only `InterruptIf` is called), so it's currently dead code. However the method is public and could be called from external bindings — and since this PR's stated purpose is to fix deadlocks, shipping a new one is worth flagging.

Suggested fix — use `try_lock` with fallback to `nullptr`, letting the worker fall through to its own connection:

```cpp
CachedConnection* ConnectionCache::GetSchemaSourceConnection() {
    if (!m_primaryDb.IsDbOpen())
        return nullptr;
    if (m_schemaConn != nullptr)      // fast path: already created, no lock needed
        return m_schemaConn.get();
    if (!m_mutex.try_lock())          // Interrupt may hold m_mutex; fall back gracefully
        return nullptr;
    recursive_guard_t lock(m_mutex, std::adopt_lock);
    if (m_schemaConn == nullptr)
        m_schemaConn = CachedConnection::Make(*this, SCHEMA_SOURCE_CONN_ID);
    return m_schemaConn.get();
}
```

Alternatively, eagerly initialize `m_schemaConn` in the `ConnectionCache` constructor, eliminating the lock on the read path entirely.

---

**2. `GetFuture()` — residual dangling `queue` pointer**

The PR correctly fixed the `this` capture (old UAF). However `queue = &GetQueue()` is a raw pointer into `ConcurrentQueryMgr::Impl::m_queue`. If the manager is destroyed while a `Future` is still alive and `future.Cancel()` is called afterward, `queue->CancelRequest(id)` dereferences freed memory.

The fix was right to move away from capturing `this`, but the `queue` pointer has its own lifetime concern. Suggest using a `weak_ptr<Impl>` in the cancel callback, or an explicit documented precondition that `Cancel()` after manager destruction is UB.

---

**3. `CancelAfterCompletionIsSafe` test doesn't exercise the bug it claims to guard**

`ClearRequest()` (which destroys the `RunnableRequest`) runs on the **worker thread** *after* `SetResponse()` satisfies the promise. `future.Get()` unblocks as soon as the promise is fulfilled — before `ClearRequest()` runs. Calling `Cancel()` immediately after `Get()` almost certainly fires while the request object is still alive, so the old `[&]` capture UAF would not crash without ASan.

```cpp
auto r = future.Get();
ASSERT_EQ(r->GetStatus(), QueryResponse::Status::Done);
// Allow the worker to call ClearRequest() (destroying the request) before Cancel().
// Without this sleep, Cancel() likely fires while the request is still alive, masking
// the use-after-free without ASAN. With the fix, the callback captures id+queue by value.
std::this_thread::sleep_for(std::chrono::milliseconds(20));
future.Cancel();
```

---

### 🟡 Optional improvements

**4. `SchemaLockGuard` inline struct — prefer `std::unique_lock` with `adopt_lock`**

The inline `struct SchemaLockGuard` is safe (construction can't throw; destructs on all paths including exceptions from `Prepare`), but it's non-idiomatic — the codebase uses `BeMutexHolder` elsewhere. Consider:
```cpp
std::unique_lock<BeMutex> schemaGuard(schemaMutex, std::adopt_lock);
```

**5. `DeadlockReproduction` — race window can be missed by a fast scheduler**

If the OS runs `mainStepThread` to completion before any worker starts a `Prepare()`, no concurrent contention occurs and the test passes even with the fix removed. A latch or brief sleep at the start of `mainStepThread` (before the first `Step()`) would force interleaving and make detection more reliable.

**6. `SCHEMA_SOURCE_CONN_ID = UINT16_MAX` — add overflow guard**

Worker connections are created with id `(uint16_t)m_conns.size() + 1`; at `poolSize >= 65534` the cast wraps and collides with `SCHEMA_SOURCE_CONN_ID`. Default pool is `min(4, hw_concurrency)` so unreachable in practice, but a `BeAssert(m_conns.size() < UINT16_MAX - 1)` in `GetConnection()` would make the invariant explicit.

---

### ⚪ Not worth changing

- **`imodel_*` functions registered on schema-source conn**: harmless — connection is never stepped for user queries.
- **`InvalidateSeekPos` targeting schema-source ECDb's InstanceReader**: correct in practice — the InstanceReader's row-id comparison in `Seek()` prevents stale state regardless. A clarifying comment would be nice but is not required.
- **Watchdog `abort()` bypassing GTest failure reporting**: the `fprintf` diagnostic before `abort()` is sufficient for CI; a deadlock makes a clean GTest failure impossible anyway.
- **`Config::Get()` snapshot semantics change**: intentional and correct; no callers relied on live-view semantics.
- **Spin-wait backoff on schema mutex**: performance/fairness note, out of scope for a bug-fix PR.

---

### Overall

The core deadlock fix (AB-BA elimination via shared schema-source connection) is sound — the new lock ordering is genuinely disjoint from the primary's locks (confirmed by tracing `Prepare(schemaDb.Schemas(), workerConn.GetDb(), ...)` through the dispatch path). The secondary fixes (`ProgressHandlerGuard`, `InvalidateSeekPos`, `GetFuture` capture, exception safety, `Config` mutex) are also correct. The main concern is item 1: the `Interrupt()` + `GetSchemaSourceConnection()` interaction introduces a new deadlock on the exact path this PR is trying to clean up.

rschili (approved):

rschili (changes requested):
Holding off approval until the comments from Arun's PR are picked up

soham-bentley:
Looks good to me overall.....but as Arun's comment mentions maybe we can remove the `ConnectionCache::Interrupt()` method as its not used anywhere.

khanaffan:
Thanks for the thorough adversarial review @aruniverse, and @soham-bentley for the follow-up. Addressed in 55794998c8.

### 🔴 Fixed

**#1 — `Interrupt()` ↔ `GetSchemaSourceConnection()` AB-BA:** Rather than papering over it with `try_lock`, I removed `ConnectionCache::Interrupt(bool, bool)` entirely. It had **no callers** in the repo (only `InterruptIf` is used), so it was dead code carrying a real future foot-gun on the exact path this PR cleans up. Deleting it eliminates the AB-BA outright. This also resolves @soham-bentley's suggestion to drop the unused method. ✅

**#3 — `CancelAfterCompletionIsSafe` didn't exercise the bug:** Good catch — `Get()` unblocks when the promise is satisfied, *before* the worker runs `ClearRequest()`. Added a 20 ms sleep between `Get()` and `Cancel()` so the request is actually destroyed before the cancel fires, exercising the post-destruction path. (The callback already captures `id` + `queue` by value, never the request.) ✅

### 🟡 Fixed

**#6 — `SCHEMA_SOURCE_CONN_ID` overflow:** Added `BeAssert(m_conns.size() < UINT16_MAX - 1)` in `GetConnection()` to make the invariant explicit. ✅

### Intentionally not changing in this bug-fix PR

**#2 — `GetFuture()` `queue` pointer lifetime:** The manager is per-`ECDb` and is drained + torn down via `Shutdown()` (through `ClearECDbCache`) before it is destroyed; calling `Cancel()` after the manager itself is gone isn't a supported lifecycle. Moving the callback to a `weak_ptr<Impl>` is reasonable hardening but out of scope here — happy to file a follow-up.

**#4 — `SchemaLockGuard` vs `unique_lock`/`adopt_lock`:** Agreed it's non-idiomatic, but it's correct (construction can't throw, destructs on every path including an exception from `Prepare`). Leaving it as a minimal local RAII guard.

**#5 — `DeadlockReproduction` scheduler race:** The test steps 400 polymorphic rows while 40 concurrent prepares run, which widens the interleaving window in practice. A latch would make it deterministic, but I'd rather not over-fit the repro.

⚪ items — agreed, no change.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -78,36 +84,100 @@ std::shared_ptr<CachedQueryAdaptor> QueryAdaptorCache::TryGet(Utf8CP ecsql, bool
...
+                // exception thrown from schema loading (which would otherwise leak the lock and wedge
+                // all future worker prepares).
+                struct SchemaLockGuard final { BeMutex& m_mutex; ~SchemaLockGuard() { m_mutex.unlock(); } } guard{ schemaMutex };
+                // Suppress logging on this attempt: a failure is expected for queries targeting an
+                // attached table space (the schema-source connection has none) or for genuinely invalid
+                // ECSQL; both are retried against the worker's own connection below, which reports
+                // errors normally.
+                status = newCachedAdaptor->GetStatement().Prepare(schemaDb.Schemas(), m_conn.GetDb(), ecsql, false);
+            }
+            preparedAgainstSchemaSource = status == ECSqlStatus::Success;
```
aruniverse, line 169: The shutdown check runs only inside the spin loop, before `schemaMutex` is acquired. If `Shutdown()` arrives while `Prepare` is in flight inside the `SchemaLockGuard` block, the worker doesn't notice until `Prepare` returns — not a deadlock, but it can stall `Shutdown()` under concurrent prepare load.

Suggest adding immediately after this line:
```cpp
if (queue.GetState() == RunnableRequestQueue::State::Stop) {
    status = ECSqlStatus::Error;
    isShutDownInProgress = true;
    ecsql_error = "Queue is shut down, cannot go ahead with the request.";
    return nullptr;
}
```
Mirrors the same pattern used in the primary-conn spin path just above.

iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
```diff
@@ -1352,4 +1377,340 @@ TEST_F(ConcurrentQueryFixture, ImportSchemaShouldClearQueryCache) {
...
+// Reproduces the deadlock from https://github.com/iTwin/itwinjs-backlog/issues/2113
+// Thread A (main): sqlite3_step holds primary SQLite mutex, ExtractInstFunc fires,
+//   calls InstanceReader::GetOrAddClass -> Dispatcher::GetIterable -> needs ECDb mutex.
+// Thread B (worker): QueryAdaptorCache::TryGet holds ECDb mutex, ECSqlStatement::Prepare
+//   resolves schemas via SchemaPersistenceHelper::GetClassId -> sqlite3_bind_text on
+//   primary db -> needs primary SQLite mutex.
+// Without the fix, this test deadlocks (times out). With the fix, the worker uses its
+// own ECDb for schema resolution, avoiding the AB-BA lock ordering violation.
+//+---------------+---------------+---------------+---------------+---------------+------
+TEST_F(ConcurrentQueryFixture, DeadlockReproduction_MainStepVsWorkerPrepare) {
```
aruniverse, line 1396: The watchdog is well-constructed (atomic `testCompleted`, `watchdog.join()` not `detach()`, 60 s timeout). ✅

One note for future maintainers: the deadlock window is **probabilistic** — it relies on the OS scheduler interleaving `mainStepThread`'s `sqlite3_step` loop with a worker `Prepare`. On a single-core or lightly loaded machine the interleaving may never occur and the test can pass on a regressed build without the watchdog firing. The test is most reliable on multi-core CI. A comment to that effect would set accurate expectations without requiring any structural change.

aruniverse (commented):
Reviewed the current branch state through commit `55794998`. The fix is structurally sound.

**Lock-ordering analysis.** The schema-source connection is physically disjoint from the primary (separate `ECDb` instance, separate SQLite handle, separate ECDb impl mutex). Acquiring `schemaMutex` on it during `Prepare` cannot form an AB-BA cycle with the main thread because:
- Main thread: `SQLite_primary` (held by `sqlite3_step`) → needs `ECDb_primary` (via `ExtractInstFunc`)
- Worker: `schemaMutex` = `ECDb_schemaConn` (held during `Prepare`) → needs `SQLite_schemaConn` (not `SQLite_primary`)

No shared lock between the two paths. ✅

**All prior review threads resolved.** The watchdog (`testCompleted` atomic + `watchdog.join()`), the variable rename (`mainThreadPrepareFailed`), and the `@deprecated` annotation on `SetDoNotUsePrimaryConnToPrepare` are all in place and correct.

**`Interrupt` removal** (`55794998`) is safe — confirmed no callers remain in the codebase.

**One suggestion** (see inline on `TryGet`) — the shutdown check inside the spin loop isn't re-evaluated after the `SchemaLockGuard` is acquired. Minor but worth a one-liner.

@rschili — the TypeScript-side concerns that prompted the CHANGES_REQUESTED have been addressed: the PRAGMA behavior was fixed in `f2774337` (native) and the corresponding TS commit, and the infrastructure follow-up is tracked as iTwin/itwinjs-core#9451. Would appreciate a re-review when you get a chance.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
khanaffan: Good catch — done in `ddfdb85de8`. Added the `queue.GetState() == Stop` re-check immediately after acquiring `schemaMutex` inside the `SchemaLockGuard`, before `Prepare`. The guard releases the mutex on the early return, and it mirrors the primary-conn spin path, so `Shutdown()` no longer stalls behind an in-flight worker prepare.

iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
khanaffan: Added a `NOTE` to the test header documenting that the deadlock window is probabilistic (scheduler-dependent, most reliable on multi-core CI), as suggested — no structural change. Done in `ddfdb85de8`. Thanks.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
aruniverse: Done in `ddfdb85de8` — placed correctly **inside** `SchemaLockGuard` before `Prepare` is called, so the mutex is released via RAII on the early return. Mirrors the primary-conn spin path exactly. ✅

iModelCore/ECDb/Tests/NonPublished/ConcurrentQueryTest_V2.cpp
aruniverse: Added at lines 1391–1394 in `ddfdb85de8`. ✅

aruniverse (commented):
Both open threads are resolved in `ddfdb85de8`:

- **Shutdown re-check** — added inside `SchemaLockGuard` before `Prepare`, RAII mutex release covers the early-return path. ✅
- **Probabilistic NOTE** — added at lines 1391–1394 of `ConcurrentQueryTest_V2.cpp`. ✅

All prior review threads (Copilot, rschili's three inline comments) were already resolved in earlier commits. The lock-ordering analysis, RAII guards, and synchronous primary-conn routing all look correct.

@rschili — everything raised in your CHANGES_REQUESTED (and the TypeScript-side concerns that prompted it) has been addressed. The native pragma fix landed in `f2774337`, the TS-side PRAGMA test is corrected, and the infrastructure follow-up is tracked as iTwin/itwinjs-core#9451. This branch is ready for a re-review when you have a moment.

aruniverse (approved):

rschili (approved):

PR #1466, opened 2026-06-15
Enable crashpad on macOS and Windows

### LINKED ISSUE #1465: Add support for crashpad on macOS and Windows
With Studio officially supporting macOS and Windows, having crashpad Sentry reporting there is advantageous, so we want to add support for it on both operating systems. (Currently, it is only supported on Linux.)

Note: this relies on [updating crashpad to build using vcpkg](https://github.com/iTwin/imodel-native/issues/1451).

tcobbs-bentley:
This pull request extends Crashpad-based crash reporting support to Windows and macOS, in addition to Linux, and improves the build and configuration experience across platforms. It also introduces environment variable and documentation updates, as well as some build system and code refactoring for better cross-platform compatibility.

**Crashpad cross-platform support and configuration:**

* Crashpad is now built and enabled by default on Windows, Linux, and macOS. The environment variable `IMODEL_ADDON_MINIDUMP_ENABLED` is now supported on all platforms to enable crash reporting, with the legacy `LINUX_MINIDUMP_ENABLED` remaining for Linux only. The handler executable is delivered and version-stamped on Windows, and the build scripts have been updated for cross-platform delivery. 

* The crash reporting implementation was refactored and renamed from `CrashReportingLinuxCrashpad.cpp` to `CrashReportingCrashpad.cpp`, with platform-specific logic and environment variable handling. The code now uses helper functions for environment variable retrieval and file path handling, and enables Crashpad based on the new variables.

* Windows no longer uses Breakpad, using Crashpad instead.

**Build and launch configuration updates:**

* The VSCode launch configuration was updated to use `IMODEL_ADDON_MINIDUMP_ENABLED` instead of the Linux-specific variable, ensuring consistent crash reporting behavior in development and testing across platforms. 

* The Crashpad static client and handler are now built for macOS in addition to Linux, and the build scripts handle platform-specific static library merging and handler delivery.

**Documentation and developer experience:**

* The Crashpad readme and developer documentation were updated to reflect the new cross-platform support and environment variable usage. Build troubleshooting steps and common errors were clarified, and a note was added to clarify that vcpkg is now used for Crashpad builds.


aruniverse (commented):
Review of cross-platform correctness, with all prior #1455 thread items checked. One blocker and a few optional cleanups inline below. Overall the PR is in good shape — jffmarker's concerns 4 & 5 (signing and versioning) are now fully addressed via `WinSignAndVersionRule.mki`.

iModelJsNodeAddon/CrashReportingCrashpad.cpp
```diff
@@ -43,14 +81,38 @@ bool JsInterop::InitializeCrashReporting(CrashReportingConfig const& cfg)
...
+#endif
     
     //.............................................................................................
     // WIP: solely rely on configuration for crash path.
     BeFileName dbPathW = cfg.m_crashDir;
     if (0 == dbPathW.size())
+        {
+#ifdef BENTLEY_WIN32
+        wchar_t tempPath[MAX_PATH];
+        DWORD len = GetTempPath2W(MAX_PATH, tempPath);
```
aruniverse, line 104: **Blocker — `GetTempPath2W` requires Windows 10 Build 19041+ (Version 2004)**

`GetTempPath2W` was introduced in May 2020 (Win10 2004 / Server 2022). It is absent on Windows 10 LTSC 2019, Server 2019, Server 2016, and earlier. Because this is a **static import** (not a `GetProcAddress` call), the Windows module loader will fail to resolve it at DLL load time on any system that does not export this symbol — the `C:\temp\crash` fallback below is never reached; the entire `.node` addon fails to load.

Please replace with `GetTempPathW`, available since Windows XP and a drop-in replacement for this use case (the only behavioural difference is how it handles impersonation, which is irrelevant for crash-dump directory selection):

```cpp
// Before:
DWORD len = GetTempPath2W(MAX_PATH, tempPath);
// After:
DWORD len = GetTempPathW(MAX_PATH, tempPath);
```

iModelJsNodeAddon/CrashReportingCrashpad.cpp
```diff
@@ -107,17 +175,23 @@ bool JsInterop::InitializeCrashReporting(CrashReportingConfig const& cfg)
     args.push_back("--no-rate-limit"); // don't restrict to once per-hour
     // args.push_back("--no-upload-gzip"); // don't compress HTTP request (for debugging purposes)
 
-    CrashpadClient client;
+    static CrashpadClient client;
+    // @todo: Decide if we want to pass true for restartable and asynchronous_start on macOS and maybe Windows.
```
aruniverse: **Optional — resolve this `@todo` before merge**

The code already passes `restartable=true, asynchronous_start=true` on non-Linux platforms (lines 191–192), so the decision has effectively been made. The `@todo` should either be accepted and removed, or replaced with a concrete follow-up issue link — leaving it implies the current values are still in question.

For the record: `asynchronous_start=true` means there is a narrow window between `StartHandler()` returning and the handler fully connecting where a crash will not produce a dump. Acceptable tradeoff for initial deployment, but worth documenting explicitly as a known limitation rather than an open question.

iModelCore/libsrc/crashpad/client.mke
```diff
@@ -56,21 +56,31 @@ always:
 #   Merge vcpkg crashpad component archives into one output expected by downstream.
 #--------------------------------------------------------------------------------------
 CrashpadClientOut      = $(o)$(stlibprefix)$(appName)$(stlibext)
-%if defined (DEBUG)
+%if $(TARGET_PLATFORM) == "Windows"
+    vcpkgCrashpadClient    = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_client$(stlibext)
+    vcpkgCrashpadCommon    = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_client_common$(stlibext)
+    vcpkgCrashpadUtil      = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_util$(stlibext)
+    vcpkgCrashpadBase      = $(vcpkgLibDir)$(stlibprefix)vcpkg_crashpad_base$(stlibext)
+    vcpkgMigOutput         = $(vcpkgLibDir)$(stlibprefix)mig_output$(stlibext)
```
aruniverse: **Optional — `vcpkgMigOutput` is dead code in the Windows variable block**

`vcpkgMigOutput` is assigned here (and in the DEBUG and release Windows branches on lines 70 and 76), but the Windows `lib` merge command below never includes it — correctly, because `mig_output` is a macOS-specific Mach IPC stub that does not exist in the Windows vcpkg output.

The library list is correct as-is; the assignment is just confusing. Suggest either removing `vcpkgMigOutput` from all three Windows branches, or adding a short comment like `# mig_output is macOS only` to make the omission intentional.

iModelCore/libsrc/WinSignAndVersionRule.mki
```diff
@@ -0,0 +1,57 @@
...
+%if !defined (winSignVersionInternalName)
+	winSignVersionInternalName = $(winSignVersionProductName)$(exeext)
+%endif
+
+%if !defined (winSignVersionOriginalName)
+	winSignVersionOriginalName = $(winSignVersionProductName)$(exeext)
+%endif
+
+always:
+	$(InjectVersionCmd) "$(winSignVersionTarget)" $(STD_VERSION_OPTS) -k"FileDescription=$(winSignVersionFileDescription)" -k"InternalName=$(winSignVersionInternalName)" -k"OriginalFilename=$(winSignVersionOriginalName)" -k"ProductName=$(winSignVersionProductName)"
```
aruniverse, line 53: **Optional — consider stamping `CompanyName` and `LegalCopyright`**

This line stamps `FileDescription`, `InternalName`, `OriginalFilename`, and `ProductName`. For a vcpkg-delivered binary built from Google's crashpad source, the `CompanyName` and `LegalCopyright` fields in the version resource will carry whatever Google's CMake embedded — not Bentley branding.

Not required for valid Authenticode signing, so not a blocker. But if these fields matter for product consistency, they can be added:
```
-k"CompanyName=Bentley Systems, Incorporated"
-k"LegalCopyright=Copyright (c) Bentley Systems, Incorporated. All rights reserved."
```
Check whether `$(STD_VERSION_OPTS)` already injects these before adding them explicitly.

aruniverse:
**No automated end-to-end crash dump verification**

The `testCrashReportingConfig` test validates that `InitializeCrashReporting` succeeds, the handler is found, and crash properties round-trip correctly — but it does not verify that an actual crash produces a `.dmp` file on disk.

Per the commit history, macOS crash dump capture was manually verified by temporarily uncommenting a `signal(3)` line in `DgnDb.test.ts`, but that test was not committed. Windows crash dump capture has not been explicitly called out as verified.

Not a blocker for this PR, but worth tracking as a follow-up: a CI smoke test that raises a signal, waits briefly, and checks for a dump file in `cfg.m_crashDir` would give ongoing confidence that the full end-to-end path works across all three platforms.

iModelJsNodeAddon/CrashReportingCrashpad.cpp
tcobbs-bentley: The change was already made to do that in non-Linux, but I forgot to remove the todo comment. I have now removed the todo comment.

iModelCore/libsrc/crashpad/client.mke
tcobbs-bentley: Fixed.

iModelCore/libsrc/WinSignAndVersionRule.mki
tcobbs-bentley: The copyright is making it in (verified in my build). The company name isn't, but this matches other executables that are stamped as part of the build (such as iTwinPlatform.dll). I believe that what we have currently is correct.

iModelJsNodeAddon/CrashReportingCrashpad.cpp
aruniverse: I honestly don't know if this is worth it

aruniverse (approved):

jffmarker (approved):
Looks like it goes out of its way to version and sign on Windows, which was a chief concern of mine.

tcobbs-bentley:
> Looks like it goes out of its way to version and sign on Windows, which was a chief concern of mine.

Correct. And the commit that moves that into a dedicated mki in libsrc makes it so that doing so will be easy for future vcpkg libraries.

PR #1470, opened 2026-06-17
Fix `-Wincompatible-pointer-types` error in `checkNoActiveStatements` (clang 19+)

rschili:
The debug-only `checkNoActiveStatements` helper in `bentley-sqlite.c` iterates prepared
statements with `sqlite3_next_stmt`, assigning between `Vdbe*` (the internal struct) and
`sqlite3_stmt*` (its opaque public handle). These are the same object, but distinct C types.

Older compilers treated the mismatch as a warning (suppressed by the build's `-w` flag).
Clang 19 and newer promote `-Wincompatible-pointer-types` to an error by default, which
`-w` does not downgrade, so debug builds fail to compile on recent toolchains (hit on
Fedora with clang 22).

Added explicit casts at the API boundary, mirroring SQLite's own internal casting between
`Vdbe*` and `sqlite3_stmt*`. No behavior change - the code is debug-only
(`#if !defined(NDEBUG)`) and the casts are between identical underlying objects.


aruniverse (approved):

iModelCore/BeSQLite/SQLite/bentley-sqlite.c
```diff
@@ -80,9 +80,11 @@ int checkNoActiveStatements(sqlite3* db)
     if (0 == db->nVdbeActive)
         return SQLITE_OK;
 
-    for (stmt = db->pVdbe; stmt != NULL; stmt = sqlite3_next_stmt(db, stmt))
+    // A prepared statement is internally a Vdbe; sqlite3_stmt is just its opaque public handle.
+    // The casts cross that public/private boundary and mirror what SQLite does internally.
+    for (stmt = db->pVdbe; stmt != NULL; stmt = (Vdbe*)sqlite3_next_stmt(db, (sqlite3_stmt*)stmt))
         {
-        if (stmt->eVdbeState != VDBE_RUN_STATE)
+        if (stmt->eVdbeState == VDBE_RUN_STATE)
```
khanaffan, line 87: I do not understand isn't this function look for non active statements as name say. Basically held but not running.

rschili: I also wondered about this. Copilot brought it up in its PR review, it's unrelated to my change.
The method "Checks that there are no active statements"

Reading the method I think this change is correct. The SUCCESS return happens further above, and this check here is just about which warning gets logged, and Copilot thinks it was reversed. From the looks of it I think it's right.
But I'm not even sure why we are doing this additional check, when the one further above should be enough.

rschili: I mean, the log message it sends literally says "Active statement" and VBDE_RUN_STATE is exactly that state. So I think the change to this check is correct.

When the code gets here it is already always returning ERROR.

RohitPtnkr1996 (approved):

soham-bentley (approved):

khanaffan (approved):

iModelCore/BeSQLite/SQLite/bentley-sqlite.c
khanaffan: As I understand we looking for cached statements which are not finalized. You will not actually find any active statement when closing db. 

This log is very useful to diagnose issues specially in production. Can you add a unit test just to confirm previous behavior is maintained.

PR #1476, opened 2026-06-24
ECSQL: support IS / IS NOT operator between operands (null-safe comparison)

### LINKED ISSUE #9439: ECSQL: support `IS` / `IS NOT` operator between operands (null-safe comparison)
### Use case

When comparing two ECSQL values that may be `NULL` (for example two GUID columns), SQL's three-valued logic makes `=` / `<>` awkward: `NULL = NULL` evaluates to _unknown_ (effectively false) and `NULL <> 'foo'` is also _unknown_. To express a **null-safe** comparison in ECSQL today you have to write a verbose boolean expansion, e.g.:

```sql
WHERE (depends.lastKnownGeometricGuid <> gm.geometryGuid)
   OR (depends.lastKnownGeometricGuid IS NULL AND gm.geometryGuid IS NOT NULL)
   OR (depends.lastKnownGeometricGuid IS NOT NULL AND gm.geometryGuid IS NULL)
```

This pattern is common in change-detection / synchronization queries across BIS schemas (it surfaced while writing such a query in iTwin/bis-schemas#727). A first-class null-safe comparison operator would make these queries far more readable and less error-prone, and benefits anyone writing ECSQL that compares nullable columns.

### Proposed solution

Extend the ECSQL `IS` / `IS NOT` operators — which today only accept `NULL`, the truth literals `TRUE`/`FALSE`/`UNKNOWN`, and the `IS (ClassName)` type predicate on the right-hand side — to also accept a second operand (a property/value expression), matching SQLite's null-safe comparison operators:

```sql
col1 IS col2
col1 IS NOT col2
```

`IS` behaves like equality but treats `NULL` as a comparable value (`IS NOT` is its negation):

```sql
NULL IS NULL      -- TRUE
NULL IS 'foo'     -- FALSE
'foo' IS 'foo'    -- TRUE
'foo1' IS 'foo2'  -- FALSE
```

The motivating query then simplifies to:

```sql
WHERE depends.lastKnownGeometricGuid IS NOT gm.geometryGuid
```

Suggested behavior:

- The `NULL` literal may appear on either side (`col IS NULL`, `NULL IS col`), preserving today's `IS [NOT] NULL` behavior.
- For multi-column operands (`Point2d`/`Point3d`, navigation properties) the comparison expands column-wise, consistent with `=` / `<>`: `IS` joins the per-column comparisons with `AND`, `IS NOT` joins them with `OR`.
- Incompatible operand types (e.g. a `string` vs a `Point3d`) are rejected at prepare time.
- The existing `IS NULL`, `IS [NOT] TRUE/FALSE/UNKNOWN`, and `IS (ClassName)` forms remain unchanged.

### Alternatives considered

The verbose `(<>) OR (IS NULL / IS NOT NULL …)` expansion shown above works but is hard to read, easy to get subtly wrong, and has to be repeated for every nullable column pair. Coalescing to a sentinel (`IFNULL`/`COALESCE`) is fragile — it requires a value that can never legitimately appear in the data and can defeat index usage.

### References

Originated from a discussion on iTwin/bis-schemas#727:
https://github.com/iTwin/bis-schemas/pull/727#issuecomment-4769700572

### Willingness to contribute

- [x] I am willing to write the code or documentation for this feature.

### AI acknowledgment

- [x] I used AI to generate this feature request.
- [x] (If the above is checked) I have reviewed the AI-generated content before submitting.

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9440

closes: https://github.com/iTwin/itwinjs-core/issues/9439

This pull request adds support for using the `IS` and `IS NOT` operators between two operands in ECSQL, extending their previous usage (which was limited to comparisons with `NULL`, boolean literals, or type predicates). Now, these operators can be used for null-safe comparisons between properties, including multi-column types like points and navigation properties, with appropriate column-wise expansion. The changes include parser, lexer, and documentation updates, as well as comprehensive tests to verify the new behavior.

**ECSQL Syntax and Semantics Enhancements:**

* Added support for `IS` and `IS NOT` between two operands (e.g., `prop1 IS [NOT] prop2`), with null-safe comparison semantics consistent with SQLite. For multi-column operands, `IS` joins per-column comparisons with `AND`, and `IS NOT` with `OR`.
* Updated the ECSQL version to `2.0.4.0` in documentation, code, and tests to reflect the new syntax support.


aruniverse (commented):
Adversarial review (automated) — 4 inline findings below, ranging from grammar-correctness question to missing conflict-count baseline. No blockers on this side of the split; the corresponding itwinjs-core PR has the BLOCKER (missing column in Comparison.ecsql.md).

iModelCore/ECDb/ECDb/ECSql/Parser/sqlbison.y
```diff
@@ -973,12 +986,34 @@ type_list:
         }
     ;
 
+/* 'opt_only' is inlined here (rather than 'opt_only table_node') so that a bare class name
+   shifts SQL_TOKEN_NAME directly. Otherwise the empty 'opt_only' reduction loses a
+   shift/reduce race against value_exp's property_path in the 'X IS (...)' context, which
+   would shadow this type predicate with a parenthesized value expression. */
```
aruniverse: **Grammar comment vs. test assertion — please reconcile**

The comment here says that a bare name inside `(...)` after `IS` — e.g. `S1 IS (S2)` — always routes through `type_predicate` (because `type_list_item` now shifts `SQL_TOKEN_NAME` directly, beating the empty `opt_only` → `value_exp` race).

Yet both the C++ test (`ECSqlStatementTests.cpp`) and the TypeScript test (`ECSqlStatement.test.ts`) assert that `S1 IS (S2)` returns **2 rows** — value-expression semantics, not a class-predicate result.

Two possible explanations:
1. `(S2)` still reduces via `truth_value → value_exp → value_exp_primary → '(' value_exp_commalist ')'`, not via `type_predicate`, and this comment overstates what the fix does.
2. `(S2)` does parse as a `type_predicate` node, and `ParseValueExp` in `ECSqlParser.cpp` silently falls back to treating the name as a property reference when the class lookup fails.

Please add one sentence clarifying which path actually fires for a bare unqualified property name in parentheses, so maintainers know what to expect if an identifier is both a valid class name and a property name.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -2434,20 +2434,49 @@ BentleyStatus ECSqlParser::ParseSearchCondition(std::unique_ptr<BooleanExp>& exp
...
+                }
+
+            // X IS [NOT] NULL|TRUE|FALSE|UNKNOWN|(ClassName) : the left side is a boolean predicate.
+            std::unique_ptr<BooleanExp> op1 = nullptr;
+            if (SUCCESS != ParseSearchCondition(op1, lhsNode))
+                return ERROR;
+
             std::unique_ptr<ValueExp> truthValueExp = nullptr;
-            if (SUCCESS != ParseTruthValue(truthValueExp, parseNode->getChild(3/*truth_value*/)))
+            if (SUCCESS != ParseValueExp(truthValueExp, rhsNode))
```
aruniverse, line 2488: **`ParseTruthValue` silently replaced by `ParseValueExp` for the legacy `IS NULL / IS (ClassName)` path — confirm no dead code**

The previous code called `ParseTruthValue` here; this PR silently swaps it for `ParseValueExp`. The regression test (`ECClassId IS (ts.Foo)` → `ec_cache_ClassHierarchy`) confirms it still works, but:

1. Is `ParseTruthValue` still defined in this file? If so, it is now dead code and should be removed (or at minimum marked `[[maybe_unused]]` / commented out with a TODO).
2. The PR description does not call out this swap, making it invisible to reviewers who only read the description. A one-line note in the commit message or PR description body would help.

iModelCore/ECDb/Scripts/BuildECSqlParser.sh
```diff
@@ -99,15 +99,23 @@ echo "Compiling Lexical Analyzer ..."
...
+# Note: this is an inherited (OpenOffice/LibreOffice) SQL grammar whose many
+#       shift/reduce and reduce/reduce conflicts are resolved by bison's default
+#       rules and are expected. They cannot be declared away with %expect /
+#       %expect-rr (%expect-rr is GLR-only, and %expect makes the reduce/reduce
+#       conflicts a hard error on this LALR parser), so the warnings are
+#       suppressed here without altering the generated parser tables.
 # ---------------------------------------------------------------------------
 echo "Compiling Parser ..."
-"$BISON" -l -pSQLyy -bSql \
+"$BISON" -l -pSQLyy -bSql -Wno-conflicts-sr -Wno-conflicts-rr \
```
aruniverse, line 118: **Consider snapshotting the known-good conflict count as a comment**

The comment block above this line explains well why `-Wno-conflicts-sr -Wno-conflicts-rr` are necessary (inherited OpenOffice/LibreOffice grammar, `%expect`/`%expect-rr` not usable on an LALR parser). The grammar grew by ~3 rules and ~11 states in this PR.

Without a numeric baseline, any *future* grammar change that silently introduces a new unintended conflict will go unnoticed — the warnings are already suppressed. Suggestion:

```bash
# Known-good conflict counts as of ECSQL 2.0.4.0 (add IS/IS NOT between operands):
#   shift/reduce: N   reduce/reduce: M
# (run: bison --verbose ... 2>&1 | grep -E 'conflict')
```

Even a human-readable note establishes a baseline for the next person changing the grammar.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
```diff
@@ -2929,7 +2930,7 @@ TEST_F(ECSqlUpdatePrepareTests, WhereBasics)
     //case insensitive tests
     EXPECT_EQ(ECSqlStatus::Success, Prepare("UPDATE ecsql.P SET I=10 WHERE B = NULL OR b = NULL"));
     EXPECT_EQ(ECSqlStatus::Success, Prepare("UPDATE ecsql.P SET I=10 WHERE i>=:myParam"));
-    EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("UPDATE ecsql.P SET I=10 WHERE I IS 123"));
+    EXPECT_EQ(ECSqlStatus::Success, Prepare("UPDATE ecsql.P SET I=10 WHERE I IS 123"));
```
aruniverse, line 2945: **Missing explanatory comment on `InvalidECSql → Success` flip (UPDATE / DELETE blocks)**

The corresponding SELECT block at line 1722 has this comment added just before the flipped assertion:
```cpp
// 'IS' / 'IS NOT' accept any value expression on either side (null-safe comparison), not just NULL.
```
The UPDATE `WhereBasics` (this line), UPDATE `MiscellaneousWithALL`, DELETE `WhereBasics`, and DELETE `MiscellaneousWithALL` blocks all flip the same assertions without the same comment. Someone reading these tests in isolation will not know why these were previously `InvalidECSql`. Suggest copying the one-liner comment to each of the four UPDATE/DELETE blocks at the equivalent position.

iModelCore/ECDb/Scripts/BuildECSqlParser.sh
khanaffan: Good idea but not useful. The grammar change is made by copilot it already do that by default where it check existing conflict and then make grammar change and then see if it has added to conflict if yes it debug and fix it. 

I attempted to reduce conflicts to zero but that seem like a large refactor of grammar. We inherited this grammar and its originally had those conflict in boolean_val subtree. The issue really is its harmless according to copilot and cost of fixing it carry high risk.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
khanaffan: @copilot add the comment

iModelCore/ECDb/ECDb/ECSql/Parser/sqlbison.y
khanaffan: Good catch — the comment overstated it. Confirmed via the `IsAndIsNotOperatorNullSafeSemantics` test: `S1 IS (S2)` returns 2 rows, i.e. `(S2)` reduces as a parenthesized `value_exp` (null-safe comparison), **not** a `type_predicate`. The type predicate is only reached for a qualified class name (`schema.Class`), an `ONLY`/`ALL` prefix, or a comma-separated list (cf. `ECClassId IS (ecsql.TH5)` / `IS (ONLY ...)`). So when an identifier is both a class and a property name, the value-expression (property) reading wins. Reworded the grammar comment to state this explicitly in a0bc9f2a0e.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
khanaffan: `ParseTruthValue` was indeed dead — a one-line shim `{ return ParseValueExp(...); }` with zero callers repo-wide — so I removed it in a0bc9f2a0e. The legacy `IS (ClassName)` path now calls `ParseValueExp` directly (which is why `ECClassId IS (ts.Foo)` still resolves via `ec_cache_ClassHierarchy`); the commit message now calls out that swap as you suggested.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
khanaffan: Done — added the `// 'IS' / 'IS NOT' accept any value expression on either side (null-safe comparison)` comment to all four UPDATE/DELETE blocks (UPDATE `WhereBasics`/`MiscellaneousWithALL`, DELETE `WhereBasics`/`MiscellaneousWithALL`), placed before each flipped assertion. Resolving.

iModelCore/ECDb/CHANGES.md
```diff
@@ -5,7 +5,15 @@ This document including important changes to syntax or file format.
 | Module  | Version   |
 | ------- | --------- |
 | Profile | `4.0.0.5` |
-| ECSQL   | `2.0.3.2` |
+| ECSQL   | `2.0.4.0` |
+
+## ## `06/23/2026`: Added `IS` / `IS NOT` operator between operands
+* ECSql version change `2.0.3.2` -> `2.0.4.0`.
+* The `IS` and `IS NOT` operators can now be used between two operands, e.g. `prop1 IS [NOT] prop2`. Each operand may be any value expression — a property, the `NULL` literal, a constant, a parameter, a function call, an arithmetic expression, etc. These map to SQLite's null-safe comparison operators (`NULL IS NULL` is true, `1 IS NULL` is false).
+* Previously `IS` / `IS NOT` only supported the right-hand operands `NULL`, the boolean literals `TRUE`/`FALSE`/`UNKNOWN`, and the class type predicate `IS (ClassName)`. Those forms are unchanged: a right-hand operand that is exactly `NULL`/`TRUE`/`FALSE`/`UNKNOWN`, or a parenthesized `(ClassName)`, keeps its original meaning. (Write a parenthesized value expression without the outer parentheses, e.g. `x IS y + 1` rather than `x IS (y + 1)`, when it could be mistaken for a type predicate.)
```
aruniverse: **Disambiguation note should say 'qualified' — consistent with the `sqlbison.y` grammar comment**

This sentence implies any parenthesized class name keeps the type-predicate meaning:
> *"a right-hand operand that is exactly `NULL`/`TRUE`/`FALSE`/`UNKNOWN`, or a parenthesized `(ClassName)`, keeps its original meaning"*

Per the grammar comment you added to `type_list_item` in `sqlbison.y`:
> *"A type predicate is reached only for a **qualified** class name (`schema.Class`), an ONLY/ALL prefix, or a comma-separated list. A single **unqualified** name in parentheses … still reduces as a parenthesized `value_exp` (null-safe comparison), not a type predicate."*

So `IS (MyClass)` with an unqualified name has **never** been a type predicate and still isn't — a user reading this note could incorrectly assume it is. The same imprecision appears in itwinjs-core's `Operators.md` (line 78) and `NextVersion.md` (line 185); commenting there too.

Suggested minimal fix (all three locations) — change:
```
or a parenthesized `(ClassName)`, keeps its original meaning
```
to:
```
or a parenthesized *qualified* class name (e.g. `(schema.Class)`), an `ONLY`/`ALL`-prefixed form, or a comma-separated list, keeps its original meaning. A single unqualified name in parentheses (e.g. `(MyProp)`) is always parsed as a value expression, even if a class by that name exists.
```

aruniverse (commented):
One follow-up from the latest round: the `CHANGES.md` disambiguation note (and the parallel wording in itwinjs-core's `Operators.md` and `NextVersion.md`) still says "a parenthesized `(ClassName)`" without specifying that the name must be **qualified**. The grammar comment in `sqlbison.y` is already accurate — this just needs to match it.

iModelCore/ECDb/CHANGES.md
khanaffan: This is actually a good catch. if `IS (abc.def)` then `abc.def` can be a class or property `abc` could be schema or class and `def` could be class or property. So `IS (<ident1>.<ident2>)`. Beside the doc, I think this need a more precious error message as well. It will work correctly and fail but it will say "class not found". But on other hand `(alias.prop)` not incorrect syntax either. 

Let me think about it.

khanaffan: So I have change so we support `IS (class.prop)` as it would be deviation from value-expr where any value expression including prop can be encapsulated  in parenthesis.

khanaffan: Done in 593ff15eb9. Rather than only rewording the doc, I implemented `IS (class.prop)` support so this case stops being ambiguous in practice: a parenthesized qualified name on the RHS of `IS`/`IS NOT` now reinterprets as a property-reference / enum-literal value expression (null-safe comparison) when it does **not** resolve to a class — so `X IS (alias.prop)` and `X IS (ts.Enum.Enumerator)` work instead of failing with "class not found". A genuine `(ClassName)` predicate, an `ONLY`/`ALL` prefix, a comma-separated list, or the `schema:class` colon form keep their original type-predicate meaning.

`CHANGES.md` and the `sqlbison.y` grammar comment now spell out the *qualified class name* (type predicate) vs *property path* (value expression) distinction you flagged, and I added prepare / statement-semantics (incl. enum + nested struct) / SQL-generation tests.

The parallel wording fix in itwinjs-core's `Operators.md` and `NextVersion.md` will go on PR #9440. Leaving this thread open for your re-review.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
```diff
@@ -1719,8 +1719,14 @@ TEST_F(ECSqlSelectPrepareTests, WhereBasics)
...
     EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P WHERE i>=:myParam"));
-    EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("SELECT NULL FROM ecsql.P WHERE I IS 123"));
+    // 'IS' / 'IS NOT' accept any value expression on either side (null-safe comparison), not just NULL.
+    EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P WHERE I IS 123"));
     EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P WHERE B IS TRUE"));
+    // a parenthesized qualified property reference '(alias.prop)' is a value expression (null-safe
+    // comparison), not the '(ClassName)' type predicate, when the name does not resolve to a class.
+    EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P WHERE I IS (P.I)"));
+    EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P p WHERE p.I IS (p.I)"));
+    EXPECT_EQ(ECSqlStatus::Success, Prepare("SELECT NULL FROM ecsql.P p WHERE p.I IS NOT (p.I)"));
```
aruniverse, line 1729: **Missing error test: `IS (alias.NonExistentProperty)` — what happens when the name is neither a class nor a valid property?**

The three lines above confirm that `IS (P.I)` / `IS NOT (p.I)` prepare successfully. But there is no test for the case where the parenthesized name passes the class-lookup check (returns null), falls through `TryParseParenthesizedNameAsValueExp`, and then constructs a `PropertyNameExp` for a name that does *not* exist as a property.

For example: `I IS (P.NonExistentProp)` — `P.NonExistentProp` is not a class, so `TryParseParenthesizedNameAsValueExp` returns `SUCCESS` with a `PropertyNameExp`. But what does downstream semantic validation do? Without a test, the failure mode (and its error message) is unanchored.

Suggested additions:
```cpp
// parenthesized name that is neither a class nor a property → prepare error
EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("SELECT NULL FROM ecsql.P WHERE I IS (P.NonExistentProp)"));
// parenthesized enum name where the enumerator does not exist → prepare error (not a silent pass)
// EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("..."));
```

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
```diff
@@ -1274,4 +1274,75 @@ TEST_F(ECSqlToSqlGenerationTests, SelectOnlyRequiredPropertiesOnSelfJoin)
...
+    EXPECT_TRUE(GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS (ts.Foo)").Contains("ec_cache_ClassHierarchy"));
+
+    // a parenthesized qualified property reference '(alias.prop)' is a value expression (null-safe
+    // comparison), not the (ClassName) type predicate, because the name does not resolve to a class
+    assertWhere("SELECT 1 FROM ts.Foo WHERE S1 IS (Foo.S2)", "WHERE [Foo].[S1] IS [Foo].[S2]");
+    assertWhere("SELECT 1 FROM ts.Foo WHERE S1 IS NOT (Foo.S2)", "WHERE [Foo].[S1] IS NOT [Foo].[S2]");
+    // multi-column point operand still expands column-wise through the parenthesized property reference
+    assertWhere("SELECT 1 FROM ts.Foo WHERE P1 IS (Foo.P2)",
+                "WHERE ([Foo].[P1_X] IS [Foo].[P2_X] AND [Foo].[P1_Y] IS [Foo].[P2_Y] AND [Foo].[P1_Z] IS [Foo].[P2_Z])");
+    }
```
aruniverse, line 1368: **Missing regression tests: `IS (ONLY ts.Foo)` and `IS (ts.Foo, ts.Bar)` after grammar refactor**

The grammar refactored `type_list_item` from `opt_only table_node` into three explicit alternatives (including new `SQL_TOKEN_ONLY` and `SQL_TOKEN_ALL` rules). The semantic guard in `TryParseParenthesizedNameAsValueExp` bails out via `opt_only->count() != 0` to keep ONLY/ALL forms as type predicates. This is correct by inspection, but no SQL-generation test verifies the end-to-end output after the refactor.

Suggested additions at the end of this test:
```cpp
// ONLY-prefixed and comma-separated type-list forms must still produce type-predicate SQL
// after the type_list_item grammar refactor — not fall into TryParseParenthesizedNameAsValueExp
EXPECT_TRUE(GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS (ONLY ts.Foo)").Contains("ec_cache_ClassHierarchy"));
EXPECT_TRUE(GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS (ts.Foo, ts.Bar)").Contains("ec_cache_ClassHierarchy"));
```

aruniverse (commented):
Two new test-coverage gaps found after reviewing commit `593ff15eb9` (`TryParseParenthesizedNameAsValueExp`). Both are in the test files. Existing implementation looks correct — these are missing negative/regression tests only.

iModelCore/ECDb/ECDb/ECSql/Parser/sqlbison.y
aruniverse: ✅ **Resolved** — addressed in two commits:
- `a0bc9f2a0e`: grammar comment rewritten to accurately state that a single unqualified name reduces as `value_exp`, and that the type-predicate path requires a qualified class name, ONLY/ALL prefix, or comma-separated list.
- `593ff15eb9`: comment further updated to reflect that a qualified name in parentheses reduces grammatically as `type_predicate` but the semantic layer (`TryParseParenthesizedNameAsValueExp`) reinterprets it as a value expression when the name does not resolve to a class.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
aruniverse: ✅ **Resolved** — commit `a0bc9f2a0e` removed the `ParseTruthValue` inline shim from `ECSqlParser.h` (it was a one-liner wrapper `{ return ParseValueExp(exp, node); }` with zero callers repo-wide). The call site in `ECSqlParser.cpp` now calls `ParseValueExp` directly. The commit message explicitly documents the swap and confirms the regression test (`ECClassId IS (ts.Foo)` → `ec_cache_ClassHierarchy`) still passes.

iModelCore/ECDb/Scripts/BuildECSqlParser.sh
aruniverse: ✅ **Acknowledged / deferred** — the author explained that Copilot-assisted grammar changes already track conflict deltas during development, and that the inherited grammar has many pre-existing conflicts that cannot be eliminated without a large refactor. The build scripts were updated with explanatory comments covering why `-Wno-conflicts-sr/-rr` is used and why `%expect`/`%expect-rr` is not viable for this LALR grammar. A numeric count would add value but is not blocking — no further action required on this thread.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
aruniverse: ✅ **Resolved** — commit `9b17160cb7e9` (copilot-swe-agent) added the `// 'IS' / 'IS NOT' accept any value expression on either side (null-safe comparison), not just NULL.` comment to all four statement-type blocks: `ECSqlUpdatePrepareTests::WhereBasics`, `ECSqlUpdatePrepareTests::MiscellaneousWithALL`, `ECSqlDeletePrepareTests::WhereBasics`, and `ECSqlDeletePrepareTests::MiscellaneousWithALL`, matching the comment already present in `ECSqlSelectPrepareTests::WhereBasics`.

iModelCore/ECDb/CHANGES.md
aruniverse: ✅ **Resolved** — commit `593ff15eb9` went further than rewording: it implemented `ECSqlParser::TryParseParenthesizedNameAsValueExp` so that `IS (alias.prop)` actually works as a null-safe value expression when the name doesn't resolve to a class. The CHANGES.md entry now accurately describes all three cases: (1) qualified class name keeps type-predicate semantics, (2) qualified non-class name is treated as a value expression (null-safe comparison), and (3) when a name is both a class and a property path, the class reading wins.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
khanaffan: Added in e2ba404e56. Two `InvalidECSql` cases in `ECSqlSelectPrepareTests::WhereBasics`, right after the `IS (P.I)` lines:

```cpp
EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("SELECT NULL FROM ecsql.P WHERE I IS (P.NonExistentProp)"));
EXPECT_EQ(ECSqlStatus::InvalidECSql, Prepare("SELECT NULL FROM ecsql.P p WHERE p.I IS NOT (p.NonExistentProp)"));
```

`P.NonExistentProp` is not a class, so `TryParseParenthesizedNameAsValueExp` returns a `PropertyNameExp`, which then fails property resolution during finalize — i.e. the failure mode is a normal "property not found" prepare error, not a silent pass (same behavior as `WHERE Garbage = 'bla'`).

The `ecsql.P` schema has no enumeration, so I anchored the enum failure mode you mentioned in `ECSqlStatementTests::IsAndIsNotOperatorNullSafeSemantics` (which has a `Status` enum) instead:

```cpp
EXPECT_NE(ECSqlStatus::Success, badProp.Prepare(m_ecdb, "SELECT 1 FROM ts.Foo WHERE S1 IS (Foo.NonExistentProp)"));
EXPECT_NE(ECSqlStatus::Success, badEnum.Prepare(m_ecdb, "SELECT 1 FROM ts.Foo WHERE Status IS (ts.Status.DoesNotExist)"));
```

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
khanaffan: Added in e2ba404e56 — with one correction to the suggested assertions.

`IS (ONLY ts.Foo)` does **not** produce `ec_cache_ClassHierarchy`. `PrepareTypeListExp` has an exact-match optimization: a list containing only non-polymorphic (`ONLY`) classes is emitted as a direct `IN (<classId>)` filter and never touches `ec_cache_ClassHierarchy` (that table is only used when the list has a polymorphic entry, via `BaseClassId IN (...)`). So checking for `ec_cache_ClassHierarchy` on the `ONLY` form would fail.

What I added (a `Bar` class was added to the schema for the comma-list cases):

```cpp
// ONLY (non-polymorphic) -> exact IN(<id>) filter, no ec_cache_ClassHierarchy
Utf8String onlySql = GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS (ONLY ts.Foo)");
EXPECT_FALSE(onlySql.empty());                              // still prepares as a type predicate
EXPECT_FALSE(onlySql.Contains("ec_cache_ClassHierarchy"));
// ... same for (ONLY ts.Foo, ONLY ts.Bar)

// polymorphic comma-separated list -> ec_cache_ClassHierarchy
EXPECT_TRUE(GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS (ts.Foo, ts.Bar)").Contains("ec_cache_ClassHierarchy"));
EXPECT_TRUE(GetHelper().ECSqlToSql("SELECT 1 FROM ts.Foo WHERE ECClassId IS NOT (ts.Foo, ts.Bar)").Contains("ec_cache_ClassHierarchy"));
```

`EXPECT_FALSE(empty())` confirms the `ONLY`/comma-list forms still route through the type-predicate path (they'd yield an empty string if they had misparsed and failed), and the poly-list cases pin the `ec_cache_ClassHierarchy` output — so the `type_list_item` refactor and the `TryParseParenthesizedNameAsValueExp` `opt_only`/comma-list guards are both exercised end-to-end.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -3466,6 +3507,103 @@ BentleyStatus ECSqlParser::ParseTypePredicate(std::unique_ptr<ValueExp>& valueEx
...
+// a "class not found" error.
+// 'valueExp' is left null - so the caller keeps the type-predicate behavior - when the predicate uses
+// an ONLY/ALL prefix, a comma-separated type list, the 'schema:class' (colon) form, or a name that
+// does resolve to a class (e.g. 'ECClassId IS (ts.Foo)' or 'ECClassId IS (main.ts.Foo)').
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+BentleyStatus ECSqlParser::TryParseParenthesizedNameAsValueExp(std::unique_ptr<ValueExp>& valueExp, OSQLParseNode const* typePredicateNode) const
+    {
+    valueExp = nullptr;
+    if (!SQL_ISRULE(typePredicateNode, type_predicate))
```
soham-bentley, line 3524: This method has no code path which returns in `ERROR`, but I think here if the `typePredicateNode` is not a `type_predicate`, we should error out.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -3466,6 +3507,103 @@ BentleyStatus ECSqlParser::ParseTypePredicate(std::unique_ptr<ValueExp>& valueEx
...
+    if (type_list->count() != 1)
+        return SUCCESS; // a comma-separated type list is always a genuine type predicate
+
+    OSQLParseNode const* type_list_item = type_list->getChild(0);
+    OSQLParseNode const* opt_only = type_list_item->getChild(0/*opt_only*/);
+    if (opt_only->count() != 0)
+        return SUCCESS; // an ONLY/ALL prefix is always a genuine type predicate
+
+    OSQLParseNode const* tableNode = type_list_item->getChild(1/*table_node*/);
+    if (!SQL_ISRULE(tableNode, table_node))
```
soham-bentley, line 3537: If the `tableNode` is not of type `table_node`, shouldn't we error out?

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
```diff
@@ -3466,6 +3507,103 @@ BentleyStatus ECSqlParser::ParseTypePredicate(std::unique_ptr<ValueExp>& valueEx
...
+            return SUCCESS;
+
+        propertyPath.Push(schemaNameOrAlias);
+        propertyPath.Push(propertyName);
+        }
+    else if (SQL_ISRULE(nameNode, tablespace_qualified_class_name))
+        {
+        Utf8StringCR tableSpaceName = nameNode->getChild(0)->getTokenValue();
+        OSQLParseNode const* qualifiedNameNode = nameNode->getChild(2/*qualified_class_name*/);
+        if (!SQL_ISRULE(qualifiedNameNode, qualified_class_name))
```
soham-bentley, line 3565: Here also same, if `qualifiedNameNode` is not a `qualified_class_name` shouldn't we error out or atleast continue with the execution as is but not branch out on `SUCCESS`? This method just branches out `SUCCESS` when we are sure that the node is a className somehow, right?

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
aruniverse: Partially addressed in `e2ba404e56`: the missing-property cases were added in `ECSqlPrepareTests.cpp`, with matching missing-property and missing-enumerator cases in `ECSqlStatementTests.cpp`. One part of the original concern remains: these assertions check only `InvalidECSql` / non-success, so the pre-fix "class not found" path would also pass. Please capture the issue (for example with `TestIssueListener`) and assert property/enumerator resolution—or at least assert that this is not the old class-resolution failure—before resolving the thread.

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
aruniverse: The original requested regressions are addressed in `e2ba404e56`: `IS (ONLY ts.Foo)` and an all-`ONLY` list are verified as exact matches without `ec_cache_ClassHierarchy`, while polymorphic comma lists for `IS` and `IS NOT` are verified to use the hierarchy table. The correction to the suggested `ONLY` assertion is valid. I posted one separate follow-up because the new test comment claims `ALL` coverage, but the distinct `SQL_TOKEN_ALL` grammar branch is still untested.

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
```diff
@@ -1274,4 +1274,96 @@ TEST_F(ECSqlToSqlGenerationTests, SelectOnlyRequiredPropertiesOnSelfJoin)
...
+
+    // a parenthesized qualified property reference '(alias.prop)' is a value expression (null-safe
+    // comparison), not the (ClassName) type predicate, because the name does not resolve to a class
+    assertWhere("SELECT 1 FROM ts.Foo WHERE S1 IS (Foo.S2)", "WHERE [Foo].[S1] IS [Foo].[S2]");
+    assertWhere("SELECT 1 FROM ts.Foo WHERE S1 IS NOT (Foo.S2)", "WHERE [Foo].[S1] IS NOT [Foo].[S2]");
+    // multi-column point operand still expands column-wise through the parenthesized property reference
+    assertWhere("SELECT 1 FROM ts.Foo WHERE P1 IS (Foo.P2)",
+                "WHERE ([Foo].[P1_X] IS [Foo].[P2_X] AND [Foo].[P1_Y] IS [Foo].[P2_Y] AND [Foo].[P1_Z] IS [Foo].[P2_Z])");
+
+    // regression: the type_list_item grammar refactor (opt_only is now inlined into three explicit
```
aruniverse, line 1351: **Missing `ALL` regression despite this comment claiming ONLY / ALL coverage.** `sqlbison.y` has a distinct `SQL_TOKEN_ALL table_node` alternative, but this block exercises only `ONLY` and unprefixed comma lists; there is no `IS (ALL ...)` case repo-wide. Please add at least `IS (ALL ts.Foo)` (and preferably `IS NOT (ALL ts.Foo)`) and assert it remains on the polymorphic type-predicate path, e.g. via `ec_cache_ClassHierarchy`.

iModelCore/ECDb/CHANGES.md
```diff
@@ -5,7 +5,15 @@ This document including important changes to syntax or file format.
...
+* ECSql version change `2.0.3.2` -> `2.0.4.0`.
+* The `IS` and `IS NOT` operators can now be used between two operands, e.g. `prop1 IS [NOT] prop2`. Each operand may be any value expression — a property, the `NULL` literal, a constant, a parameter, a function call, an arithmetic expression, etc. These map to SQLite's null-safe comparison operators (`NULL IS NULL` is true, `1 IS NULL` is false).
+* Previously `IS` / `IS NOT` only supported the right-hand operands `NULL`, the boolean literals `TRUE`/`FALSE`/`UNKNOWN`, and the class type predicate `IS (ClassName)`. Those forms are unchanged: a right-hand operand that is exactly `NULL`/`TRUE`/`FALSE`/`UNKNOWN`, or a parenthesized name that resolves to a class — a qualified `(schema.Class)`, an `ONLY`/`ALL`-prefixed form, or a comma-separated list — keeps its original type-predicate meaning. A parenthesized name that does **not** resolve to a class, e.g. a property reference `(alias.prop)`, is parsed as an ordinary value expression (null-safe comparison), exactly as it would be without the parentheses. When a parenthesized qualified name is both a class and a valid property path, the class (type-predicate) reading wins.
```
aruniverse, line 13: **Please narrow the fallback wording to dot-qualified names.** `TryParseParenthesizedNameAsValueExp` explicitly excludes the `schema:class` and `tablespace.schema:class` colon forms; an unresolved colon-form name remains a type predicate and fails class resolution instead of becoming a value expression. As written, "a parenthesized name that does not resolve to a class" overstates the behavior. Please say *dot-qualified* (for example `alias.prop`) and mirror the constraint in the linked core docs.

iModelCore/ECDb/Tests/NonPublished/ECSqlPrepareTests.cpp
khanaffan: Fixed.

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
khanaffan: Fixed.

iModelCore/ECDb/Tests/NonPublished/ECSqlToSqlGenerationTests.cpp
khanaffan: Fixed.

iModelCore/ECDb/CHANGES.md
khanaffan: Fixed.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
khanaffan: The caller only calls this method when the node is already a `type_predicate`, so this branch cannot be hit. Returning `ERROR` here would be dead code, and `SUCCESS` with a null `valueExp` is how this method says "nothing to reinterpret, keep the type predicate" - it is not a success/failure flag. I added a `BeAssert` so a bad node is caught in debug builds, and documented the contract in the method comment.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
khanaffan: The grammar guarantees this: `type_list_item` is always `opt_only table_node`, so child 1 is always a `table_node`. The check is only defensive, so I added a `BeAssert` instead of an `ERROR` return (same style as `ParseTypePredicate`). Returning `ERROR` would change behavior for a case that cannot happen.

iModelCore/ECDb/ECDb/ECSql/ECSqlParser.cpp
khanaffan: Same as above - the grammar rule is `tablespace_qualified_class_name: SQL_TOKEN_NAME '.' qualified_class_name`, so child 2 is always a `qualified_class_name`. Added a `BeAssert` here too.

To clarify the return value: this method returns `SUCCESS` in two different situations - either it produced a value expression, or it decided the name is not a value expression and left `valueExp` null so the caller keeps the type-predicate path. `ERROR` is only for a real parse failure. I added this to the method comment so it is clear.

aruniverse (approved):

RohitPtnkr1996 (approved):

PR #1479, opened 2026-06-26
SchemaView performance improvements

rschili:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9431

New schema view fragment pragma that will be used by iTwinjs-core to load just subsets of an iModels' schema.
Plus a few fixes, please read the linked itwinjs-core PR description and issue for full details.


iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
```diff
@@ -1064,8 +1064,11 @@ std::string QueryHelper::FormatQuery(const char* query) {
 // @bsimethod
 //---------------------------------------------------------------------------------------
 void QueryHelper::BindLimits(ECSqlStatement& stmt, QueryLimit const& limit) {
-    const auto idxCount = stmt.GetParameterIndex(LIMIT_VAR_COUNT);
-    const auto idxOffset = stmt.GetParameterIndex(LIMIT_VAR_OFFSET);
+    // A PRAGMA (or any statement FormatQuery left unwrapped) has no LIMIT/OFFSET parameters. Use the
+    // non-logging TryGetParameterIndex so probing for their absence does not emit "No
+    // parameter index found" errors - GetParameterIndex would log one per missing parameter.
+    const auto idxCount = stmt.TryGetParameterIndex(LIMIT_VAR_COUNT);
```
rschili, line 1168: @khanaffan are you okay with this change?
The previous code was logging a warning because pragma going through concurrent query does not have those two parameters.
This change makes it so we don't log those warnings anymore.
Unrelated change, sneaking it in because it's small and insignificant.

iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
```diff
@@ -982,52 +1073,107 @@ DbResult PragmaSchemaView::Read(PragmaManager::RowSet& rowSet, ECDbCR ecdb, Prag
...
+	// Parse the comma-separated decimal id list into a de-duplicated set; reject any malformed id
+	bvector<Utf8String> tokens;
+	BeStringUtilities::Split(idList.c_str(), ",", tokens);
+	std::unordered_set<int64_t> schemaIds;
+	for (Utf8StringR token : tokens) {
+		token.Trim();
+		if (!IsAllDigits(token))
+			return reportError(Utf8PrintfString("invalid schema id '%s'; ids must be positive decimal integers.", token.c_str()));
+		int64_t id = (int64_t)strtoll(token.c_str(), nullptr, 10);
+		schemaIds.insert(id).second; // de-duplicate
```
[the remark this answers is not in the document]
rschili: This was a conscious decision, I do not want to fail this on duplicate IDs, so dedup is correct, I missed an error message and test spot that need adjusting. Both are fixed now.

RohitPtnkr1996 (approved):

iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
```diff
@@ -982,52 +1079,117 @@ DbResult PragmaSchemaView::Read(PragmaManager::RowSet& rowSet, ECDbCR ecdb, Prag
...
+	if (!val.IsString())
+		return reportError("expects a single string argument: a comma-separated list of schema names, optionally prefixed with a 'v<N>;' format-version token.");
+
+	Utf8String arg(val.GetString().c_str());
+
+	// Optional leading 'v<N>;' format-version token. It rides inside the single string argument
+	// because the pragma grammar allows only one. Schema names are ECNames, so a ';' always means a
+	// version prefix.
+	uint8_t requestedVersion = CURRENT_FORMAT_VERSION;
+	Utf8String nameList = arg;
```
soham-bentley: We are just doing parsing of `arg` here. Can't we use `string_view` here? instead of `string`?

iModelCore/ECDb/ECDb/SchemaViewWriter.cpp
```diff
@@ -548,9 +562,9 @@ DbResult SchemaViewWriter::WriteEnumTable(DbCR db)
 
     while (stmt.Step() == BE_SQLITE_ROW)
         {
-        PutU32(SafeU32Id(stmt.GetValueInt64(7))); // schemaEcId
+        PutU32(SafeU32Id(stmt.GetValueInt64(7))); // schemaECId
         PutSRef(Safe(stmt.GetValueText(1)));       // name
-        PutU8((uint8_t)stmt.GetValueInt(4));        // primitiveType
+        PutU16(SafeU16(stmt.GetValueInt(4), "enumeration primitiveType")); // full ECN::PrimitiveType, e.g. 0x501 Integer / 0x901 String - does not fit in a byte
```
aruniverse, line 567: **Blocking compatibility issue:** this widens `EnumTable.primitiveType` from one byte to two while the blob header still declares format v1 (`FORMAT_VERSION = 1`). Existing v1 readers consume U8, while the linked new TS reader consumes U16. Therefore old frontend/new native and new frontend/released native both become byte-misaligned at the first enumeration. This defeats the frontend version pinning intended to support native/frontend version skew. Please introduce format v2, preserve the legacy U8 encoding whenever v1 is requested, and add cross-version fixtures containing an enumeration. The linked iTwin.js PR should request/read v2 while retaining a v1 reader.

aruniverse: @rschili we can ignore this as you have already accepted this and we have 0 usage of the api today correct?

rschili: yes exaxtly, it was either this or publishing a v2 format that is basically a bugfix of the v1 format. Seems overkill when nobody is using it yet.

iModelCore/ECDb/ECDb/ECSql/ECSqlPragmas.cpp
rschili: Good call, done. The version-prefix parsing now works on `std::string_view` slices over the single
materialized argument string; `nameList` stays a suffix of it, so its `data()` remains a valid
null-terminated string for `BeStringUtilities::Split`. The per-token `Utf8String`s stay - `Trim`,
`ECNameValidation::IsValidName` and the bind need them anyway.

iModelCore/ECDb/ECDb/ConcurrentQueryManagerImpl.cpp
rschili: Affan is out; RohitPtnkr1996 approved the PR including this change. It only suppresses a spurious warning, so I'll take it as-is - happy to revisit when Affan is back.

aruniverse (approved):

PR #1487, opened 2026-07-02
Switch libpng build to vcpkg

### LINKED ISSUE #1485: Use vcpkg to build libpng
As one of the third party libraries that has had a number of CVEs in the past year, libpng would benefit greatly from being built using vcpkg instead of bmake. Update its build to do this.

tcobbs-bentley:
## TL;DR

* Use vcpkg to build libpng instead of bmake.

<hr>

This pull request migrates the `png` (libpng) library in `iModelCore/libsrc/` from a vendored, file-by-file build to a vcpkg-based build, updating the build system, documentation, and related configuration files. The migration includes updating to libpng version 1.6.58, improving triplet support for multiple platforms, and clarifying best practices for future library migrations.

**Key changes:**

**Build System Migration and Improvements**
* Replaces the old file-by-file build of `png` with a vcpkg-based build in `png.mke`, delivering headers and static libraries directly from vcpkg outputs and simplifying the Makefile by removing legacy build logic.
* Adds custom vcpkg triplet files for `arm64-android`, `arm64-ios`, and `arm64-osx`, ensuring correct static linkage and symbol visibility to avoid runtime conflicts with Electron/Chromium. [[1]](diffhunk://#diff-82c061dec59d30387b487ceed793626b6b80930c49a101ed4435f8119b7913dfR1-R20) [[2]](diffhunk://#diff-ab3362a3473ff83657546d6f83d6abfe7893c47bf281b13be0ad9d5e5e7ea1a1R1-R16) [[3]](diffhunk://#diff-c883da8277407bbd49c11c7abe7e03b6696e63ebd9fa94e52edaf653b9f6d566R1-R14)
* Updates the vcpkg install chain in `vcpkg.PartFile.xml` and documentation to include `vcpkg_install_png` in the sequence, ensuring sequential and correct dependency installation. [[1]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1R21) [[2]](diffhunk://#diff-526001470b099fdcc0d8696d2c54cd37672f6a1bdf87e562ec92d594a3b3097cL168-R171) [[3]](diffhunk://#diff-13f392ce73a7b3c843fa657dc3b1b232f603762f9f9c13c8be6a4a346622b66fR8-R13)

**Documentation and Process Updates**
* Updates documentation in `VCPKG.md` and `.github/skills/vcpkg/SKILL.md` to clarify how to insert new libraries into the vcpkg chain, best practices for ordering, and the process for migrating previously vendored libraries (including when to delete old source). [[1]](diffhunk://#diff-526001470b099fdcc0d8696d2c54cd37672f6a1bdf87e562ec92d594a3b3097cL20-R20) [[2]](diffhunk://#diff-526001470b099fdcc0d8696d2c54cd37672f6a1bdf87e562ec92d594a3b3097cR32-R33) [[3]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1L74-R92) [[4]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1R148-R157)
* Updates the library table in `README.md` to reflect the new libpng version (1.6.58) and its vcpkg-managed status.

**Configuration and Metadata**
* Updates `PngNugetLicense.json` to match the new libpng version, SPDX ID, and copyright.
* Removes obsolete `.gitignore` entries related to the old vendored `png` source layout.
* Cleans up `png.PartFile.xml` to remove legacy VS2012 build logic and improve bindings for the vcpkg-based build. [[1]](diffhunk://#diff-13f392ce73a7b3c843fa657dc3b1b232f603762f9f9c13c8be6a4a346622b66fL23-L31) [[2]](diffhunk://#diff-13f392ce73a7b3c843fa657dc3b1b232f603762f9f9c13c8be6a4a346622b66fR8-R13)
* Updates the zlib override in `compress/vcpkg.json` to a new revision.

These changes modernize the handling of the `png` library, streamline maintenance, and set a clear pattern for future third-party library migrations to vcpkg.


iModelCore/libsrc/png/triplets/x64-linux.cmake
```diff
@@ -0,0 +1,5 @@
+set(VCPKG_TARGET_ARCHITECTURE x64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE static)
+
+set(VCPKG_CMAKE_SYSTEM_NAME Linux)
```
aruniverse, line 5: **Symbol visibility regression on Unix targets.** The old bmake build compiled BePng with hidden visibility (repo default — libs that want exports opt in via `GCC_DEFAULT_VISIBILITY=default`, cf. `BeJsonCpp.mke`, `freetype2.mke`; `BeCurl.mke` sets hidden specifically to avoid Node symbol-interposition crashes). The vcpkg-built `libpng16.a` has default-visibility `png_*` symbols, which end up dynamic-exported from `imodeljs.node` (linked in via `IModelJsNative_input_libs.mki:18`).

That's the exact hazard class the OpenSSL migration guarded against in its triplets (`openssl/triplets/x64-linux.cmake` — "prevents clashes with the OpenSSL that Node bundles (the Node 18 segfault)"), and Electron/Chromium bundle their own libpng, so exported preemptible `png_*` can silently bind to the wrong implementation.

Suggest mirroring the openssl triplets in all five non-Windows png triplets:
```cmake
set(VCPKG_C_FLAGS "${VCPKG_C_FLAGS} -fvisibility=hidden -fvisibility-inlines-hidden")
set(VCPKG_CXX_FLAGS "${VCPKG_CXX_FLAGS} -fvisibility=hidden -fvisibility-inlines-hidden")
```
(libpng is pure C so the C flag is the essential one.) Alternatively, demonstrate with `nm -D imodeljs.node | grep ' T png_'` old-vs-new that nothing new is exported, and record that in the PR.

iModelCore/libsrc/vcpkg.PartFile.xml
```diff
@@ -17,12 +17,18 @@
...
         <!-- LibType="Static": the chain always runs static-only.  Dynamic consumers depend on
              their chain part with LibType="Static" and read the packages from the static
              install root; they do not run vcpkg themselves. -->
         <SubPart PartName="vcpkg_install_compress" LibType="Static"/>
     </Part>
+    <Part Name="vcpkg_install_openssl" BentleyBuildMakeFile="vcpkg_install_openssl.mke">
+        <!-- LibType="Static": the chain always runs static-only.  Dynamic consumers depend on
+             their chain part with LibType="Static" and read the packages from the static
+             install root; they do not run vcpkg themselves. -->
+        <SubPart PartName="vcpkg_install_png" LibType="Static"/>
```
aruniverse, line 30: This re-parents openssl onto png, but the sibling comment in `vcpkg_install_openssl.mke:6` still says "Runs after vcpkg_install_compress as part of the sequential chain" — now stale. One-line fix there.

Relatedly (minor): inserting png *mid-chain* contradicts the repo's own documented convention — `VCPKG.md:20` and the vcpkg skill both say to add new links **at the end of the chain**. Functionally harmless (the chain stays linear and every consumer depends on its own named part), but the next migration author gets conflicting signals. Either move `vcpkg_install_png` to the end (after crashpad) or relax the wording in VCPKG.md/SKILL.md.

iModelCore/libsrc/png/PngNugetLicense.json
```diff
@@ -16,7 +16,7 @@
         "licenseUrl": "http://www.libpng.org/pub/png/src/libpng-LICENSE.txt",
         "name": "libpng",
         "SPDX-ID": "Libpng",
-        "version": "1.6.37"
+        "version": "1.6.58"
```
aruniverse, line 19: Since this stanza is being touched to claim 1.6.58, the surrounding metadata should be corrected too: libpng ≥ 1.6.36 is licensed under the *PNG Reference Library License v2* — SPDX **`libpng-2.0`**, not `Libpng` (confirmed by the port manifest at the pinned baseline: `ports/libpng/vcpkg.json` → `"license": "libpng-2.0"`). The copyright holder for current versions is the PNG Reference Library Authors (maintainer Cosmin Truta), not Glenn Randers-Pehrson. As-is the file asserts stale license metadata for the exact version it now ships.

iModelCore/libsrc/png/vcpkg.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "dependencies": [
+    {
+      "name": "libpng",
+      "version>=": "1.6.58"
+    }
+  ],
+  "overrides": [
+    { "name": "libpng", "version": "1.6.58" },
+    { "name": "zlib", "version": "1.3.2" }
```
aruniverse: Nit: an override without an explicit port-version resolves to `1.3.2#0`, while the baseline default is `1.3.2#1` — so this silently skips the baseline's packaging fix. It's the identical string compress uses, so there's no cross-tree skew (which is what actually matters for ABI), but note compress pins minizip *with* port-version (`1.3.1#1`), so the repo is inconsistent about the convention. If you touch this, change compress's zlib override in the same PR to keep the two trees in lockstep.

iModelCore/libsrc/png/png.PartFile.xml
```diff
@@ -5,11 +5,12 @@
...
 
     <Part Name="pnglib" BentleyBuildMakeFile="png.mke">
+        <SubPart PartName="vcpkg_install_png" PartFile="iModelCore/libsrc/vcpkg" LibType="Static"/>
         <SubPart LibType="Static" PartName="Zlib" PartFile="iModelCore\libsrc\compress\Zlib"/>
         <Bindings>
             <VendorAPI Domain="png"/>
-            <Libs>Delivery/$(stlibprefix)BePng$(stlibext)</Libs>
-            <Libs>Delivery/$(stlibprefix)BePng$(stlibdebugext)</Libs>
+            <Libs ProductDirectoryName="BeStaticLibDir" Required="false">Delivery/$(stlibprefix)BePng$(stlibext)</Libs>
+            <Libs ProductDirectoryName="BeStaticLibDir" Required="false" IfNotPresent="Continue">Delivery/$(stlibprefix)BePng$(stlibdebugext)</Libs>
```
aruniverse, line 13: Nit: this debug-ext binding can never resolve — the new `png.mke` only ever delivers the `$(stlibext)` name (in DEBUG builds it maps the debug vcpkg lib *content* under the release-style name, same as `Zlib.mke`). Harmless due to `IfNotPresent="Continue"` and identical to `Zlib.PartFile.xml`, so fine to keep for pattern parity — but it's dead config; consider dropping it here and in Zlib's PartFile in a follow-up.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
```diff
@@ -0,0 +1,414 @@
...
+
+> **Done.** Inserted `vcpkg_install_png` into the chain in
+> [`../vcpkg.PartFile.xml`](../vcpkg.PartFile.xml) **after compress and before openssl** with
+> `<SubPart PartName="vcpkg_install_compress" LibType="Static"/>` (and openssl now chains off
+> `vcpkg_install_png`), keeping it static-only. Updated the "How It Works" chain description
+> in [`../VCPKG.md`](../VCPKG.md) to `compress → png → openssl → crashpad` and added `png.mke`
+> to the consumer-mke list.
+
+In [`../vcpkg.PartFile.xml`](../vcpkg.PartFile.xml), append a new link at the **end** of the
+chain (after `vcpkg_install_crashpad`) so no two vcpkg processes ever run concurrently:
```
aruniverse: The plan body here still instructs appending **after `vcpkg_install_crashpad`** (sample XML below chains off crashpad), contradicting the Done banner above ("after compress and before openssl") and the actual `vcpkg.PartFile.xml`. Same issue in Step 5 (~line 221): "there is **no** `CREATE_STATIC_LIBRARIES` conditional" — but the final `png.mke:24-28` has one (correctly, per the build-verified blockquote).

The openssl migration doc carries an explicit top-of-file "retained for historical context" banner; this one doesn't, so the contradictions read as the current state. Suggest either the same banner or trimming the plan bodies to match the final state.

While editing: worth one sentence in Step 8 noting that arm64 macOS/Android now execute NEON filter/palette paths (and all x64 targets SSE2 de-filtering, via the port's `PNG_INTEL_SSE` default) that the old build disabled everywhere — output-identical and the industry default, but useful to have discoverable if an image regression is ever bisected to this PR. Also nit: the "Current state" table links `vendor/…` paths this PR deletes.

aruniverse (commented):
Adversarial review of the meaningful diff (`git diff main...HEAD -- ':!iModelCore/libsrc/png/vendor'`), checked against the compress/openssl migrations as the reference pattern, the pinned registry baseline `81de6771`, and consumers in this repo + `imodel-native-internal`.

**Overall: no blockers.** Chain wiring, version pinning (libpng 1.6.58 + zlib 1.3.2, matching compress at the identical baseline), install-root static/dynamic redirection, debug-lib naming (`libpng16d` verified in the port on all six platforms), license delivery (`share/libpng/copyright` is the genuine upstream LICENSE), and the `png_VS2012` removal (no dangling refs anywhere) all check out. The old iOS NEON workaround is properly obsolete: the port forces `PNG_HARDWARE_OPTIMIZATIONS=OFF` on iOS (identical to old behavior, so the old link error can't recur), and on arm64 macOS/Android NEON is now coherently compiled *and* linked.

**Main actionable item** (inline): the vcpkg-built libpng has default symbol visibility, whereas the old bmake build was hidden — `png_*` now gets exported from `imodeljs.node` on Unix platforms. Same hazard class the OpenSSL migration explicitly guarded against; one-block fix per triplet.

Smaller items are inline. Two things verified clean but worth a one-time manual check before de-WIP since they're not provable from this repo:
- **PngNuget layout**: the binding switch to `ProductDirectoryName="BeStaticLibDir" Required="false"` matches Zlib convention, but worth a one-time diff of the produced nupkg layout old-vs-new in case any consumer addresses the lib by internal path.
- **Dropped internal headers** (`pnginfo.h`/`pngstruct.h`): zero consumers in this repo and `imodel-native-internal` (only libpng user is `ImageSource.cpp`, opaque-pointer API only); external PngNuget consumers are the residual the migration doc already defers.

Noted but explicitly *not* worth changing here: vcpkg's stock `/RTC1` in Windows debug flags (pre-existing, shared with compress/openssl/crashpad; worth tracking once in VCPKG.md, not per-library), `arm64-osx` triplet not pinning `VCPKG_OSX_DEPLOYMENT_TARGET` (same in all three migrations — fix centrally if at all), Android API 28 in the triplets (matches precedent, safe under minSdk 31), and the NEON/SSE2 runtime paths now being enabled (bit-identical output, industry default — suggested doc note inline).

iModelCore/libsrc/png/triplets/x64-linux.cmake
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg.PartFile.xml
tcobbs-bentley: Fixed.

iModelCore/libsrc/png/PngNugetLicense.json
tcobbs-bentley: Fixed.

iModelCore/libsrc/png/vcpkg.json
tcobbs-bentley: Fixed.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
tcobbs-bentley: Fixed.

iModelCore/libsrc/png/triplets/x64-linux.cmake
aruniverse: Addressed in a95b8b4b82 ("Build libpng with hidden visibility to avoid exporting png_* symbols", 2026-07-09 13:02 -0600). `-fvisibility=hidden -fvisibility-inlines-hidden` added to all 5 non-Windows triplets (arm64-android, arm64-ios, arm64-osx, x64-android, x64-linux), with a comment explaining the rationale (matches the repo's hidden-by-default convention, avoids clashing with Electron/Chromium's bundled libpng). Windows correctly left untouched (static .lib has no exports). Verified: all 5 files now contain the flag pair.

iModelCore/libsrc/vcpkg.PartFile.xml
aruniverse: Both parts addressed in 2d10056645 ("Fix to reflect png not being added to the end of the chain", 2026-07-09 13:08 -0600): (1) the stale sibling comment in `vcpkg_install_openssl.mke` now reads "Runs after vcpkg_install_png"; (2) `VCPKG.md` and `.github/skills/vcpkg/SKILL.md` were updated to document that chain position only needs to stay linear (mid-chain insertion is now an explicitly supported pattern, with guidance to update the sibling .mke comment on whichever link's predecessor changed) rather than mandating end-of-chain. Verified against current file contents.

iModelCore/libsrc/png/PngNugetLicense.json
aruniverse: Addressed in 52f5e17d90 ("Fix png license", 2026-07-09 13:10 -0600): `SPDX-ID` changed `Libpng` → `libpng-2.0`, and `copyright` changed "Glenn Randers-Pehrson" → "PNG Reference Library Authors", matching the port manifest at the pinned baseline. Verified against current file contents.

iModelCore/libsrc/png/vcpkg.json
aruniverse: Addressed in ffcd7425d3 ("Fix zlib version", 2026-07-09 13:24 -0600): both `compress/vcpkg.json` and `png/vcpkg.json` now pin `zlib` to `1.3.2#1` (explicit port-version, matching the baseline default), keeping the two trees in lockstep as suggested. Verified against current file contents.

iModelCore/libsrc/png/png.PartFile.xml
aruniverse: Not changed — this was flagged as an optional nit ("fine to keep for pattern parity"), and it's still present as of the current head. No action needed unless you want to clean it up alongside Zlib's PartFile in a follow-up.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
aruniverse: Addressed in 6d95588aba ("Fix VCPKG_MIGRATION.md", 2026-07-09 13:32 -0600): added a top-of-file "Note (post-migration)" banner stating the ✅ Done banners are authoritative and the prose below each is historical-only, plus inline "Historical note" callouts at the Step 4 chain-position text and the Step 5 CREATE_STATIC_LIBRARIES text explicitly reconciling them with the shipped state. Also added a NEON/SSE2 behavior-delta note per the earlier suggestion. Verified against current file contents.

aruniverse (commented):
Round 2 re-review after the 5 fix commits (`a95b8b4b82`, `2d10056645`, `52f5e17d90`, `ffcd7425d3`, `6d95588aba`) pushed in response to the prior review. Replied inline on each of the 6 original threads with when/where addressed. Ran 3 fresh-context reviewers over the current head to verify the fixes and scan for anything new.

**All 5 fixes verified fully correct, no regressions:**
- **Visibility** (`a95b8b4b82`): all 5 non-Windows triplets (arm64-android, arm64-ios, arm64-osx, x64-android, x64-linux) got matching `-fvisibility=hidden -fvisibility-inlines-hidden` on both C and CXX flags, byte-consistent with the openssl precedent; Windows correctly excluded (no MSVC equivalent needed for a static lib). Confirmed `-fvisibility=hidden` on a static archive only affects the *dynamic* export table of the final link, not intra-process symbol resolution between object files — the identical mechanism already ships for OpenSSL in this repo and works. Diff-integrity check: the commit is a pure append per file, nothing reordered.
- **Chain/doc fix** (`2d10056645`): `vcpkg_install_openssl.mke`, `VCPKG.md`, and `SKILL.md` are now mutually consistent (`compress → png → openssl → crashpad` everywhere); repo-wide grep found no contradictory stragglers in these three files.
- **License** (`52f5e17d90`): valid JSON; `SPDX-ID: libpng-2.0` / `copyright: PNG Reference Library Authors` confirmed correct against the vcpkg port manifest at the pinned baseline.
- **zlib version** (`ffcd7425d3`): both `compress/vcpkg.json` and `png/vcpkg.json` now pin `1.3.2#1` identically.
- **Migration doc** (`6d95588aba`): the Step 4/Step 5 contradictions this fixed are resolved and match the checked-in files.

**Two small new items found** (inline below), both residual stale prose in `VCPKG_MIGRATION.md` that the doc-fix commit didn't fully sweep — same category as what it already fixed elsewhere in the file, just missed in Step 7 and Step 9. Neither affects build correctness (both sit under "historical plan" prose, and the file's own "banners are authoritative" disclaimer covers the substantive contradiction; the dead link is the one thing the disclaimer doesn't cover, since it's not a contradiction, just a broken link).

**No blockers.** Also re-confirmed via fresh full-diff scan: `png.mke`, `png.PartFile.xml`, `vcpkg.json`/`vcpkg-configuration.json`, all triplets, `README.md`, and the `.gitignore` removal have no other new build-correctness, licensing, or supply-chain issues beyond what's already covered.

Optional/opinion items (not requiring action): the "chain stays linear" guidance in `VCPKG.md`/`SKILL.md` doesn't explicitly warn against giving a new link two `SubPart`s (which would silently defeat the mutual-exclusion the chain exists for) — worth one clarifying sentence sometime; and `VCPKG_MIGRATION.md`'s 440 lines are more build-diary than reference doc at this point, now that `VCPKG.md`/`SKILL.md` hold the durable material — could be trimmed post-merge, not a blocker.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
```diff
@@ -0,0 +1,440 @@
...
+## Step 7 — ✅ Update `README.md` and docs
+
+> **Done.** Bumped the `png` row in [`../README.md`](../README.md) to `libpng | 1.6.58 | Yes`.
+> The [`../VCPKG.md`](../VCPKG.md) "How It Works" chain description and consumer-mke list were
+> already updated in Step 4 (`compress → png → openssl → crashpad`).
+
+- [`../README.md`](../README.md) library table: bump the `png` row version `1.6.56` → `1.6.58`
+  and change the vcpkg column from `No` → `Yes`.
+- [`../VCPKG.md`](../VCPKG.md): add libpng to the chain description in "How It Works"
+  (`compress → openssl → crashpad → png`) and the consumer-mke list.
```
aruniverse, line 378: Same category of issue `6d95588aba` fixed in Step 4/5, missed here: this Step 7 historical-plan line still gives the stale order `compress → openssl → crashpad → png`, directly contradicting the correct Step 7 "Done" banner 4 lines above (which correctly says `compress → png → openssl → crashpad`) and the actual `VCPKG.md`/PartFile wiring. Cosmetic only — no build/runtime impact, and generically covered by the top-of-file "banners are authoritative" disclaimer — but worth folding into the same cleanup pass if you touch this file again.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
```diff
@@ -0,0 +1,440 @@
...
+> - The `png_VS2012` part was already removed in Step 6.
+> - Bumped the stale `libpng` version in [`PngNugetLicense.json`](PngNugetLicense.json)
+>   (`1.6.37` → `1.6.58`).
+> - Remaining png dir is now just the vcpkg wiring: `vcpkg.json`, `vcpkg-configuration.json`,
+>   `triplets/`, `png.mke`, `png.PartFile.xml`, `PngNugetLicense.json`, and this doc.
+
+- Delete the checked-in `vendor/` libpng source tree.
+- Remove the `png_VS2012` part (if retired in Step 6) and `PngNugetLicense.json` references that
+  point at `vendor/`.
+- Update [`.gitignore`](.gitignore) if it referenced `vendor/` build artifacts.
```
aruniverse, line 439: Dead link: this points at `iModelCore/libsrc/png/.gitignore`, which this same PR deletes. Unlike the Step 4/5/7 prose contradictions, this isn't something the "banners are authoritative" disclaimer covers (it's a broken hyperlink, not a claim that disagrees with the final state) — a renderer will 404 on it regardless of framing. Trivial fix: drop the markdown link syntax and just reference `.gitignore` as inline code.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
tcobbs-bentley: This matches the intro, which states that the original plan is preserved for historical reasons.

iModelCore/libsrc/png/VCPKG_MIGRATION.md
tcobbs-bentley: This matches the intro, which states that the original plan is preserved for historical reasons.

aruniverse (approved):

iModelCore/libsrc/png/VCPKG_MIGRATION.md
```diff
@@ -0,0 +1,440 @@
+# Migrating the libpng build to vcpkg
```
tcobbs-bentley, line 1: Need to decide what to do about libsrc-main and libsrc-vendor before merging this.

@DanRod1999

DanRod1999 (approved):

iModelCore/libsrc/png/VCPKG_MIGRATION.md
tcobbs-bentley: OK, libsrc-Main and libsrc-Vendor should be ready to fix once this merges.

PR #1488, opened 2026-07-06
Enabling EC Schema Evolution via Feature-Driven Architecture

RohitPtnkr1996:
A feature driven approach which will enable us to continue ecshema evolution beyond ECXml 3.2 where we are stuck right now.

# Profile Version 4.0.0.6 and `ec_Feature` table
The new profile version 4.0.0.6 introduces a new ec table `ec_Feature` which will get added to any new ECDb created or as part of the profile upgrade to 4.0.0.6.
The table is of the format:
```sql
CREATE TABLE ec_Feature (
    Name        TEXT PRIMARY KEY NOT NULL COLLATE NOCASE,  -- canonical string identifier (e.g. "json-primitive-type")
    Description TEXT NOT NULL,                             -- human-readable feature description (e.g. "JSON Primitive Types")
    Compat      TEXT NOT NULL,                             -- how older code should behave: "Warn", "ReadOnly", "NoSchemaImport", "NoChangesetGeneration" or "Refuse". Intentionally unconstrained, so a future runtime can persist a compat mode that doesn't exist yet.
    Fallback    TEXT NOT NULL COLLATE NOCASE               -- how a runtime that doesn't recognize `Compat` should degrade instead. Constrained to the closed set of modes {"Warn", "ReadOnly", "NoSchemaImport", "NoChangesetGeneration" and "Refuse"}
        CHECK(Fallback IN ('Warn','ReadOnly','NoSchemaImport','NoChangesetGeneration','Refuse'))
);
```

When ECDb opens a file, it reads the `ec_Feature` table.
For each row, it checks whether the feature name is known to the current software version:
- **Known**: proceed normally, the software supports this feature.
- **Unknown**: resolve the *effective* compat mode and act accordingly:
  - If `Compat` itself is a mode this runtime recognizes, that mode is used directly - `Fallback` is irrelevant in this case.
  - Otherwise, if `Fallback` is a mode this runtime recognizes, that mode is used instead.
  - Otherwise (neither is recognized - should not normally happen, since `Fallback` is DB-constrained to the known set), the runtime fails closed to `Refuse`.
- **Compat modes**
  - `Warn`: open the file normally, log a warning. The feature doesn't affect data integrity for older readers (e.g., a new informational schema attribute).
  - `ReadOnly`: open the file in read-only mode. The feature changes how data is stored or interpreted, and older writers could corrupt it, but reading is safe (e.g., JSON primitive types : older code sees plain strings).
  - `NoSchemaImport`: open the file in read-write, but a gate in the schema import logic will stop any incoming schemas. The file can otherwise be written to with no restrictions.
  - `NoChangesetGeneration`: open the file in read-write and it can be read/written to locally without restriction, but this runtime is forbidden from *publishing* a changeset from it. This protects other briefcases/runtimes downstream that may not tolerate the feature at all.
  - `Refuse`: refuse to open the file entirely, in both read-write and read-only modes. The feature changes the storage layout or data model in ways that older code cannot safely interpret even for reading (e.g., a new MapStrategy that reorganizes tables).

> Features are **independent dimensions**: an iModel can have any combination, consumers adopt incrementally, and optional/experimental changes don't force-upgrade the entire ecosystem. This gives us a more granular control over an ECDb file instead of a more binary approach of all-or-nothing.

## Protecting older writers

The `ReadOnly` compat mode prevents older software from writing invalid data through JSON properties it doesn't understand. This is the real risk raised in [#1943](https://github.com/iTwin/itwinjs-backlog/issues/1943) - without gating, older ECDb would happily write past validation CAs and corrupt the data. The `ec_Feature` table makes this safe: older code opens in read-only mode and can't corrupt the validated JSON.

## Revalidating after changeset operations

Enforcement is not limited to the initial `Db Open()` call. A pulled or reversed changeset can introduce or remove `ec_Feature` rows independently of how the file was originally opened, so feature validation is re-run:

- **After merging a changeset** (`TxnManager::MergeDataChanges`): if the merge introduced a feature this runtime cannot honor (`ReadOnly` or `Refuse`), the merge is **backed out** (the applied changes are reversed) rather than left half-applied, and `ChangesetStatus::UnknownFeatureAfterMerge` is returned to the caller. After backing out, `RevalidateFeatures()` is called again to refresh the cached restriction lists; if it *still* reports a non-`OK` status at that point, that indicates some other, unrelated feature is restricting this runtime (the backed-out merge itself is not the cause), which is logged for visibility.
- **After reversing a changeset** (`TxnManager::ReverseChangeset`): restrictions are recomputed, since reversing can remove the `ec_Feature` rows that had introduced them. This call happens *before* the reversed changes are saved, so if it comes back `BE_SQLITE_ERROR` (a `Refuse`-level feature is somehow still present), the reverse itself is abandoned and an exception is thrown rather than committing a state this runtime cannot honor at all. A `BE_SQLITE_READONLY` result is logged as a warning instead, since the in-memory connection can't be retroactively downgraded to true read-only.
- **After reverting/replaying timeline changesets** (`TxnManager::RevertTimelineChanges`): same recomputation, for the same reason - but here it runs *after* each changeset in the batch is already saved, so a non-`OK` result can only be logged, not backed out (earlier iterations in the same batch are already committed).

## Diagnostics

For tooling and diagnostics, we expose feature information via read-only PRAGMAs:

```sql
PRAGMA ecdb_known_features;   -- list all features the current software version knows about
PRAGMA ecdb_used_features;    -- list features present in this file's ec_Feature table
```

- We also expose `ECDb::TryGetBlockingFeatures(blockingFeatureNames, fileName)` - a **static** method that reads a file's `ec_Feature` rows via a plain, read-only `BeSQLite::Db` open, WITHOUT going through ECDb's own open-time validation. This exists specifically to diagnose a file this runtime already refused to open (`Compat` effectively resolves to `Refuse`): since a `Refuse`d file cannot be opened as an `ECDb` at all, this is the only way to report *which* feature caused the failure instead of a bare "could not open" error.

## Open discussion points

1. **Feature scope (maybe a `Scope` column ?)**: `ec_Feature` currently only supports file-wide features. There's no way to say "this feature only applies to schema X" or "this feature only applies to class Y". Every feature is all-or-nothing for the whole file.
2. **Consistency of `ReverseChangeset` vs. `RevertTimelineChanges` on a fatal post-reverse feature state**: `ReverseChangeset` aborts (abandon + throw) if revalidation comes back `Refuse`-level, since it can still be safely backed out at that point. `RevertTimelineChanges` can only log the same situation, since by the time it revalidates, the changeset in question is already committed. Should `ReverseChangeset` be simplified to log-only as well for consistency?
3. **Feature versions**: Should features be versioned ?
4. **Compat severity**: A feature applied to a branch new class might be okay (warn), but for an existing class, this might be more severe.
5. **Removing featues**: If a changeset/schema update removes a feature from use, how do we track and reconcile it? I have added a function pointer in the `FeatureInfo` struct for this. The runtime inserting this feature must also specify how this feature is to be removed, including any data.
6. **Stale feature rows from a future runtime**: A row written by a future runtime for a feature this runtime doesn't know about has no detector registered here at all, so if the current runtime's activity (schema import, local edits, or merging a changeset) happens to eliminate that feature's actual usage from the file, there is no code path that will ever delete the now-stale ec_Feature row.
7. **Features below EC Layer**: What if a feature works purely at the BeSqlite layer, and the ECDb's feature registration and tracking never gets hit?

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/2159


rschili:
Rohit just gave a presentation of this in our call and I have some immediate thoughts. Posting it here for tracking, it's more persistent than chat.

### Apply Changeset can make the currentl iModel unusable

I agree the check needs to happen during apply changeset, and rejecting the apply is probably the right move. Question is what to do if the incoming changeset would make the briefcase reasonly. Can we safely switch the iModel into read-only mode? The easiest approach is probably to just reject any apply changeset that restricts the use of the current briefcase.

### Profile update

Like I raised in the call, we have to be aware, this is a profile update with everything that comes with it. Until we have maintenance mode, these are painful.

I also have a few unrelated ideas that I'd like to include in a profile update and am wondering if we should include this here. We have no mechanism for combining multiple PRs for a single profile update....

My changes are small: I'd like to add a generation counter to the ec_Schema table, which increments every time a schema is modified. So it will be a number alongside the schema version. It will allow us to reliably track schema state as it changes. Especially since dynamic schemas do not increment their versions, this flag will be helpful for "you need to reload" mechanisms. Should be included in ECDbMeta as well, so it has some ripples.

### Compatibility tests

You brought up that these need a new layer and need work. I agree, we have neglected them for a long time, they are important, and we have limited resources. I'm not sure what the solution is, but yes, this is a problem we have to follow up on.

### Removing a feature

I think you've overcomplicated this. Is it a hard requirement by anybody?
Imho if you enable a feature, that is like running a profile update. You cannot undo that. Even if you change the iModel to a state where you could consider that feature "turned off", it's an odd mechanism.

If you choose to enable a feature for an iModel, I say the mechanism to disable it is to remove that changeset and everything that followed from your history. Automatically disabling features when they go unused seems really clunky and difficult, and distracts from the primary goal.

Maybe, if we absolutely need it, we could add a pragma that would re-evaluate and update feature use on demand.

### Lower level features like compressed changesets

It seems you thought a lot about this, but I'd roll with the simple answer: Anything that's lower level than EC, should not be a concern of the ec_Feature table. So yes, lower level stuff will not be covered by this, but I think that's fine. Let's avoid taking in additional complexity.

iModelCore/ECDb/ECDb/FeatureManager.h
```diff
@@ -41,19 +39,38 @@ enum class Feature
     {Feature::SystemPropertiesHaveIdExtendedType, ProfileVersion(4, 0, 0, 2)} \
     }
 
+enum class Compat { Warn, ReadOnly, NoSchemaImport, NoChangesetGeneration, Refuse };
```
rschili: I'm thinking we may want to cut 'NoChangesetGeneration' and 'NoSchemaImport' from this list. For schemas we already have 'ImportRequiresVersion' and 'UseRequiresVersion' which overlap with this, people will get confused about which to use when. For changeset generation: I think people would hate if they were able to make edits and at the end learn that they cannot push their changes. ReadOnly is the better flag for this purpose.
So I'm saying I believe less is more here.

iModelCore/ECDb/ECDb/FeatureManager.h
```diff
@@ -41,19 +39,38 @@ enum class Feature
...
+using FeatureDetector = bool (*)(ECDbCR);
+
+struct ECDB_EXPORT FeatureInfo
+    {
+    Utf8CP name;
+    Utf8CP description;
+    //! The precise compat mode. May be a mode that older runtimes do not recognize.
+    Compat compat;
+    //! For unknown compat, fallback to a known compat.
+    Compat fallback;
```
rschili, line 55: The idea with the fallback column is good, but the more I look at it the more it feels YAGNI.
We will probably be fine if we say "unknown Compat -> Refuse".

iModelCore/ECDb/ECDb/FeatureManager.h
```diff
@@ -41,19 +39,38 @@ enum class Feature
     {Feature::SystemPropertiesHaveIdExtendedType, ProfileVersion(4, 0, 0, 2)} \
     }
 
+enum class Compat { Warn, ReadOnly, NoSchemaImport, NoChangesetGeneration, Refuse };
+
+//! Function that returns true if the feature is currently in use in the given ECDb.
+//! Set to nullptr for write-once features (e.g. profile-upgrade changes)
+using FeatureDetector = bool (*)(ECDbCR);
```
rschili, line 46: I think right now there is no example of somebody setting this, so it's tough to judge...
Like I said in my PR comment, my impression still is that feature removal is overcomplicated and probably not necessary. (Detector, Reconcile, DeleteFeature). If you really accidentally use a feature in an iModel and it turns out you want to go back on that decision, you can remove the changeset from the history. That is the cleaner path backwards.

If you used a feature and no longer use it.... that's odd. Maybe you used a json property and removed it in a major schema upgrade. Still I'd argue if automatic removal buys us anything.
Maintaining and evaluating it is a cost that seems disproportionate to its usefulness.

iModelCore/Bentley/Bentley/nonport/BeStringUtilities.cpp
```diff
@@ -1332,19 +1332,33 @@ void BeStringUtilities::Split(Utf8CP str, Utf8CP delimiters, Utf8CP escapeChars,
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-template<typename T> static Utf8String joinStrings(bvector<T> const& strings, Utf8CP delim)
+template<typename T> static Utf8String joinStrings(bvector<T> const& strings, Utf8CP delim, bool addQuotes)
```
rschili: This seems way too specific to make it into BeStringUtilities.
It doesn't escape embedded quotes in the string.
I'd just build the string locally where you need this, I believe there is only one caller atm.

iModelCore/ECDb/ECDb/ECDb.cpp
```diff
@@ -249,6 +249,58 @@ ECN::IECSchemaLocater& ECDb::GetSchemaLocater() const { return m_pimpl->GetSchem
...
+//--------------------------------------------------------------------------------------
+// @bsimethod
+//---------------+---------------+---------------+---------------+---------------+------
+DbResult ECDb::RevalidateFeatures() const { return m_pimpl->ValidateECFeatures(); }
+
+//--------------------------------------------------------------------------------------
+// @bsimethod
+//---------------+---------------+---------------+---------------+---------------+------
+//static
+BentleyStatus ECDb::TryGetBlockingFeatures(std::vector<Utf8String>& blockingFeatureNames, BeFileNameCR fileName)
```
rschili, line 266: This seems redundant to the issue reporter.
Granted, it's more convenient, but I'd rather keep the issue reporter as the primary source and not have two ways to report this. We could always add this later if needed.

iModelCore/ECDb/ECDb/ProfileUpgrader.h
```diff
@@ -67,6 +67,12 @@ BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
                        "CREATE UNIQUE INDEX uix_ec_FormatCompositeUnit_FormatId_Ordinal ON " TABLE_FormatCompositeUnit "(FormatId,Ordinal);" \
                        "CREATE INDEX ix_ec_FormatCompositeUnit_UnitId ON " TABLE_FormatCompositeUnit "(UnitId);"
 
+#define TABLEDDL_Feature "CREATE TABLE " TABLE_Feature "(" \
+            "Name TEXT PRIMARY KEY NOT NULL COLLATE NOCASE," \
```
rschili, line 71: COLLATE NOCASE seems like the right choice to me, but this contradicts the s_knownFeatures which is case sensitive.

iModelCore/ECDb/ECDb/ECDb.cpp
```diff
@@ -249,6 +249,58 @@ ECN::IECSchemaLocater& ECDb::GetSchemaLocater() const { return m_pimpl->GetSchem
...
+
+    Db db;
+    if (BE_SQLITE_OK != db.OpenBeSQLiteDb(fileName, Db::OpenParams(Db::OpenMode::Readonly)))
+        return ERROR;
+
+    if (!db.TableExists(TABLE_Feature))
+        return SUCCESS;
+
+    Statement stmt;
+    if (BE_SQLITE_OK != stmt.Prepare(db, "SELECT Name, Compat, Fallback FROM main." TABLE_Feature))
```
rschili, line 278: This doesn't call IsFeatureKnown, it just reports every Refuse-Level feature, no?

iModelCore/ECDb/ECDb/ECDbImpl.cpp
```diff
@@ -218,14 +218,119 @@ DbResult ECDb::Impl::OnDbCreated() const
...
+                // The file itself is safe to read and write locally, but this runtime must not publish a changeset from it
+                m_featuresBlockingChangesetGeneration.push_back(featureName);
+                break;
+
+            case Compat::Refuse:
+                m_issueReporter.ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties,
+                    IssueType::ECDbIssue, ECDbIssueId::ECDb_0744,
+                    "ECDb file uses unknown feature '%s'. The file cannot be opened by this ECDb runtime.",
+                    featureName.c_str());
+                return BE_SQLITE_ERROR;
```
rschili: We should iterate everything and report ALL problems before returning ERROR

iModelCore/ECDb/ECDb/FeatureManager.cpp
```diff
@@ -52,4 +54,182 @@ bool FeatureManager::IsAvailable(ProfileVersion const& actualVersion, Feature fe
...
+    if (info == nullptr)
+        {
+        ecdb.GetImpl().Issues().ReportV(IssueSeverity::Error, IssueCategory::BusinessProperties, IssueType::ECDbIssue, ECDbIssueId::ECDb_0745,
+            "Feature '%s' is not a known feature and cannot be inserted into the database.", featureName.c_str());
+        return BentleyStatus::ERROR;
+        }
+
+    // Insert the row into the ECDb file
+    Statement stmt;
+    if (stmt.Prepare(ecdb, "INSERT OR IGNORE INTO " TABLE_Feature " (Name, Description, Compat, Fallback) VALUES (?,?,?,?)") != BE_SQLITE_OK)
```
rschili, line 176: If the row already exists with a different Compat, this would keep the old value. Probably fine, I wouldn't want to make this an error.... or should we overwrite the column using upsert? Seems better to me.

iModelCore/ECDb/ECDb/ProfileUpgrader.h
```diff
@@ -67,6 +67,12 @@ BEGIN_BENTLEY_SQLITE_EC_NAMESPACE
                        "CREATE UNIQUE INDEX uix_ec_FormatCompositeUnit_FormatId_Ordinal ON " TABLE_FormatCompositeUnit "(FormatId,Ordinal);" \
                        "CREATE INDEX ix_ec_FormatCompositeUnit_UnitId ON " TABLE_FormatCompositeUnit "(UnitId);"
 
+#define TABLEDDL_Feature "CREATE TABLE " TABLE_Feature "(" \
```
rschili, line 70: Tracking for me.... I think the new schema sync v2 should skip this table, and we should calculate it locally on each briefcase. Otherwise we'd sync enabled features into briefcases which may not want them...

rschili (commented):
The core looks right to me, sorry I haven't given this a lot of attention yet. We need to get it absolutely right before we roll it out, we cannot un-roll it.

iModelCore/ECDb/ECDb/FeatureManager.h
RohitPtnkr1996: 'NoChangesetGeneration' was just an idea I had which I didn't give too much thought over.
I'm okay to remove that.

As for 'NoSchemaImport', I believe we do have some need of it at the very least.
The two CAs communicate the runtime that they require, whereas the compat tells us what is already in the file (a feature has modified the importer in a way that might corrupt the import itself).
Also, the two CAs are profile-version gated, which we want to try to step away from with features.

iModelCore/ECDb/ECDb/FeatureManager.h
RohitPtnkr1996: This was another idea I had in vitro.
I think it all hinges on how much we expect the compat vocab to change.
Because once this is out there, there is no going back without a very ugly profile version revert.
If we're sure these compats can and will satisfy all the needs of now and the future, then we can remove the fallback column.

iModelCore/ECDb/ECDb/FeatureManager.h
RohitPtnkr1996: You point totally makes sense, I'm just a bit scared of the consequence of having a file locked off from old readers even if it no longer is using said feature.
I thought about just having a pragma like you suggested in your PR comment, but it seems a bit dangerous. The sql will end up in a changeset by itself.
Reversing the changeset does seem like the cleaner path.

iModelCore/Bentley/Bentley/nonport/BeStringUtilities.cpp
RohitPtnkr1996: I'll move this over to the feature code.

iModelCore/ECDb/ECDb/ECDb.cpp
RohitPtnkr1996: Yup you're right, I'll fix it.

iModelCore/ECDb/ECDb/FeatureManager.cpp
RohitPtnkr1996: I'd say this goes back to our discussion about compat vocab evolving.
A divergent compat means the current runtime has an updated version of the compat from the one originally inserted in the file.
Hypothetically, if we had mutiple briefcases at different compats for the same feature, the last one writing the upsert will win, creating more changeset churn.

This was the reason I had a discussion point in the ppt about feature versioning.

iModelCore/ECDb/ECDb/FeatureManager.h
rschili: you are right, independent of your comment it just occurred to me, we probably want to have that one (NoSchemaImport).

iModelCore/ECDb/ECDb/FeatureManager.h
rschili: For the upcoming EC 3.3 changes I have been wondering if we should roll them out through a profile upgrade, or on demand through the feature table. Some of those will have to add columns to the ec_ table. Which means if we model that as a feature, that feature needs an "upgrade", requires a full schema lock, and also has to update the schemasync db ec_ tables if schemasync is enabled.
Which makes this even harder to roll back.
We probably have to think a bit more about these kinds of features..

iModelCore/ECDb/ECDb/ProfileUpgrader.h
rschili: Now that schema sync has been merged I'm still not sure about this.
We could consider enabling a feature an upgrade which means you have to push immediately and use a full lock for it.
Or we only track the table in the briefcase and blacklist it in schemasync. Both have their tradeoffs. I lean towards not putting it into schemasync.

PR #1491, opened 2026-07-07
Add support for concurrent definition "reservations" via SchemaSync

wgoehrig:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9483

The native changes for this are minimal compared to the [TypeScript PR](https://github.com/iTwin/itwinjs-core/pull/9483), I just needed to:
- Add `options` to `OnElementPropsArg` and allow `onInsert` to mutate those options
- Fix a typo in `BeBriefcaseId::LastValidBriefcaseId`
- Add `BeBriefcaseId::SchemaSyncElementReserved` (mirrors [BriefcaseIdValue.SchemaSyncElementReserved](https://github.com/iTwin/itwinjs-core/blob/wbg/component-schemasync/core/common/src/BriefcaseTypes.ts#L42C3-L42C20) in @itwinjs/core-common)
  - This is not actually used by the native code, I just added for consistency


kabentley:
I suggest we use the term "synchronous channel" somehow instead of "schema sync" so it's clear this is about synchronizing concurrent additions to the shared part of an iModel. Terminology is hard but important and worth debating.

rschili (approved):

khanaffan (approved):

RohitPtnkr1996 (approved):

PR #1500, opened 2026-07-10
Add workflow to sync iTwin.js core dependencies

DanRod1999:
## Summary

Adds a GitHub Actions workflow (`.github/workflows/sync-itwinjs-core-deps.yml`) that automatically updates `@itwin/*` devDependencies when itwinjs-core publishes a new release.

Companion to https://github.com/iTwin/itwinjs-core/pull/9498

## How it works

1. **Trigger**: itwinjs-core dispatches a `repository_dispatch` event of type `itwinjs-core-released` with `next_dev_version`, `tag`, and `release_url` in the payload
2. **Update**: The workflow updates these devDependencies in `iModelJsNodeAddon/api_package/ts/package.json`:
   - `@itwin/build-tools`
   - `@itwin/core-bentley`
   - `@itwin/core-common`
   - `@itwin/core-geometry`
3. **Lock file**: Regenerates `package-lock.json` via `npm install --ignore-scripts`
4. **PR**: Creates (or updates) a PR targeting `main` with git identity `imodeljs-admin`

## Security considerations

- All `${{ }}` expressions are routed through `env:` blocks to prevent expression injection
- Explicit `permissions: contents: write, pull-requests: write`
- Concurrency group prevents race conditions
- Idempotent: uses `git checkout -B` + `--force-with-lease` and guards `gh pr create` with an existence check

## Testing

The workflow also supports `workflow_dispatch` for manual testing with the same inputs.


.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,145 @@
...
+          else
+            echo "has_changes=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Commit and push
+        if: steps.changes.outputs.has_changes == 'true'
+        run: |
+          git add "$PACKAGE_DIR/package.json" "$PACKAGE_DIR/package-lock.json"
+          git commit -m "Update iTwin.js packages to ^${NEXT_DEV_VERSION}"
+          git push --force-with-lease origin "$BRANCH_NAME"
```
aruniverse: `--force-with-lease` here will likely be rejected on any run after this branch has been pushed once. `actions/checkout@v4` (line 48) only fetches the default branch — it never establishes a local `refs/remotes/origin/bot/sync-itwinjs-core-deps` tracking ref, so `--force-with-lease` has no expected-SHA basis for this branch. If a second release fires while the previous auto-PR from this workflow is still open, this push fails, and since there's no `continue-on-error`, the "Create or update pull request" step below is skipped too — that release's dependency bump is silently dropped with no PR update and no alert.

Suggested fix: `git fetch origin "$BRANCH_NAME" || true` before `checkout -B` (line 62) so the lease has a real basis, or use a plain `--force` push here since this is a single-writer, bot-owned branch already serialized by the `concurrency:` group (lines 26-28).

.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,145 @@
...
+              '@itwin/build-tools',
+              '@itwin/core-bentley',
+              '@itwin/core-common',
+              '@itwin/core-geometry',
+            ];
+            for (const name of packages) {
+              console.log('Updating ' + name + ' to ' + version);
+              pkg.devDependencies[name] = version;
+            }
+            fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n');
```
aruniverse: `iModelJsNodeAddon/api_package/ts/package.json` is CRLF at HEAD; `JSON.stringify(pkg, null, 2) + '\n'` emits LF, so this rewrites the entire file (~40 lines) instead of producing the real 4-line devDependency diff on every automated run. I reproduced this against the current file content. This is the same failure mode that caused a prior manual attempt at this automation to be reverted (imodel-native#1443) — it buries the real diff in line-ending noise and risks the same reviewer/CI confusion that led to that revert.

Suggested fix: preserve the original line endings (detect + re-apply CRLF after stringify, or do a targeted string/regex edit on the devDependencies block instead of a full `JSON.parse`/`JSON.stringify` round-trip).

.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,145 @@
...
+  group: sync-itwinjs-core-deps
+  cancel-in-progress: true
+
+jobs:
+  update-deps:
+    runs-on: ubuntu-latest
+    env:
+      NEXT_DEV_VERSION: ${{ github.event.client_payload.next_dev_version || github.event.inputs.next_dev_version }}
+      TAG: ${{ github.event.client_payload.tag || github.event.inputs.tag }}
+      RELEASE_URL: ${{ github.event.client_payload.release_url || github.event.inputs.release_url }}
```
aruniverse, line 36: `NEXT_DEV_VERSION` is validated non-empty below (line 42-45), but `TAG` and `RELEASE_URL` aren't. If the companion itwinjs-core workflow (itwinjs-core#9498) ever renames one of these payload keys, this workflow would currently proceed with blank `TAG`/`RELEASE_URL` values and produce a PR with cosmetically broken title/body text rather than failing loudly.

Suggested fix: add the same kind of non-empty guard for `TAG` alongside the existing `NEXT_DEV_VERSION` check, so a future key mismatch between the two workflows fails loudly instead of silently.

.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,145 @@
...
+          node -e "
+            const fs = require('fs');
+            const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+            const version = '^' + process.env.NEXT_DEV_VERSION;
+            const packages = [
+              '@itwin/build-tools',
+              '@itwin/core-bentley',
+              '@itwin/core-common',
+              '@itwin/core-geometry',
+            ];
```
aruniverse, line 83: Optional: this 4-package list is hardcoded a second time in the PR-body table further down (around line 126-129). If a 5th package is ever added to this automation, both lists have to be updated in lockstep with no shared source. Consider generating the PR-body table from this same `packages` array so there's one source of truth.

.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,145 @@
...
+        description: "GitHub release URL"
+        required: false
+        default: ""
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: sync-itwinjs-core-deps
```
aruniverse, line 27: Optional: this concurrency group is shared by both `repository_dispatch` and `workflow_dispatch` with `cancel-in-progress: true`. A manual `workflow_dispatch` test run queued while a real release-triggered run is in flight would cancel the real run rather than queue behind it. Low likelihood since both triggers require elevated access, but consider a per-version group key (e.g. include `NEXT_DEV_VERSION` in the group name) or `cancel-in-progress: false` if this is a concern.

aruniverse (commented):
Reviewed alongside the companion PR (iTwin/itwinjs-core#9498) using three independent passes: correctness, security, and operational robustness. Security posture is solid — verified line-by-line that all `${{ }}` expressions are routed through `env:` blocks, and could not construct a working shell/expression-injection vector even with a maliciously crafted tag containing `$()`. Left inline comments on two things worth fixing before this runs in production (a `--force-with-lease` push that will likely fail on repeat runs, and a CRLF/LF rewrite reproduced against the real target file), plus a few lower-priority suggestions.

.github/workflows/sync-itwinjs-core-deps.yml
aruniverse: Addressed in ad8fdcaa ("Add workflow to sync iTwin.js core dependencies", pushed 2026-07-14 17:33 UTC) — the push now uses a plain `git push --force origin "$BRANCH_NAME"` (line 115) instead of `--force-with-lease`, which sidesteps the missing-tracking-ref problem entirely since the branch is single-writer/bot-owned and already serialized by the concurrency group. Combined with the concurrency group now queueing instead of cancelling (see below), two close-together releases will run sequentially rather than racing or silently dropping a push. Confirmed by re-reading the current file.

.github/workflows/sync-itwinjs-core-deps.yml
aruniverse: Addressed in ad8fdcaa (2026-07-14 17:33 UTC) — the script now detects the source file's original EOL (`content.includes('\\r\\n')`, line 75) and re-applies it to the stringified output (line 88) instead of hardcoding `\\n`. I re-verified against the real file: with this change, only the 4 intended devDependency lines change and the rest of the file's CRLF endings are preserved. I also separately checked the lockfile regen step for the same class of issue — ran `npm install --ignore-scripts --package-lock-only` against the real `package-lock.json` (also 100% CRLF) with a real version bump, and npm's own writer preserves the existing CRLF endings there too, so no equivalent fix is needed on the lockfile side.

.github/workflows/sync-itwinjs-core-deps.yml
aruniverse: Addressed in ad8fdcaa (2026-07-14 17:33 UTC) — a `[ -z "$TAG" ]` guard was added alongside the existing `NEXT_DEV_VERSION` check (lines 46-49), so a future payload-key mismatch on the itwinjs-core side now fails the job loudly instead of producing a PR with a blank title/body field. (`RELEASE_URL` remains unvalidated, but that's only ever used for an optional cosmetic line in the PR body, so no action needed there.)

.github/workflows/sync-itwinjs-core-deps.yml
aruniverse: Addressed in ad8fdcaa (2026-07-14 17:33 UTC) — the devDependencies step now emits a `packages_table` output built from the same `packages` array (lines 91-93) and the PR-body step consumes it via `${{ steps.update.outputs.packages_table }}` (line 121, interpolated at line 139) instead of a second hardcoded list. Single source of truth now, confirmed by re-reading the current file.

.github/workflows/sync-itwinjs-core-deps.yml
aruniverse: Addressed in ad8fdcaa (2026-07-14 17:33 UTC) — changed to `cancel-in-progress: false` (line 28), so an overlapping `workflow_dispatch` test run now queues behind an in-flight release-triggered run instead of cancelling it. This also nicely reinforces the force-push fix above by keeping runs strictly sequential.

.github/workflows/sync-itwinjs-core-deps.yml
```diff
@@ -0,0 +1,155 @@
...
+          fi
+          echo "Updating iTwin.js deps to ^${NEXT_DEV_VERSION} (triggered by ${TAG})"
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
```
aruniverse: we should prob use a newer node ver than 20

aruniverse (approved):

PR #1512, opened 2026-07-15
Implement SQLite change tracking and changeset management in SQLiteDb

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9509


iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -800,15 +812,101 @@ struct SQLiteDb : Napi::ObjectWrap<SQLiteDb>, SQLiteOps<Db> {
...
+
+    //! Write out the changes captured since startChangeTracking() was called, to a changeset
+    //! file in the same on-disk format used for iModel changesets. Used only for testing.
+    void CreateChangeset(NapiInfoCR info) {
+        auto& db = GetWritableDb(info);
+        REQUIRE_ARGUMENT_STRING(0, pathname);
+        if (m_changeTracker.IsNull())
+            THROW_JS_IMODEL_NATIVE_EXCEPTION(info.Env(), "change tracking was not started", IModelJsNativeErrorKey::BadArg);
+
+        m_changeTracker->EnableTracking(false);
```
aruniverse, line 869: `EnableTracking(false)` changes persistent tracker state, but both `writer.Initialize()` and `FromChangeTrack()` can throw below. In either failure path `EndTracking()` is skipped, so a caller that fixes the output path and retries without calling `startChangeTracking()` will silently omit all subsequent changes because the session remains disabled. Please use scope-based cleanup to restore tracking for a retry, or consistently end/unregister the tracker before propagating the error.

iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -800,15 +812,101 @@ struct SQLiteDb : Napi::ObjectWrap<SQLiteDb>, SQLiteOps<Db> {
...
+        bool containsSchemaChanges = false;
+        BeSQLite::DdlChanges ddlChanges;
+        auto stat = changesetReader.MakeReader()->GetSchemaChanges(containsSchemaChanges, ddlChanges);
+        if (stat != BE_SQLITE_OK)
+            THROW_JS_BE_SQLITE_EXCEPTION(info.Env(), "error reading changeset file", stat);
+
+        if (containsSchemaChanges && !ddlChanges._IsEmpty()) {
+            stat = db.TryExecuteSql(ddlChanges.ToString().c_str());
+            if (stat != BE_SQLITE_OK)
+                THROW_JS_BE_SQLITE_EXCEPTION(info.Env(), "error applying ddl changes from changeset", stat);
```
aruniverse, line 904: Could this rollback before throwing? `TryExecuteSql` may execute an earlier DDL statement before a later statement in this concatenated DDL payload fails. Unlike the DML failure path below, this branch does not call `AbandonChanges()`, leaving the caller able to `saveChanges()` and persist a partially applied changeset. Please roll back the default transaction on this error path (preferably with shared cleanup for both DDL and DML failures).

iModelJsNodeAddon/IModelJsNative.cpp
aruniverse: Verified in 38d6aaf: `endTracking` is installed before tracking is suspended, so both writer failure paths now call `EndTracking()`. This resolves the disabled-tracker retry issue.

iModelJsNodeAddon/IModelJsNative.cpp
aruniverse: The new guard fixes the originally reported partial-DDL path when a default transaction is active. There is still a follow-up atomicity/isolation issue with using `AbandonChanges()` as the rollback mechanism; I have left a separate inline comment on the guard.

iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -800,15 +828,112 @@ struct SQLiteDb : Napi::ObjectWrap<SQLiteDb>, SQLiteOps<Db> {
...
+        auto stat = changesetReader.MakeReader()->GetSchemaChanges(containsSchemaChanges, ddlChanges);
+        if (stat != BE_SQLITE_OK)
+            THROW_JS_BE_SQLITE_EXCEPTION(info.Env(), "error reading changeset file", stat);
+
+        // Roll back any partially-applied changes (DDL or DML) on failure. TryExecuteSql may commit
+        // some statements of the concatenated DDL payload before a later one fails, and ApplyChanges
+        // may fail partway through the row-level changes. In either case the caller must not be left
+        // able to saveChanges() and persist a partially-applied changeset.
+        bool applied = false;
+        SQLiteDbScopeGuard rollbackOnError([&]() { if (!applied) db.AbandonChanges(); });
```
aruniverse: `AbandonChanges()` is not scoped to this apply: on a normal default transaction it discards the caller’s earlier unsaved work too, and with `defaultTxn: No` it cannot roll back a multi-statement DDL payload whose first statement has already autocommitted. Please bracket the DDL+DML apply in an explicit savepoint/transaction owned by this method, committing only after `ApplyChanges` succeeds and rolling back/releasing it on every failure. Add coverage for both an existing uncommitted target change and `defaultTxn: No` with a later failing DDL statement.

khanaffan: This all this is beyond scope of this feature.  It's suppose to be used in checkpoint service and they would not have uncommitted work. If changeset fail there job just fails. 

Ideally we should not even calling abandonChanges() this should be the caller that does that. They can choose what transaction model they want. If they want to make sure they have no uncommitted changes before calling applyChangeset() or they create save point etc. 

Instead of making it more safe I would change it so its Caller responsibility since SqlDb is at very low level we cannot enforce any rules because we do not know what user is doing with it.

aruniverse (approved):

soham-bentley (approved):

PR #1513, opened 2026-07-16
Fix vcpkg Windows clang

tcobbs-bentley:
The old setup was using MSVC to compile the vcpkg libraries even during a Windows clang build. This is designed to fix that.


tcobbs-bentley:
The latest commit fixes the pipeline, and I confirmed that all the vcpkg-built libraries except crashpad are now built using clang. Unfortunately, crashpad it doesn't use cmake, and the changes didn't work for it, so it is still build using MSVC on the Win64 Clang build. I am fixing that next.

tcobbs-bentley:
Crashpad now builds with clang instead of MSVC in the Windows clang build.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -173,6 +173,11 @@ set "OVERLAY_TRIPLETS=%MANIFEST_DIR%\triplets"
 set "OVERLAY_ARG="
 if exist "%OVERLAY_TRIPLETS%" set "OVERLAY_ARG=--overlay-triplets=%OVERLAY_TRIPLETS%"
 
+rem Allow a manifest to ship overlay ports (e.g. a locally-patched crashpad that
+rem builds with clang-cl) in a "ports" subdirectory alongside vcpkg.json.
+set "OVERLAY_PORTS=%MANIFEST_DIR%\ports"
+if exist "%OVERLAY_PORTS%" set "OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports=%OVERLAY_PORTS%"
```
aruniverse: The new overlay port is only forwarded by this Windows wrapper. `vcpkg_install_crashpad.mke` uses `vcpkg_run_install.sh` for Linux, macOS, and Android, but that wrapper passes only `--overlay-triplets`; those builds therefore resolve the registry `crashpad` port instead of `crashpad/ports/crashpad/`. That bypasses this port’s Linux/Android patches, pinned sources, and library-name/configuration work. Please mirror the optional `MANIFEST_DIR/ports` detection and `--overlay-ports="$OVERLAY_PORTS"` argument in `vcpkg_run_install.sh`.

aruniverse: Please quote the overlay argument values. A checkout/manifest path containing spaces is valid on Windows, but `--overlay-ports=%OVERLAY_PORTS%` is split by `cmd.exe`, so vcpkg receives an invalid extra argument and cannot load this new overlay. Use `--overlay-ports="%OVERLAY_PORTS%"`; please also quote the analogous triplets argument while touching this construction.

tcobbs-bentley: This is intentional. We only need all the extra setup in ports for the Windows clang build. The rest were already working without that, and we don't want to change them.

tcobbs-bentley: Fixed.

aruniverse (approved):

iModelCore/libsrc/crashpad/ports/crashpad/crashpad-memset-errors-5758170.diff
```diff
@@ -0,0 +1,45 @@
+From 575817027d03ecce3c5d9ee0991a04e00babae3b Mon Sep 17 00:00:00 2001
```
tcobbs-bentley, line 1: These new files leave the documentation out of date for future version updates.

tcobbs-bentley: Fixed.

PR #1514, opened 2026-07-16
Adding schema reservation mechanism to schema sync

soham-bentley:
Fixes https://github.com/iTwin/itwinjs-backlog/issues/2192
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9513

## Problem

When two briefcases import the same schema offline, `IdFactory` seeds each `ec_*` id sequence from
the local `MAX(Id)+1`. Because briefcases can diverge from each other, they hand out different ids
for the same content. This makes conflict-free merges impossible.

The same problem applies to **shared-column slot assignments** in `TablePerHierarchy` / `ShareColumns`
hierarchies: each briefcase computes overflow and slot reuse independently from its own `ec_ClassMap` /
`ec_Column` state, so two briefcases starting from even slightly different local states assign
different `ec_Column.Id` and `ec_Column.Ordinal` to the same property. When those change-sets are
later merged the column tables diverge and the db is inconsistent.

---

## Solution

Extend `IdFactory` and `SchemaSync` so that, when SchemaSync is active, every `ec_*` metadata id
and every shared-column assignment is resolved by **content key** from a reservation store in the
sync db — rather than from a local `MAX(Id)`. Two briefcases that reserve the same schema therefore
produce identical ids and identical column assignments, making merges collision-free.

---

### Content-key scheme

Every `ec_*` metadata row is identified by a stable, content-derived string key — the same key is
used by both the reservation-write path and the consume path, so keys match byte-for-byte. Keys
compare case-insensitively (`CompareIUtf8Ascii`).

| Table | Key |
| --- | --- |
| `ec_Schema` | `schemaName` |
| `ec_SchemaReference` | `schemaName→referencedSchemaName` |
| `ec_Class` | `schemaName:className` |
| `ec_ClassHasBaseClasses` | `schemaName:className→baseClassFullName` |
| `ec_Property` | `schemaName:className:propertyName` |
| `ec_Enumeration` / `ec_KindOfQuantity` / `ec_UnitSystem` / `ec_Phenomenon` / `ec_Unit` / `ec_Format` / `ec_PropertyCategory` | `schemaName:name` |
| `ec_FormatCompositeUnit` | `schemaName:formatName:ordinal` |
| `ec_RelationshipConstraint` | `schemaName:relClassName:end(0\|1)` |
| `ec_RelationshipConstraintClass` | `schemaName:relClassName:end:constraintClassFullName` |
| `ec_CustomAttribute` | `containerKey:caClassFullName` |

For the column reservation store, the key per physical column is `schemaName:className:accessString`
(derived by `SchemaWriter::DerivePropertyColumnKey` / `SchemaReservationHelper::DeriveColumnKey`),
where `accessString` identifies one physical leaf of the property (e.g. `PropName`, `Geo.X`,
`Struct.Field`). Multi-column properties (`Point2d`, `Point3d`, structs) produce one key per leaf.

### Sync-db store schema (created by native during `SchemaSync::Init`)

```sql
-- One row per ec_* metadata table; KeyMap is a FlexBuffer blob: key → id (uint64).
CREATE TABLE IF NOT EXISTS [schema_reservation_ids] (
    [TableName]      TEXT    NOT NULL PRIMARY KEY,
    [LastReservedId] INTEGER NOT NULL DEFAULT 0,
    [KeyMap]         BLOB
);

-- One row per physical (TPH root/overflow) table; KeyMap: propertyKey → [columnOrd, columnId].
-- The high-water mark (lastUsedColumnOrd) is NOT stored as a column — it is derived as
-- max(columnOrd) across the table's KeyMap entries when the store is loaded into memory.
CREATE TABLE IF NOT EXISTS [schema_reservation_columns] (
    [PhysicalTableName] TEXT NOT NULL PRIMARY KEY,
    [KeyMap]            BLOB
);
```

#### Init-time seeding (`SchemaSync::Init` → `SeedReservationStoreInternal`)

The reservation stores are seeded **once** from the container's base db immediately after
`PushInternal` completes, so that `ReserveSchemaImport` never needs to re-read the local `MAX(Id)`
from any individual briefcase:

```
SeedReservationStoreInternal(syncDbUri)
  ├─ opens m_pendingReservationDb (RW)
  ├─ CREATE TABLE IF NOT EXISTS schema_reservation_ids / schema_reservation_columns
  ├─ SchemaReservationHelper::SeedReservationStoreFromLocalDb(m_conn, resStore)
  │     for each schema in localDb.Schemas().GetSchemas(true):
  │       SeedSchemaFromLocalDb(localDb, schema, store, visited)
  │         — derives key and records key→persistedId using ECObject::GetId() for
  │           id-bearing elements and targeted SQL queries for link/join rows:
  │           LookupSchemaReferenceId, LookupClassHasBaseClassesId,
  │           LookupFormatCompositeUnitId, LookupRelConstraintId,
  │           LookupRelConstraintClassId, LookupCustomAttributeId
  │     SeedLastReservedIdsFromLocalDb(localDb, store)   [seeds MAX(Id) counters]
  └─ SchemaReservationHelper::SeedColumnStoreFromLocalDb(m_conn, resStore, colStore)
        SeedLastUsedColumnOrdsFromLocalDb(localDb, colStore)   [MAX(Ordinal) per table]
        SeedColumnKeyMapsFromLocalDb(localDb, idStore, colStore)
          — SQL join over ec_PropertyMap/ec_PropertyPath/ec_Property/
            ec_Class/ec_Schema/ec_Column/ec_Table for Primary and Overflow tables
          — records propertyKey → (columnOrd, columnId) and mirrors columnId into
            idStore.column
  CommitPendingReservation()   [SaveChanges + CloseDb]
```

This means the stores reflect the **exact container baseline** at `Init`. Every subsequent briefcase
reads the already-seeded stores and only adds keys for genuinely new rows — independent of that
briefcase's local `MAX(Id)`.

#### Reserve path (called internally at the top of `ImportSchemas`, under the sync-db write-lock)

```
SchemaSync::ReserveSchemaImport(schemas, syncDbUri)
  ├─ Phase 1 — metadata ids
  │   ├─ opens m_pendingReservationDb (RW); CREATE TABLE IF NOT EXISTS reservation tables
  │   ├─ LoadReservationStoreFromSyncDb → SchemaReservationStore
  │   ├─ WalkSchemaForReservation(schema, store)  [for each schema]
  │   │     derive content key for every ec_* metadata row the import would produce
  │   │     for each key not already present: id = ++lastReservedId
  │   └─ WriteReservationStoreToSyncDb
  └─ Phase 2 — column assignments
      ├─ LoadColumnStoreFromSyncDb → SchemaReservationColumnStore
      ├─ CollectClassIndex(schema, classIndex)  [for each schema]
      │     builds classKey → ECClassCP index for ancestor/descendant occupant checks
      ├─ WalkSchemaForColumnReservation(schema, store, colStore, classIndex)  [for each schema]
      │     for each owned non-navigation property in a TPH shared-column class:
      │       derive leaf columnKeys; if not yet in colStore:
      │         scan existing slots for a reusable slot (no ancestor/descendant occupant);
      │         reuse or extend (columnOrd = ++lastUsedColumnOrd, derived from KeyMap max);
      │         columnId = idStore.column.GetOrAllocate(propertyKey)
      │         record propertyKey → (columnOrd, columnId)
      └─ WriteColumnStoreToSyncDb
  (no savepoint, no schema import, no rollback)
  SaveChanges deferred — caller commits via CommitPendingReservation()
                         or abandons via AbandonPendingReservation()
```

`ReserveSchemaImport` is a **pure read-modify-write** of the sync-db: it does not read the local
briefcase's `MAX(Id)` or `MAX(Ordinal)` — those baselines are set at `Init`. The `classIndex` built
by `CollectClassIndex` drives the ancestor/descendant test that decides whether a slot can be reused
across sibling classes (mirroring `ClassMapColumnFactory::ReuseOrCreateSharedColumn` /
`IsColumnUsedByAnyDerivedClass` without any live SQL query).

#### Consume path (inside `ImportSchemas`, after reservation)

```
MainSchemaManager::ImportSchemas
  ├─ ReserveSchemaImport(schemas, syncDbUri)   [under sync-db write-lock; see above]
  ├─ LoadReservationStore(resolvedSyncDbUri) → SchemaReservationStore
  ├─ IdFactory::SetKeyedMode(store)        [all sequences now in keyed mode]
  ├─ LoadColumnStore(resolvedSyncDbUri) → SchemaReservationColumnStore
  │   (stored on SchemaImportContext via AllocateColumnStore / GetColumnStore)
  ├─ SchemaWriter::ImportSchemas
  │   └─ every NextId() → IdSequence::NextIdForKey(contentKey)
  │         key present  → return reserved id
  │         key absent   → return BeInt64Id(0) → BindId binds NULL → INSERT fails
  └─ MapSchemas (DbMappingManager / ClassMapColumnFactory)
      └─ AllocateSharedColumn checks ctx.GetColumnStore() first:
           reserved entry found → DbTable::AddSharedColumn(DbColumnId(entry->columnId))
           no entry found       → fall back to normal ReuseOrCreateSharedColumn heuristic
```

---

## Key Properties

- **No deterministic import order required.** Keys are content-based so iteration order in
  `SchemaWriter`, `DbMappingManager`, and `RemapManager` is irrelevant.
- **Idempotent reservation.** `GetOrAllocate` is a no-op for keys already in the store; a second
  `ReserveSchemaImport` for the same schema advances no counters.
- **Backward compatible.** SchemaSync disabled or schema lock held → existing local-`MAX` path is
  untouched.
- **Columns reserved too.** Each new property gets the same `(physical table, column slot)` on
  every briefcase; slot reuse follows the same sibling/ancestor/descendant logic as
  `ClassMapColumnFactory::ReuseOrCreateSharedColumn`. Freed columns are never reused for reserved
  imports.
- **Init-seeding records key→persistedId.** Not just the `MAX(Id)` counter — the full `key → id`
  map and `propertyKey → (columnOrd, columnId)` map for all rows already in the base db are
  captured. A later import that touches a pre-existing element resolves to its already-persisted id
  with no counter bump.
- **High-water mark derived, not stored.** The column store has no `LastUsedColumnOrd` column.
  The high-water mark is computed as `max(columnOrd)` across a table's `KeyMap` entries when the
  store is loaded into memory — one extra O(N) in-memory pass, no extra I/O.
- **`overrideContainer` handled.** When `Init` is called with `overrideContainer=true`,
  `SeedReservationStoreInternal` still snapshots the current local baseline into the new container.
- **Transaction discipline.** `ReserveSchemaImport` leaves the sync-db write transaction open.
  `ImportSchemas` commits it via `CommitPendingReservation` on success or abandons it via
  `AbandonPendingReservation` on failure.

---

## Files Changed

| File | Change |
| --- | --- |
| [iModelCore/ECDb/ECDb/ECDbImpl.h](../../iModelCore/ECDb/ECDb/ECDbImpl.h) | `IdSequence::NextIdForKey(key)` keyed lookup; `SetKeyedMode` / `ClearKeyedMode` per-sequence and on `IdFactory` |
| [iModelCore/ECDb/ECDb/SchemaWriter.cpp](../../iModelCore/ECDb/ECDb/SchemaWriter.cpp) | All `NextId()` calls → `NextIdForKey(DeriveXKey(…))`; 16 key-derivation functions |
| [iModelCore/ECDb/ECDb/SchemaSync.h](../../iModelCore/ECDb/ECDb/SchemaSync.h) | Declare all new helpers on `SchemaReservationHelper`; DDL constants; `RES_TABLE_*` name constants; link-row lookup helpers; `CommitPendingReservation` / `AbandonPendingReservation`; `SeedReservationStoreInternal` |
| [iModelCore/ECDb/ECDb/SchemaSync.cpp](../../iModelCore/ECDb/ECDb/SchemaSync.cpp) | Implement `SeedSchemaFromLocalDb`, `SeedReservationStoreFromLocalDb`, `SeedColumnKeyMapsFromLocalDb`, `SeedColumnStoreFromLocalDb`; `SeedReservationStoreInternal`; `ReserveSchemaImport` (pure read-modify-write, two-phase); `CommitPendingReservation` / `AbandonPendingReservation`; `WalkSchemaForReservation`; `WalkSchemaForColumnReservation`; `CollectClassIndex`; `FindPrimaryTableForClass`; all link-row lookup helpers; seed step wired into `Init` |
| [iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp](../../iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp) | `ImportSchemas` calls `ReserveSchemaImport` → `CommitPendingReservation` / `AbandonPendingReservation`; `LoadReservationStore` → `SetKeyedMode`; `LoadColumnStore` |
| [iModelCore/ECDb/ECDb/SchemaImportContext.h](../../iModelCore/ECDb/ECDb/SchemaImportContext.h) | `AllocateReservationStore()`, `AllocateColumnStore()`, `GetColumnStore()` carrying both stores for the import lifetime |
| [iModelCore/ECDb/ECDb/ClassMapColumnFactory.cpp](../../iModelCore/ECDb/ECDb/ClassMapColumnFactory.cpp) | `AllocateSharedColumn` checks `ctx.GetColumnStore()` before falling back to heuristic; creates shared column with pre-assigned id via `DbTable::AddSharedColumn(DbColumnId(entry->columnId))` |
| [iModelCore/ECDb/ECDb/DbSchema.cpp / .h](../../iModelCore/ECDb/ECDb/DbSchema.cpp) | `DbTable::AddSharedColumn(DbColumnId)` — creates an `ec_Column` with a pre-assigned id |
| [iModelCore/ECDb/PublicAPI/ECDb/SchemaManager.h](../../iModelCore/ECDb/PublicAPI/ECDb/SchemaManager.h) | `SchemaReservationTableStore`, `SchemaReservationStore`, `SchemaReservationColumnTableStore`, `SchemaReservationColumnStore`, `SchemaReservationColumnEntry` data model |
| [iModelCore/ECDb/Tests/NonPublished/SchemaImportReservationTest.cpp](../../iModelCore/ECDb/Tests/NonPublished/SchemaImportReservationTest.cpp) | 11 new unit tests (see below) |

---

## Tests (`SchemaImportReservationTest.cpp`)

| # | Test name | What it verifies |
| --- | --- | --- |
| 1 | `ImportSingleClass_ReservationStorePopulated` | Importing a class creates `schema_reservation_ids` with `LastReservedId > 0` for `ec_Class` and `ec_Schema` |
| 2 | `ImportClassWithProperties` | Importing a TPH class with properties populates both `schema_reservation_ids` (`ec_Property`) and `schema_reservation_columns` (physical table row) |
| 3 | `AddPropertyByVersionBump` | Version-bumped import adding a property returns `OK`; both properties visible afterward |
| 4 | `AddClassByVersionBump` | Version-bumped import adding a class returns `OK`; both classes visible |
| 5 | `CrossBriefcaseColumnDeterminism` | Two briefcases importing the same property-adding upgrade produce identical `ec_Column.Id` and `ec_Column.Ordinal` |
| 6 | `CrossBriefcaseIdDeterminism` | Two briefcases importing the same schema produce identical `ec_Class.Id` for `ClassA` |
| 7 | `ReimportSameSchemaIsIdempotent` | Re-importing the same schema version does not advance `LastReservedId`; class id unchanged |
| 8 | `InitFromNonEmptyBase_ReservationStorePrePopulated` | `Init` on a briefcase with a pre-existing schema seeds `schema_reservation_ids` and `schema_reservation_columns` immediately — without any `ImportSchemas` call |
| 9 | `InitSeeding_ExistingClassIdPreservedAcrossBriefcases` | A class imported before `Init` is assigned the same id in a later briefcase; `ClassB` added after `Init` is identical across both briefcases |
| 10 | `OverrideContainerReInit_SeedsFromCurrentLocal` | Re-`Init` with `overrideContainer=true` seeds the new container from the current local baseline |
| 11 | `TwoSchemaImport_ThenVersionBumpAddsSharedColumn` | b1 imports two schemas (`TestSchema1` with a TPH `ShareColumns` hierarchy of `ABC→DEF` and `TestSchema2` adding `XYZ : DEF`); b2 is created after b1 pushes, then imports only `TestSchema1 v1.0.1` which adds `defProp2` to `DEF`. Asserts: (a) existing `defProp`/`xyzProp` column ids and ordinals match across b1/b2; (b) the new `defProp2` gets a fresh slot whose ordinal does not collide with the already-reserved `xyzProp` slot; (c) all properties share `ABC`'s single physical table; (d) sync-db reservation tables are populated |

---


iModelCore/ECDb/ECDb/SchemaSync.cpp
```diff
@@ -1405,5 +1472,1042 @@ DbResult SchemaSyncHelper::UpdateProfileVersion(DbR conn, SchemaSync::SyncDbUri
...
+    Utf8String sk = SchemaWriter::DeriveSchemaKey(schema);
+    for (ECN::IECInstancePtr ca : schema.GetCustomAttributes(false))
+        store.customAttribute.GetOrAllocate(SchemaWriter::DeriveCustomAttributeKey(sk, ca->GetClass()));
+}
+
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+BentleyStatus SchemaSync::LoadReservationStore(SyncDbUri const& syncDbUri, SchemaReservationStore& store) const {
+    Db syncDb;
```
rschili, line 2526: ReserveSchemaImport does not call SaveChanges() and this here opens a second read-only connection to the same file, I think this will not contain the uncommitted writes so this contains the state from before the import's reservations, no?

iModelCore/ECDb/ECDb/ECDbImpl.h
```diff
@@ -20,10 +20,41 @@ struct IdFactory final: NonCopyableClass {
...
+            //! Content-key-based id lookup.  In keyed mode returns the reserved id for @p key;
+            //! returns BeInt64Id(0) (invalid) if the key is absent — caller must propagate as an error.
+            //! An empty @p key is treated as "no key available" and falls back to NextId() even in keyed mode
+            //! (used by schema-upgrade / replace paths that do not have a content key readily available).
+            //! In all other modes falls back to NextId() so call-sites work identically in both modes.
+            BeInt64Id NextIdForKey(Utf8StringCR key) const {
+                if (m_keyedMode && m_keyToIdMap != nullptr && !key.empty()) {
+                    auto it = m_keyToIdMap->find(key);
+                    if (it == m_keyToIdMap->end())
+                        return BeInt64Id(0); // unreserved-key: BindId will bind NULL → INSERT will fail
```
rschili, line 41: This should probably fail on a miss. BindIf maps an invalid id to BindNull which auto assigns a value, I think...

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -1293,49 +1301,56 @@ SchemaImportResult MainSchemaManager::ImportSchemas(SchemaImportContext& ctx, bv
     auto& schemaSync = m_ecdb.Schemas().GetSchemaSync();
     const auto isSchemaSyncDisabled = schemaSync.IsSchemaSyncDisabled();
     auto resolvedSyncDbUri = syncDbUri.IsEmpty() ? schemaSync.GetDefaultSyncDbUri() : syncDbUri;
-    const auto localDbInfo = schemaSync.GetInfo();
 
-    if (!isSchemaSyncDisabled)
+    if(!isSchemaSyncDisabled && !resolvedSyncDbUri.IsEmpty()) {
```
rschili, line 1299: syncDbUri vs resolvedSyncDbUri. This guards on resolvedSyncDbUri but passes syncDbUri to ReserveSchemaImport.

soham-bentley: done

iModelCore/ECDb/ECDb/SchemaSync.cpp
soham-bentley: Yes done.

iModelCore/ECDb/ECDb/ECDbImpl.h
soham-bentley: Is it? I just checked till BindNull....and it seemed that it will bindNull and for primary keys that should fail when we will STEP on the INSERT Statement. Let me check with a test

soham-bentley:
# Schema Import Reservation — Open Problems

## Problem 1: Shared Column Reuse Across Briefcase Boundaries (Fix-Implementation in progress)

**Context:** When a derived class in a separate briefcase (one that was not present during the original schema import) owns properties that map to shared columns in a TPH hierarchy, the reservation system has no way to know which shared columns are already claimed by descendants living in other briefcases.

This is exercised by test `TwoSchemaImport_ThenVersionBumpAddsSharedColumn`: `b2` imports only `TestSchema1 v1.0.1` (which adds `defProp2` to `DEF`) without ever having seen `TestSchema2`, so the reservation layer cannot see that `xyzProp` of `XYZ : DEF` (defined in `TestSchema2`) already occupies a shared column in the same physical table.

**Current workaround:** The column allocator is conservative — it always advances the shared-column high-water mark rather than attempting to fill gaps. This prevents collisions but can leave many unused shared column slots when different briefcases import different subsets of a schema hierarchy.

**Proposed fix:** Add a new table to the sync-db that tracks the class hierarchy for every class that participates in a shared-column physical table:

```
schema_class_hierarchy
  ClassKey   TEXT  NOT NULL  -- "schemaName:className"
  Ancestors  BLOB  NOT NULL  -- FlatBuffer: repeated string "schemaName:className"
  Descendants BLOB NOT NULL  -- FlatBuffer: repeated string "schemaName:className"
  PRIMARY KEY (ClassKey)
```

- `Ancestors` encodes the full set of base classes (transitively) so that, given any class, the reservation layer can immediately identify every physical table that class shares with its roots.
- `Descendants` encodes every known derived class so that, when a new property is added to `DEF`, the system can skip ordinals already claimed by `XYZ`, `XYZ`'s subtypes, and any other descendant imported by any briefcase.
- Both blobs use FlatBuffers, keeping deserialization cost negligible (see Problem 2).
- The table is written during `ReserveSchemaImport` and seeded from the local `ec_ClassHasBaseClasses` graph during `Init`.

With this table, the column allocator can switch from a pure high-water-mark strategy to true gap-aware allocation: it reads the union of all ordinals claimed by ancestors **and** all known descendants before picking a slot for the new property, eliminating spurious gaps while still guaranteeing no collisions across briefcases.

---

## Problem 2: Excessive In-Memory State During Reservation (Fix-Implementation to be started)

**Context:** The current `ReserveSchemaImport` implementation loads a significant amount of data into memory — reservation id maps, column assignment maps, property path sets, and class hierarchy information — all kept alive for the duration of a single import call. For large schemas or schema upgrades that touch many classes and properties, this results in unnecessarily high peak memory usage.

**Proposed fix:** Exploit the fact that the underlying blob columns already use FlatBuffers. FlatBuffer objects support zero-copy, lazy field access: a field is only parsed when it is actually read, so holding a raw `ByteVector` in memory is nearly free compared to a fully-materialised C++ struct.

Concretely:
- Replace eagerly-built `std::unordered_map` / `std::vector` reservation caches with on-demand queries against the sync-db, backed by prepared `CachedStatement`s.
- For the column assignment blobs, keep only the raw FlatBuffer byte span in the reservation object; deserialize individual column entries only when a lookup is needed.
- Introduce a lightweight "reservation view" type that wraps a `ByteVector` and exposes a single `FindColumn(propertyKey)` method, deferring all parsing to call time.

This brings peak memory during a schema import from O(all reserved ids × all reserved columns) down to O(1) per lookup, with no change to the on-disk format or the correctness invariants.

---

## Problem 3: Reserving ids for mapping tables like `ec_PropertyMap` , `ec_PropertyPath`, etc

`WalkSchemaForMappingReservation` (reserve side) so `ec_Table` / `ec_PropertyPath` /
`ec_PropertyMap` ids are reproducible across offline briefcases. Consume side already calls
`NextIdForKey`. Key risks in this implementation are:

- **Under-reservation is fatal, not degrading.** A missing key makes `NextIdForKey` return an invalid
  id and the `INSERT` fails, so any *incomplete* walk breaks the whole import — not just determinism.
  The feature must stay behind its flag until coverage is complete or the import falls back to the
  schema-write lock.

- **Not yet covered (would fail hard if hit):**
  - System per-table property maps (`ECInstanceId` / `ECClassId`, one `ec_PropertyMap` per physical table).
  - Relationship-class property maps (source/target, link-table).
  - Joined-table placement (non-shared props are assumed to land in the primary table).
  - Auto-generated indexes (`ec_Index` / `ec_IndexColumn`) — **deliberately deferred**; the riskiest,
    since generated index names can reference generated shared-column names/ordinals.

- **Depends on the common-base invariant.** Divergent *import order* across briefcases is **not** a
  risk: reservation runs under the sync-db write-lock, and
  `GetOrAllocate` is idempotent by content key — so whoever reserves a key first fixes its id/placement
  and everyone else reads it back. The invariant instead guards another point:  the **one-time
  seed** reads local `MAX(Id)`/`MAX(Ordinal)`, so it must be taken against a base every briefcase
  shares (over-seeded ordinals would mis-place shared columns).

- **Reserve/consume key-derivation drift.** Reserve and consume derive the same string keys in two
  places. `ec_PropertyMap` embeds the physical table name, which reserve reads from the **column store**
  (single source of truth) rather than recomputing — so the mapping walk must run *after* the column
  walk. Any future change to a key format must land in both sites together.

- **Property overrides / renames** use `prop->GetClass()` + name for the `ec_PropertyPath` key;
  overridden properties that produce a distinct root property id are an untested edge case.

- **Over-reservation is safe.** Allocating an id that import never consumes is idempotent and harmless;
  only under-reservation is dangerous. When in doubt the walk reserves more, not less.

rschili:
We have been working on a new variant as outlined in the parent story. PR for the prototype is here: https://github.com/iTwin/itwinjs-core/pull/9604

rschili:
Closing in favor of https://github.com/iTwin/imodel-native/pull/1539

PR #1518, opened 2026-07-21
Support for Linux ARM

### LINKED ISSUE #1169: Can iTwin.js run on Linux with arm64 architecture?
I'm trying to run an iTwin.js service on an ARM64-based server, but the download of the @bentley/imodeljs-native dependency fails during installation.

naveedkhan8067:
imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/1141

Adding support for LinuxARM64 architecture.


khanaffan (approved):

iModelJsNodeAddon/api_package/ts/src/test/utils.ts
```diff
@@ -43,7 +43,7 @@ export function loadLocalBuildOfAddon(): any {
 
   const platformSubDirs = {
     win32: "Winx64",
-    linux: "LinuxX64",
+    linux: process.arch === "arm64" ? "LinuxARM64" : "LinuxX64",
```
aruniverse: couldnt we do the same template string darwin is using below?

aruniverse (approved):

PR #1520, opened 2026-07-21
Use raw sqlite changeset format for SQLiteDb changesets

khanaffan:
Switches the internal `SQLiteDb` create/apply changeset methods to use the raw sqlite changeset format (DML only).

itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9527

Closes iTwin/itwinjs-backlog#2233

### DIFF
--- a/iModelJsNodeAddon/IModelJsNative.cpp
+++ b/iModelJsNodeAddon/IModelJsNative.cpp
@@ -853,9 +853,12 @@ public:
     }
 
     //! Write out the changes captured since startChangeTracking() was called, to a changeset
-    //! file in the same on-disk format used for iModel changesets. Used only for testing.
+    //! file holding the raw sqlite changeset (the plain, uncompressed byte stream produced by
+    //! the sqlite session extension) - this is *not* the same format used for iModel changesets.
+    //! Raw sqlite changesets cannot represent DDL/schema changes, so this throws if any DDL was
+    //! captured since startChangeTracking() was called.
     void CreateChangeset(NapiInfoCR info) {
-        auto& db = GetWritableDb(info);
+        GetWritableDb(info);
         REQUIRE_ARGUMENT_STRING(0, pathname);
         if (m_changeTracker.IsNull())
             THROW_JS_IMODEL_NATIVE_EXCEPTION(info.Env(), "change tracking was not started", IModelJsNativeErrorKey::BadArg);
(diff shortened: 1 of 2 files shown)


iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -853,9 +853,12 @@ struct SQLiteDb : Napi::ObjectWrap<SQLiteDb>, SQLiteOps<Db> {
     }
 
     //! Write out the changes captured since startChangeTracking() was called, to a changeset
-    //! file in the same on-disk format used for iModel changesets. Used only for testing.
+    //! file holding the raw sqlite changeset (the plain, uncompressed byte stream produced by
+    //! the sqlite session extension) - this is *not* the same format used for iModel changesets.
+    //! Raw sqlite changesets cannot represent DDL/schema changes, so this throws if any DDL was
+    //! captured since startChangeTracking() was called.
```
aruniverse, line 859: Optional/non-blocking doc-consistency nit: this reworded comment drops the previously-present "Used only for testing." caveat, which the paired `NativeLibrary.ts` ambient declaration for the same method still keeps (`"...startChangeTracking() was called. Used only for testing."`). Since these comments are hand-synced (no codegen), consider re-adding it here too, e.g. append "Used only for testing." after this line.

aruniverse (commented):
Adversarial review (automated, fresh-context reviewer focused on correctness/compatibility, cross-checked against the paired JS PR iTwin/itwinjs-core#9527 and the linked issue iTwin/itwinjs-backlog#2233).

**Correctness — verified, no blocking issues found:**
- `CreateChangeset`'s DDL-rejection check (`HasDdlChanges()`) runs before any changeset bytes are written, and the throw is a genuine C++ exception, so `SQLiteDbScopeGuard endTracking` still fires via stack unwind (`EndTracking()` runs on the throw path — no leaked/inconsistent tracker state).
- `BeSQLite::ChangeSet`'s default `_OnConflict` is `BeAssert(false)`, but `ApplyChangesArgs::ConflictCallback` returns `Abort` directly when `SetAbortOnAnyConflict(true)` is set (as used here) and never reaches `_OnConflict` — confirmed by reading `ChangeSet.cpp`, so this path is safe.
- Old vs. new changeset file formats are intentionally, permanently cross-incompatible (old = LZMA-compressed custom header + optional DDL via `ChangesetFileWriter`; new = raw uncompressed session-extension byte stream via `ChangeSet::Read`/`Write`). Grepped both repos for other callers of `SQLiteDb`'s `createChangeset`/`applyChangeset`/`startChangeTracking`/`executeDdl` outside the test file — found none; the only other `applyChangeset` caller (`BriefcaseManager.ts` → `DgnDb::ApplyChangeset`) is a distinct, unrelated native method. The "internal/test-only" scope this PR relies on is accurate, so the format break is low-risk.

**Note (optional, non-blocking):** the C++ doc comment on `CreateChangeset` (lines 855-859) drops the "Used only for testing." caveat that the paired `NativeLibrary.ts` declaration still retains verbatim. Since these two are hand-kept-in-sync with no generation tooling, a reader of only the C++ source wouldn't learn this method is testing-only. Minor, but easy to fix while this comment is already being touched.

**Cross-repo merge-sequencing note:** itwinjs-core's `core/backend/package.json` still exact-pins the native addon to a version predating this PR, and itwinjs-core#9527's new tests assume this PR's behavior is already published. Worth coordinating publish/bump order before merging either side (flagged in detail on #9527).

aruniverse (approved):

soham-bentley (approved):

PR #1525, opened 2026-07-28
Switch pugixml build to vcpkg

### LINKED ISSUE #1503: Use vcpkg to build pugixml
Build pugixml with vcpkg instead of bmake.

tcobbs-bentley:
## TL;DR

* Use vcpkg to build pugixml instead of bmake.
* Update pugixml from version 1.12 to 1.15.

---

# Build pugixml with vcpkg (and bump 1.12 → 1.15)

Migrates `iModelCore/libsrc/pugixml` from a checked-in, file-by-file compiled source tree to a
vcpkg-driven build, following the pattern already established for `compress`, `png`, and
`openssl`. The upstream version moves from **1.12** to **1.15** (newest release at our pinned
registry baseline `81de6771…`).

## What makes pugixml different from the previously migrated libraries

Unlike zlib/png/openssl, pugixml is not pure upstream code:

1. **We still compile our own wrapper.** `src/BePugiXml.cpp` (plus `BePugiXmlHelper.h`) is
   Bentley code that adapts `pugi::xml_node` to a `BeXml`-compatible API. vcpkg can't build it,
   so `pugixml.mke` cannot become a thin re-delivery — it compiles `BePugiXml.cpp` and
   **combines** the result with vcpkg's upstream pugixml archive.
2. **The `pugi::` surface is part of the public contract.** Shipped ecobjects headers
   (`ECSchema.h`, `ECInstance.h`, `ECUnit.h`) include `<pugixml/src/pugixml.hpp>` and use
   `pugi::` types directly, so `iTwinPugixml` must keep exporting the whole pugixml API. That
   export is driven by `__declspec(dllexport)` annotations in `pugiconfig.hpp`, not by a `.def`.

## Approach

vcpkg builds **upstream pugixml as a static archive on every platform** (mirroring the OpenSSL
decision), and `pugixml.mke` produces the `iTwinPugixml` deliverable from that archive plus our
`BePugiXml.o`:

| Build | Deliverable | How |
|-------|-------------|-----|
| Static (all platforms; thin-archive input to `imodeljs.node`) | one static archive | merge `BePugiXml.o` + vcpkg `libpugixml.a`: `lib -OUT` (Windows), `libtool -static` (macOS/iOS), `merge_static_libs.sh` (Linux), `ar qcL` (Android) |
| Dynamic | shared lib + import lib | standard `dlmlink` machinery (`linkLibrary.mki`) over `BePugiXml.o` + vcpkg `libpugixml.a` |

**Deliverable names and include paths are unchanged**: `iTwinPugixml` and the VendorAPI headers
under `VendorAPI/pugixml/src/` are all byte-for-byte identical in name and location, so no consumer changes are required.

`pugixml-license.md` is now `pugixml-license.txt` since it's just a text file, but delivery of that is wired up to happen automatically.

### New vcpkg wiring (`iModelCore/libsrc/pugixml/`)
- `vcpkg.json` / `vcpkg-configuration.json` — manifest pinned to pugixml 1.15 against the shared
  registry baseline, with an overlay-ports directory.
- `overlay-ports/pugixml/` — overlay port that:
  - injects the Bentley dllexport/dllimport block into `pugiconfig.hpp` via
    `vcpkg_replace_string` (no `.patch` file, so it survives upstream byte changes), and
  - passes `-DPUGIXML_BUILD_DEFINES=__PUGIXML_BUILD__` so pugixml 1.15's **static** CMake target
    emits the dllexport side of that annotation (upstream only annotates the shared target).
- `triplets/` — the standard 8 triplets (arm64/x64 android, arm64 ios, arm64/x64 linux,
  arm64 osx, x64 windows-static and windows-static-clang), matching `png`'s set. This adds
  **Linux ARM64** support alongside the other platforms.

### Build integration
- `vcpkg_install_pugixml.mke` (new) and `vcpkg.PartFile.xml` — inserts `vcpkg_install_pugixml`
  into the sequential install chain: `compress → png → pugixml → openssl → crashpad`, and adds it
  to `vcpkg_install_all`.
- `pugixml.PartFile.xml` — `Prewire` and `PugiXml` now SubPart `vcpkg_install_pugixml`
  (`LibType="Static"`); the existing bindings are unchanged (they already tolerate both linkages).
- `pugixml.mke` — no longer compiles `src/pugixml.cpp`; resolves the vcpkg triplet/install root
  (dynamic builds read from the static install root, since the chain only runs static) and
  performs the merge/link described above.
- `prewire.mke` — stages `pugixml.hpp`, the patched `pugiconfig.hpp`, and the license from the
  vcpkg install tree instead of `src/`. The Bentley wrapper headers still come from `src/`.
  Destination paths are unchanged.

### Shared infrastructure
- `merge_static_libs.sh` — now accepts a raw `.o`/`.obj` input in addition to `.a` archives (one
  of pugixml's merge inputs is our own `BePugiXml.o`), and collects objects with `find`/`xargs`
  instead of a fixed glob.
- `windows-clang-cl.toolchain.cmake` — appends `/D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH`.
  The MSVC STL in VS2022 17.14+ (MSVC 14.44) hard-errors with `STL1000: Unexpected compiler
  version, expected Clang 19.0.0 or newer`, and CI still ships Clang 18. This matches what the
  rest of the native Windows clang build (`winntclangmdl.mki`) and crashpad's portfile already
  do. Appended after `windows.cmake` so it survives that file's flag initialization, and guarded
  so repeated inclusions (compiler detection, `try_compile`) don't duplicate the define.
- `README.md` — pugixml row updated to 1.15 / vcpkg = Yes.
- `VCPKG.md` — install chain and consumer `.mke` lists updated to include pugixml.

### Documentation
- `pugixml/VCPKG_MIGRATION.md` — the step-by-step migration plan and rationale, including the
  deviations from the original plan as implemented.

## Follow-up (not in this PR)

- Deleting the now-vendored upstream sources (`src/pugixml.cpp`, `src/pugixml.hpp`,
  `src/pugiconfig.hpp`, `LICENSE.md`) is intentionally left to a separate commit so the diff of
  the wiring stays reviewable.

## Testing

No source or API changes for consumers; `iTwinPugixml`, its exported symbols, and the delivered
header paths are unchanged. Verification is per-platform CI builds (Windows MSVC + clang, macOS,
Linux x64 + ARM64, iOS, Android) in both static and dynamic configurations.


iModelCore/libsrc/merge_static_libs.sh
```diff
@@ -16,19 +18,29 @@ OUTPUT="$1"
...
+            LIBNAME=$(basename "$LIB" .a)
+            mkdir -p "$TMPDIR/$LIBNAME"
+            (cd "$TMPDIR/$LIBNAME" && ar x "$LIB")
+            ;;
+    esac
 done
 
 rm -f "$OUTPUT"
-ar rcs "$OUTPUT" "$TMPDIR"/*/*.o
+find "$TMPDIR" -type f -name '*.o' -print0 | xargs -0 ar rcs "$OUTPUT"
```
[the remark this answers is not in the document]
tcobbs-bentley: The script was instead updated to remove .obj support, since those don't exist on the platforms where the script is used.

iModelCore/libsrc/merge_static_libs.sh
```diff
@@ -16,19 +19,29 @@ OUTPUT="$1"
...
 
 for LIB in "$@"; do
-    # Extract each archive into a unique subdirectory to avoid name collisions
-    LIBNAME=$(basename "$LIB" .a)
-    mkdir -p "$TMPDIR/$LIBNAME"
-    (cd "$TMPDIR/$LIBNAME" && ar x "$LIB")
+    case "$LIB" in
+        *.o)
+            # Raw object file: stage it directly (into a unique subdirectory to avoid collisions).
+            OBJDIR="$TMPDIR/obj_$(basename "$LIB")"
```
aruniverse: The staging directory is not unique when two raw inputs share a basename. For example, `a/same.o` and `b/same.o` both map to `obj_same.o/same.o`, so the second `cp` silently overwrites the first and the merged archive loses `a/same.o` (I reproduced this against the PR head; `nm` showed only the second object's symbol). Please use an input index or another collision-free directory name for each argument. The archive branch has the same basename-derived assumption, so indexing every input would fix both paths.

tcobbs-bentley: Fixed.

iModelCore/libsrc/merge_static_libs.sh
```diff
@@ -16,19 +19,42 @@ OUTPUT="$1"
...
+            (cd "$INPUTDIR" && ar x "$ARCHIVE")
+            ;;
+    esac
 done
 
 rm -f "$OUTPUT"
-ar rcs "$OUTPUT" "$TMPDIR"/*/*.o
+# sort -z keeps member order deterministic: find returns entries in readdir order, which varies
+# between machines and filesystems and would otherwise make the archive byte-differ run to run.
+find "$STAGEDIR" -type f -name '*.o' -print0 | sort -z | xargs -0 ar rcs "$OUTPUT"
```
aruniverse: The indexed staging fixes the direct basename overwrite, but this final pipeline still has two ways to silently produce an incomplete archive:

1. `set -e` does not propagate failures from `find` or `sort` without `pipefail`. I injected a `sort` that emitted one pathname and exited 42; `xargs`/`ar` succeeded, the script returned 0, and the archive contained only one of two objects.
2. `xargs` may split a large input at `ARG_MAX`. Each later `ar rcs` invocation replaces an existing same-named member. Simulating a legitimate split with one object per invocation reduced two different `same.o` inputs to the second object only.

Please enable `pipefail` and ensure archive construction either happens in one invocation (failing rather than splitting) or uses a manifest/append strategy that preserves duplicate member names across chunks.

iModelCore/libsrc/merge_static_libs.sh
aruniverse: Confirmed the original staging collision is fixed in `62649f7743`: every input now gets its own indexed directory. I reran the original `a/same.o` / `b/same.o` reproduction and both symbols are retained; two same-named archives also retain both members, including when each archive contains `same.o`. Relative/absolute paths and spaces passed too.

I found a separate remaining edge case in the final `find | sort | xargs | ar` pipeline and left the reproduction here: https://github.com/iTwin/imodel-native/pull/1525#discussion_r3675904892

iModelCore/libsrc/merge_static_libs.sh
tcobbs-bentley: Fixed.

aruniverse: Confirmed fixed at `389b7d9a01`. With an injected `sort` that emits one entry and exits 42, the script now returns 42 and creates no archive, so the partial-pipeline failure is no longer masked. The `xargs` path is also gone: the NUL-delimited Bash array feeds one final `ar` invocation, and duplicate `same.o` inputs retain both members/symbols.

I also re-tested raw/raw, archive/archive, and mixed raw/archive duplicate basenames, relative/absolute paths with spaces, no-object handling, and cleanup after failure. All passed; no new actionable issue found in this change.

aruniverse (approved):

DanRod1999 (approved):

PR #1526, opened 2026-07-28
Ship iTwinDaemon in iModelNativeUPack

mathieu-lemuzic:
### What this does

Adds the `ITwinDaemon-exe` subpart to the `iModelNativeUPack` product so that `iTwinDaemon` is delivered alongside the native SDK in `native/bin`, and wires up the two `ProductDirectory` entries required to make that delivery resolve cleanly.

### Why this is needed

The imodel tiler (that currently consume `iModelNativeUPack`) requires a local `.bim` file to produce tiles. We want to experiment with consuming the iModel through a CloudSQLite SAS URL instead, so that only the blocks actually touched get downloaded — a sizeable footprint reduction, especially for incremental exports where the changed subset is small relative to the whole file.

The code paths that read from the `.bim` run across **multiple processes** (Parasolid's multi-threading limitations force process-level parallelism rather than threads). A CloudSQLite cache can only be shared safely across processes through the iTwin daemon, which is why it has to ship alongside the iModel native libraries rather than being an optional side-install.

Use V2 checkpoint SAS URL instead of full .bim file:  https://github.com/iTwin/itwin-tiling-backlog/issues/44

### Changes

All changes are confined to a single file: `iModelCore/iModelCore.PartFile.xml`.

1. **`<SubPart PartName="ITwinDaemon-exe" .../>` added to `iModelNativeUPack`.**

2. **Declared `StaticAssemblies` (static) → `native/bin`.**
   `ITwinDaemon` is always linked statically, so its executable binding uses `ProductDirectoryName="StaticAssemblies"`. That name was not declared in `iModelCoreNugetDirs`, so BentleyBuild would have silently skipped the executable with only a `Product specification is incomplete` warning. This entry is what actually puts `iTwinDaemon` in the package.

3. **`Libs` / `LibType="Static"` / `Platforms="x64"` set to `Deliver="false"`.**
   This is the fix for the CI failure:
   ```
   ####Cannot link: '...\Product\iModelNativeUPack\native\lib\iTwinOpenSSL.lib'
   to: '...\build\iTwinOpenSSL\iTwinOpenSSL.lib',
   because it is already linked to: '...\static\build\iTwinOpenSSL\iTwinOpenSSL.lib'
   ```
   On x64 the product is the *dynamic* SDK, but the daemon drags in a parallel *static* dependency chain (openssl, curl, proxyres, …). Those static `.lib` files have the same filenames as the dynamic import libraries already delivered to `native/lib`, so the two symlinks collide. (`iTwinCurl.lib` would have been the next failure.)

   These static libraries are link-time inputs for the daemon executable only — nothing that consumes this package links against them — so the correct resolution is to not deliver them rather than relocate them.

### Why the impact is minimal

- **Single file touched.** No `.mke`, no source, no build-flow changes.
- **The directory list is not shared outside this file.** `iModelCoreNugetDirs` has exactly three references, all inside `iModelCore.PartFile.xml` (`iModelNativeNuget`, `iModelNativeUPack`, and the list itself). There are zero references in `imodel-native-internal` or anywhere else.
- **No effect on the Node addon.** `iModelJsNodeAddon` products use their own `iModelJsNodeAddonDeliveryList`, which is untouched. That list already declares `StaticAssemblies` and does *not* declare `Libs` at all — and `iModelJsNative-Dynamic` (x64) already ships `ITwinDaemon-exe` today. This PR simply brings `iModelNativeUPack` in line with that existing, proven configuration.
- **No effect on non-Windows platforms.** The `Deliver="false"` entry is scoped with `Platforms="x64"`. Linux, macOS, iOS and Android build this product entirely static, where the static libraries *are* the real deliverable; they keep the default entry and their package contents are byte-for-byte unchanged.
- **No dynamic artifacts removed.** The x64 `native/bin`, `native/include`, `native/assets` and `native/notices` folders are unchanged, and every dynamic import library still lands in `native/lib` exactly as before.
- **Static libs were never consumable from the x64 package.** The only static-libtype `Libs` binding previously reaching `native/lib` on x64 was `proxyres.lib`, which arrived there as an accident of the binding model — it cannot be linked against the dynamic SDK. `native/lib/static` (`BeStaticLibDir`: BeZlib, BePng, lzma) uses a distinct binding name and is unaffected.

### Net result on x64

| Path | Before | After |
| --- | --- | --- |
| `native/bin` | dynamic binaries | dynamic binaries **+ `iTwinDaemon`** |
| `native/lib` | import libs + stray `proxyres.lib` | import libs only |
| `native/lib/static` | BeZlib, BePng, lzma | unchanged |
| `native/include`, `native/assets`, `native/notices` | — | unchanged |

### Testing

Validated by CI package build. Cannot be reproduced locally (requires an initialized Bentley build environment).


DanRod1999:
What's this needed for?

mathieu-lemuzic:
> What's this needed for?

I updated the description

iModelCore/iModelCore.PartFile.xml
```diff
@@ -55,10 +56,20 @@
...
         <ProductDirectory Name="Libs"                                       RelativeTo="Root" Path="lib"/>
         <ProductDirectory Name="Libs"                   LibType="Static"    RelativeTo="Root" Path="lib"/>
+        <!--
+            On x64 this product is built Dynamic, but ITwinDaemon-exe drags in a parallel *static*
+            dependency chain (openssl, curl, rustls, proxyres, ...). Those static libraries have the
+            same file names as their dynamic import libraries, so they cannot share "lib". Give them
+            their own sub-directory (same convention as BeStaticLibDir below). On the platforms that
+            build this product entirely Static there is only one chain, so "lib" stays correct there.
+        -->
+        <ProductDirectory Name="Libs"                   LibType="Static"    Platforms="x64" RelativeTo="Libs" Path="static"/>
```
wgoehrig: This does not seem right to me - we should not be delivering these static libs at all for x64, even if we're now including the daemon...

mathieu-lemuzic: my bad, this version was stale and did not reflect the PR description, it's updated now to exclude static libs from the delivery on x64

wgoehrig (approved):

PR #1528, opened 2026-07-29
Fix silent schema import failure when a force-remapped struct property splits across tables

rschili:
Fixes a connector `dynamicSchemaImportError` where a dynamic schema upgrade failed without any error message.

## Bug description

When a schema upgrade adds local property overrides to a class (EC2->EC3 multi-inheritance flattening does this), the remap manager frees the class's old columns. If columns are freed in both the primary table and its overflow table, all freed columns are blocked from reuse to prevent cross-table circular remaps (@RohitPtnkr1996 this was a change you recently made afaik). A TPH sibling class whose properties sit on those blocked columns is then force-remapped.

During that remap, `EvaluateIfPropertyGoesToOverflow` can decide a struct property must move to the overflow table as a whole. The reuse path in `ClassMapColumnFactory::Allocate` ignored that decision: any member whose old column was *not* freed kept its registration in the primary table (found through the derived classes' cloned maps in the column resolution scope), while the remaining members were allocated in overflow. The struct then spanned two tables, which `DbMappingManager::Classes::ProcessProperty` treats as a fatal error - previously without reporting anything.

## Changes

- `ClassMapColumnFactory::Allocate`: only reuse a registered column when it does not contradict the overflow decision for the current property.
- `ProcessProperty`: report an error (`ECDb_0742`) naming class and property when a property maps to columns spanning multiple tables, instead of failing silently.
- `ClassMap::Update` and `MainSchemaManager::MapClass`: log the failing property/class on remap failure.
- Fixed two no-op asserts in `MoveProperty` (`BeAssert("string")` is always true) and added error logs.
- Renamed the `DoMapSchemas` perf scope to "Map classes" - it previously shared the name "Persist mappings" with `SaveDbSchema`, which sent the initial investigation of this bug to the wrong import phase.

## Failure diagnostics

Mapping-phase failures now emit a diagnostics issue (`ECDb_0743`) containing a remapping summary (modified classes, added properties/base classes, cleaned mappings, blocked columns) plus the last 100 column mapping decisions. Decisions are only recorded when the import freed columns, so regular imports pay no cost and the buffer is capped for large dynamic imports. This surfaces in connector job logs, where this class of failure previously only showed up as a generic error.

## Tests

- `ForcedStructRemapSplitsAcrossTables`: minimal repro of the customer scenario (verified equivalent against the customer iModel); asserts the import succeeds and the struct ends up in a single table.
- `MappingFailureDiagnosticsReport`: forces a mapping failure after remapping work and asserts the diagnostics report is issued with summary and decision trace.


iModelCore/ECDb/ECDb/ClassMapColumnFactory.cpp
```diff
@@ -510,6 +510,8 @@ void ClassMapColumnFactory::EvaluateIfPropertyGoesToOverflow(uint32_t columnsReq
     uint32_t requiredRemainingColumns = columnsRequired - sharedColumnThatCanBeCreated;
     if (requiredRemainingColumns > nSharedColumns)
         { //no need to check, we know there won't be enough columns
+        if (ctx.RemapManager().HasFreedColumns())
+            ctx.AddMappingDecision(Utf8PrintfString("%s: property requiring %u columns goes to the overflow table (more columns required than exist in primary/joined table)", m_classMap.GetClass().GetFullName(), columnsRequired));
         m_putCurrentPropertyToOverflow = true;
```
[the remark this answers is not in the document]
rschili: fixed.

rschili:
The fix introduced a regression, I'm investigating. Apparently it blocks column reuse too eagerly.

iModelCore/ECDb/ECDb/ClassMapColumnFactory.cpp
```diff
@@ -568,17 +572,40 @@ DbColumn* ClassMapColumnFactory::HandleOverflowColumn(DbColumn* column) const
...
         {
         if (IsCompatible(*column, type, param))
             {
             if (!ctx.RemapManager().IsColumnFreed(*column))
-                return HandleOverflowColumn(column);
+                {
+                // If the current property as a whole was decided to go to the overflow table, do
+                // not reuse a registered column from the primary/joined table. Doing so would
+                // split a compound property (e.g. a struct) across two tables, which is invalid.
+                if (!m_putCurrentPropertyToOverflow || column->GetTable().GetType() == DbTable::Type::Overflow)
```
RohitPtnkr1996: I noticed the CI builds failing, there seems to be a regression.
I think this is a bit too broad of a check and it's affecting single column properties as well.
Maybe we gate this only for compound properties ?

rschili: Yeah, I just pushed a fix, I think it will be fine now.

khanaffan (approved):

Issue #1530, opened 2026-07-30
We want the full Vertical CRS support we have in desktop products


AlainRobertAtBentley:
Following support by PP and other desktop products we want iTwin to support the same complete functionality for Vertical CRS

PR #1531, opened 2026-07-30
New Vertical CRS subsystem

AlainRobertAtBentley:
This pull request implements the support on the native side for the more complete Vertical CRS as implemented in other desktop products.
It does not implement the representation in itwinjs-core just yet.

The new Vertical CRS changes the former Vertical CRS support to the limited 5 keyword specific current system (ELLIPSOID, NGVD29, NAVD88, ...) to a complete system where the Vertical CRS is based on a definition that includes the type (ELLIPSOID/GEOID) and provides ways to indicate what Geoid model is necessary to make conversion between various Vertical CRS.

A complete documentation will be added to itwinjs-core once the implementation on the non-native side is made.

Ref: https://github.com/iTwin/imodel-native/issues/1530

The pull request also implement small bug fixes as well as the restructuration of the WKT parser in preparation for the upcoming WKT2 support.


iModelCore/libsrc/csmap/Source/CS_gtxFile.c
```diff
@@ -844,7 +845,10 @@ int CSopnBinaryGtxFile (struct cs_GtxFile_ *__This,long32_t bufrSize)
     __This->density [LNG] = deltaLng;
     __This->density [LAT] = deltaLat;
 
-    __This->headerCount = sizeof (double) * 6;
+    if (__This->fileHasDoubleRowsColumns)
+        __This->headerCount = sizeof (double) * 6;
+    else
+        __This->headerCount = sizeof (double) * 4 + sizeof(long) * 2;
```
aruniverse: **Blocker (LP64):** the row/column fields above are read as `long32_t`, but this offset uses `sizeof(long)`. On macOS/Linux/iOS/Android, `long` is 8 bytes, so a NOAA GTX header is treated as 48 bytes instead of 40; grid reads start two floats late and can run past EOF. Please use `sizeof(long32_t) * 2` (or `sizeof(nRows) + sizeof(nColumns)`) and add a known-cell interpolation test on an LP64 job.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -4263,77 +4436,127 @@ GeoCoordParseStatus SetVerticalCS (BaseGCSR baseGCS, Utf8StringR wkt) const
...
+        
+        if (std::string::npos == endPosRight)
+            return GeoCoordParse_ParseError;
+        
+        if ((std::string::npos == endPosLeft) || (endPosRight < endPosLeft))
             {
-            // Angular units are not degree. We support that only for longitude/latitude based GCS (which is the case)
-            // This only means adjusting the prime meridian of the GCS.
-            baseGCS.SetOriginLongitude(baseGCS.GetOriginLongitude() * conversionToDegree);
+            wkt = wkt.substr(endPosRight);
```
aruniverse: **Blocker:** this leaves the cursor on the closing `]` instead of consuming it. The caller then mistakes the `METADATA` terminator for the enclosing CRS terminator and can skip required clauses that follow. The nested-section branch has the same issue with `[` and immediately hits the no-progress check. Please advance past each delimiter (`+ 1`) while maintaining depth, and cover nested metadata followed by projection/parameter clauses.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -4585,524 +4672,624 @@ GeoCoordParseStatus GetGeographicToCoordSys (Utf8StringR wkt, Utf8StringR geogra
             // East and North axis is the default and need not be set.
             if (horizontalAxis != AxisDirection::EAST || verticalAxis != AxisDirection::NORTH)
                 {
-                // We have a special quadrant ...
                 if (horizontalAxis == AxisDirection::WEST && verticalAxis == AxisDirection::NORTH)
-                    coordinateSystem.SetQuadrant(2);
+                    baseGCS.SetEPSGQuadrant(2);
```
aruniverse, line 5096: **Blocker:** this rewrite changed the old parser's `SetQuadrant(...)` calls to `SetEPSGQuadrant(...)`. The latter only updates `csdef.epsg_qd`; coordinate ordering/orientation still uses `csdef.quad`. Non-default WKT axes (for example Northing/Easting) therefore parse with the default operational quadrant and silently change coordinates/round-tripping. Please restore `SetQuadrant` here (and in the analogous block around 5206); set both fields if EPSG metadata is also needed.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -8782,7 +8969,7 @@ class VerticalGeoidSeparationGridTransform : public VerticalTransform
         return GEOCOORDERR_InvalidTransform;
     }
 
-    StatusInt InitializeTransform() override
+    StatusInt InitializeTransform(BaseGCSCR targetGCS) override
     {
         if (nullptr != m_csGeoidHeight)
```
aruniverse, line 9380: **High:** this cached state is target-CRS-specific, but the transform objects come directly from the dictionary and are shared. After one conversion initializes `m_datumConverter`, another conversion using the same vertical transform with a different horizontal CRS returns here and reuses the first converter. `VerticalDatum::ReleaseTransforms()` also releases that shared state for other users. Please separate immutable dictionary definitions from per-converter runtime state (or clone transforms per `VerticalDatum`) and test two target CRSs sequentially/concurrently.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -8792,6 +8979,26 @@ class VerticalGeoidSeparationGridTransform : public VerticalTransform
...
             return ERROR;
 
+        // This datum converter is provided to change latitude/longitude from a datum which isnon coincident
+        // to WGS84 to WGS84 to access the grid files that are based on WGS84 coincident datums.
+        BaseGCSPtr ll84 = BaseGCS::CreateGCS("LL84");
+
+        if (m_direction == GridFileDirection::DIRECTION_INVERSE)
+            m_datumConverter = DatumConverter::Create(targetGCS, *ll84);
+        else
+            m_datumConverter = DatumConverter::Create(*ll84, targetGCS);
```
aruniverse: **High:** the coordinates passed to `GetElevation` are already in `targetGCS`'s horizontal datum (`inLatLongDestDatum`), and the grid is WGS84-based. This direct branch constructs the converter in the opposite direction (`LL84 -> targetGCS`), so a non-WGS84 target samples the grid at the wrong location. Vertical direct/inverse should affect only the offset sign; the horizontal converter should consistently be `DatumConverter::Create(targetGCS, *ll84)`.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -8841,21 +9048,31 @@ class VerticalGeoidSeparationGridTransform : public VerticalTransform
...
 
         if ((nullptr == m_csGeoidHeight) || (0 == m_gridFiles.size()))
             return GEOCOORDERR_GeoCoordNotInitialized;
 
         elevationType = ElevationType::Offset;
 
-        const double pointLL[2] = { ptIn.longitude, ptIn.latitude };
+        GeoPoint effectivePointIn = ptIn;
+        if (m_datumConverter != nullptr)
+            m_datumConverter->ConvertLatLong3D (effectivePointIn, ptIn);
```
aruniverse: **Medium:** the datum-conversion status is discarded. If the horizontal conversion is out of range or fails because a path/grid is unavailable, the code still queries the geoid grid and may return success from unchanged/invalid coordinates. Please capture and propagate the `ReprojectStatus` before calling `CScalcGeoidHeight`.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -18448,12 +18949,50 @@ VerticalDatumPtr BaseGCS::CreateVerticalDatumFromName(Utf8CP verticalDatumName,
 /*---------------------------------------------------------------------------------**//**
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-void BaseGCS::SetVerticalDatum(VerticalDatumPtr verticalDatum)
+StatusInt BaseGCS::SetVerticalDatum(VerticalDatumPtr verticalDatum)
 {
     if (m_verticalDatum.IsValid())
         m_verticalDatum = nullptr;
```
aruniverse: **Medium:** the existing valid datum is cleared before the replacement's extent is validated. If the new datum does not overlap and the method returns `GEOCOORDERR_CantSetVerticalDatum`, the object has nevertheless lost its previous datum. Please validate first and assign only on success; clear immediately only for an explicit null candidate.

aruniverse (commented):
Adversarial review found several concrete correctness and compatibility issues. Details and suggested fixes are inline.

iModelCore/GeoCoord/Tests/NonPublished/SpecificVerticalTransformTests.cpp
```diff
@@ -158,22 +158,22 @@ static bvector<verticalConversionTest> s_listOfVerticalConversionTests =
     {"LL84", "LL84", "OSGM02", "WGS84", 1.684774, 52.330884, 0.234, 44.74, 0.02},
     {"LL84", "LL84", "OSGM02", "OSGM15", 1.684774, 52.330884, 0.234, 0.251, 0.02},
 
-    {"LL84", "LL84", "WGS84", "NAP2018", 5.387203657, 52.155172897, 43.2551, 0.000, 0.02},
+    //{"LL84", "LL84", "WGS84", "NAP2018", 5.387203657, 52.155172897, 43.2551, 0.000, 0.02},
```
aruniverse, line 161: **High validation gap:** this PR comments out ten existing conversion cases, including the NAD83, ETRS89/SWEREF99, NAP2018, and Latvia paths most likely to expose the new datum-direction and target-specific initialization bugs. Please restore these cases after fixing the runtime path rather than merging with the coverage disabled; also add a sequential case that reuses one vertical transform with two horizontal target datums.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
```diff
@@ -6249,11 +6273,13 @@ void SetTarget(const Utf8String& target);
...
 * Must be defined by all subclasses of VerticalTransform, initialize this VerticalTransform
-* in preparation for a following GetElevation() call.
+* in preparation for a following GetElevation() call. 
+* The targetGCS is used to prepare to transform the provided latitude/longitude from the
+* given coordinate to the geodetic datum of the transform (usually WGS84 or coincident)
 * @return   SUCCESS, ERROR or GEOCOORDERR_***, will be defined by the subclass.
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-virtual StatusInt InitializeTransform() = 0;
+virtual StatusInt InitializeTransform(BaseGCSCR targetGCS) = 0;
```
aruniverse: **Compatibility:** this is a public pure-virtual signature change. Any out-of-tree `VerticalTransform` subclass implementing `InitializeTransform()` stops compiling, and previously built subclasses no longer satisfy the virtual contract. If this hierarchy is intended to remain public, please retain the old entry point and add a separately named/context-aware method with a compatibility default (or explicitly transition/version the API).

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: The change is intentional. While very few entries in the dictionary have a Quadrant other than 1 most are erroneous and effectively meant an EPSGQuadrant other than 1. Yes this changes the behavior in theory, it effectively corrects an invalid interpretation of the axes in the WKT.

iModelCore/GeoCoord/Tests/NonPublished/SpecificVerticalTransformTests.cpp
AlainRobertAtBentley: The deactivation is intentional. With the correction of the defect in the GTX file interpretation above tests may become invalid (it is effectively invalid for the NAP2018). The tests also requires that the CRS definitions (from the dedicated cloud storage) be also updated which it will be later in the month of August. Once the dataset is updated these tests will be corrected, augmented then reactivated.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -3686,207 +3718,398 @@ class SRSWKTParser: public SRSGeneralParser
...
+*
 *   @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-SRSWKTParser()
+bool StartsWithKeyword(const Utf8String& wkt, const Utf8String& keyword) const
     {
+    Utf8String temp(wkt);
+    temp.Trim();
+
+    return ((temp.length() >= keyword.length()) && (0 == BeStringUtilities::Strnicmp(temp.substr(0, keyword.length()).c_str(), keyword.c_str(), keyword.size())));
```
tcobbs-bentley: Minor nit: even though they return the same value, using both `keyword.size()` and `keyword.length()` in the same line of code seems misguided.

tcobbs-bentley: I'm not sure how big `wkt` is expected to be, but copying `wkt`, then trimming whitespace off both ends, then checking the first `n` characters (where `n` is the length of `keyword`) against `keyword` seems very inefficient.

```suggestion
    size_t pos = wkt.find_first_not_of(" \f\n\r\t\v");
    if (pos != Utf8String::npos)
        {
        return ((wkt.length() - pos >= keyword.length()) && (0 == BeStringUtilities::Strnicmp(wkt.substr(pos, keyword.length()).c_str(), keyword.c_str(), keyword.length())));
        }
    return false;
```

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -3686,207 +3718,398 @@ class SRSWKTParser: public SRSGeneralParser
...
+    wkt.Trim();
+    
+    return true;
+    }    
+    
 /*---------------------------------------------------------------------------------**//**
+*   @description This method extracts from the provided stream the double
+*   The first non white character must be the number to extract.
+*
+*   @param wkt IN The WKT portion that contains the number to extract.
```
tcobbs-bentley: Since this is a ref and the value referred to is being modified by the function, it seems like it is IN/OUT.

tcobbs-bentley: Also, "extracts" in the description to me is ambiguous. It is parsing the number and returning its value, but also removing that number from `wkt`. These observations apply to all of the `Get...` functions below.

iModelCore/libsrc/csmap/Source/CS_gtxFile.c
AlainRobertAtBentley: Fixed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Fixed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Fixed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Flawed code eliminated

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: fixed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Fixed

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
AlainRobertAtBentley: Fixed as part of resolving architectural issue above.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: I have changed all .size() to .length() for strings to normalize.

AlainRobertAtBentley: The proposed change was integrated, thanks.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: I made a massive review of the documentation for the parser classes. It will not yet be perfect yet but I will try to ameliorate it in the course of future pull-requests.

tcobbs-bentley: GitHub isn't happy trying to show me changes now, but line 4113 now has "aqnd" instead of "and".

AlainRobertAtBentley: I corrected again a batch of typos. Both diff programs and AI assistant are struggling with the size of the source file. I intend to split the file eventually.

iModelCore/libsrc/csmap/Source/CS_gtxFile.c
aruniverse: Verified at latest head `e026c1fe`: `CS_gtxFile.c:851` now computes the short header from `sizeof(nColumns) + sizeof(nRows)`, so the LP64 offset defect is fixed. The requested known-cell LP64 regression test has not been added, so the code fix is addressed but the test portion remains open.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified at latest head `e026c1fe`: `GetRidOfSection` now advances past both delimiters (`endPosRight + 1` and `endPosLeft + 1` at current lines 4808/4816), fixing the premature enclosing-section termination/no-progress failure. The requested nested-METADATA regression test was not added, so the implementation is fixed but coverage remains open.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: This is only partially addressed at `e026c1fe`. Moving the target-specific `DatumConverter` into `VerticalDatum` avoids storing it on the dictionary transform, but the cache still keys only on the target vertical name at line 26570 even though the converter is built from the horizontal `targetGCS` at line 26597. Also, `GetDirectVerticalDatumTransform` still returns dictionary-owned transform pointers, and each `VerticalDatum::ReleaseTransforms` can release their shared grid handles. Two horizontal targets sharing a vertical datum can therefore reuse the wrong converter, and one owner can still invalidate another. Please include horizontal target identity in the cache and make mutable transform runtime state per owner (or clone transforms), with sequential/concurrent coverage.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified addressed at latest head `e026c1fe`: direction-dependent horizontal conversion was removed. `VerticalDatum::InitializeTransforms` now consistently creates the converter from `targetGCS` to the transform's required horizontal CRS at lines 26589-26598; direct/inverse vertical direction is handled only when applying the elevation offset.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified addressed at latest head `e026c1fe`: `ConvertLatLong3D` now stores its status at line 26697 and stops before `GetElevation` when conversion fails (lines 26698-26701), so grid lookup no longer proceeds with invalid/unconverted coordinates.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified addressed at latest head `e026c1fe`: non-null candidates are now validated first, failure returns at line 19245, and `m_verticalDatum` is assigned only after validation at line 19247. Explicit null still clears immediately.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
aruniverse: Verified addressed at latest head `e026c1fe`: the public pure-virtual contract is restored to the original no-argument `InitializeTransform()` at `BaseGeoCoord.h:6290`, matching `main`. Target-horizontal conversion context was moved outside the transform API.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Partially addressed at latest head `e026c1fe`: the numeric/string extraction helpers now document `wkt` as `[in,out]` and state that the parsed text is removed (for example current lines 4113 and 4155), and the reported typo is gone. A few mutating methods still say `[in]`, notably `GetKeyword` around line 6261 while it replaces `wkt` at 6298, plus `GetProjected`/`GetCompound`.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -26926,15 +26573,42 @@ StatusInt VerticalDatum::InitializeTransforms(const Utf8String& target, const Ge
...
+
+            // Check if a datum transform is required for this vertical transform
+            Utf8String requireHorizontalCRS = transform->GetRequiredHorizontalCRSBase();
+
+            if (requireHorizontalCRS.length() > 0)
+                {
+                BaseGCSPtr transformHorizontalCRSBase = BaseGCS::CreateGCS(requireHorizontalCRS.c_str());
+
+                if (transformHorizontalCRSBase.IsValid())
+                    datumConverter = DatumConverter::Create(*targetGCS, *transformHorizontalCRSBase);
```
aruniverse, line 26693: **High:** a non-empty `requireHorizontalCRS` means this transform requires coordinates in another horizontal CRS, but failure to create that CRS or its `DatumConverter` leaves `datumConverter == nullptr` and still pushes the transform at line 26611. `GetElevation` then treats null as “no conversion needed” and samples the grid with unconverted target-datum coordinates, potentially returning a plausible but wrong elevation. Please distinguish a proven null/equivalent converter (the `IsNullTransform()` case) from converter-creation failure and propagate the latter; add a test where the required horizontal path is unavailable.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -4576,533 +4984,618 @@ GeoCoordParseStatus GetGeographicToCoordSys (Utf8StringR wkt, Utf8StringR geogra
...
             return GeoCoordParse_ParseError;
         }
 
-    if (datumValid && datumPresent)
-        return GeoCoordParse_Success;
-    else if (datumPresent && datumStatus != GeoCoordParse_Success)
-        return datumStatus;
-    else
-        return GeoCoordParse_UnknownDatum;
+    return PostGetProjected(baseGCS, name, authorityID, geocsPresent, geocsValid, "WKT2");
```
aruniverse: **Medium:** this returns post-processing success without requiring `sectionCompleted`. If an otherwise valid `PROJCS` ends at EOF without its final `]`, the loop exits because `wkt` is empty and this can still return success. Conversely, after a closing `]`, arbitrary trailing text is left in `wkt` and `Process` returns the child status without checking the residue. The same pattern appears in the other section loops. Please require `sectionCompleted`, reject non-whitespace trailing input at the top level, and add missing-final-bracket/trailing-garbage tests.

aruniverse (commented):
Follow-up review of the latest head found three additional correctness issues. Existing review threads were replied to separately with verified dispositions.

**High — VERTCON input datum is not declared (`basegeocoord.cpp:9361`):** unlike the geoid-grid transform, `VerticalOffsetGridTransform` never sets `m_requiredHorizontalCRSBase`, so the new per-transform conversion layer passes target-CRS coordinates straight through. The underlying CSMap API explicitly names the input `ll83` (`CSselectVertconUS`/`CScalcVertconUS`, `cs_map.h:8522-8524`), so VERTCON lookup requires NAD83 coordinates. Please set the required base to `LL83` and test the same cell from NAD27/WGS84 and NAD83 targets. This line is not on an added diff line, so this finding is included here rather than inline.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Good find! Corrected.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: I added code to make sure the section is completed. I will not add fail code when the residue of parsing is non-white chars because the various WKT clauses can be imbricated within other WKT clauses and for upcoming WKT2 the geodetic transform clauses are provided as separate clauses that may follow. (This can also apply to WKT too though we do not support that just yet)

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: changed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified at latest head `bb298b2e`: this specific failure mode is fixed. A non-empty required horizontal CRS now requires a successfully created `DatumConverter` (current lines 26688-26699); null is accepted only after `IsNullTransform()` proves the conversion is equivalent (26701-26708). No regression test was added. I found a separate cleanup regression on this new failure branch and posted it inline.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Partially addressed at latest head `bb298b2e`: `sectionCompleted` checks now reject EOF before the closing delimiter in `PROJCS` and the other loop-based sections (for example lines 5124-5125). The top-level trailing-residue half remains intentionally unaddressed. Also, `GetCompound` still returns at line 5192 without verifying or consuming the outer `COMPD_CS` closing `]`, so a compound WKT missing only that final bracket can still succeed. Please fix the compound delimiter ownership and add targeted missing-bracket tests.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified at latest head `bb298b2e`: the mutating parser helpers now consistently mark `wkt` `[in,out]` and describe removal, including `GetProjected`, `GetCompound`, and `GetKeyword`. Minor cleanup remains in `GetCompound`: the comment says `COMP_CS` instead of `COMPD_CS`, and the implementation currently does not remove the outer closing delimiter despite the documentation claim.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -4576,533 +5083,630 @@ GeoCoordParseStatus GetGeographicToCoordSys (Utf8StringR wkt, Utf8StringR geogra
...
             return GeoCoordParse_ParseError;
         }
 
-    if (datumValid && datumPresent)
-        return GeoCoordParse_Success;
-    else if (datumPresent && datumStatus != GeoCoordParse_Success)
-        return datumStatus;
-    else
-        return GeoCoordParse_UnknownDatum;
+    if (!sectionCompleted)
```
aruniverse, line 5124: **Medium regression:** this new completion guard rejects the intentionally supported `PROJCS` form with no linear `UNIT`. `GetProjectionToCoordSys` explicitly accepts that form at lines 6086-6090, but it consumes the enclosing `PROJCS` `]`; control returns here with an empty `wkt` and this parent `sectionCompleted` still false, so a form that previously used the default meter now returns `GeoCoordParse_ParseError`. Please leave the parent delimiter for `GetProjected` to consume (or explicitly communicate that it was consumed), and add a missing-`UNIT` regression test.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -26926,15 +26669,46 @@ StatusInt VerticalDatum::InitializeTransforms(const Utf8String& target, const Ge
...
+            {
+                BaseGCSPtr transformHorizontalCRSBase = BaseGCS::CreateGCS(requireHorizontalCRS.c_str());
+
+                if (transformHorizontalCRSBase.IsValid())
+                    datumConverter = DatumConverter::Create(*targetGCS, *transformHorizontalCRSBase);
+
+                if (nullptr == datumConverter)
+                {
+                    status = GEOCOORDERR_InvalidGeodeticTransform;
+                    break;
```
aruniverse, line 26698: **Medium:** this failure branch occurs after `transform->InitializeTransform()` at line 26679 but before the transform is inserted into `m_initializedTransforms` at line 26711. The error cleanup iterates only that vector, so it never releases the just-initialized grid handle. This persists until dictionary teardown; `VerticalOffsetGridTransform` also has an empty destructor, so its `m_vertconUS` allocation is not reclaimed there. Please clean up the current transform on this path—preferably by making initialized runtime state per-owner/RAII—and add the unavailable-required-converter test requested in the original thread.

aruniverse (commented):
Follow-up review of the latest correction commits. The required-horizontal-converter failure and VERTCON LL83 fixes are present. The prior horizontal-cache/shared-transform ownership blockers remain open in their existing thread. Two regressions introduced or exposed by the latest fixes are inline. Current aggregate validation is also red because the iTwinJS integration-test job failed.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -1473,6 +1680,7 @@ BaseGCS::ProjectionCodeValue GetProjectionCodeFromParseName (Utf8StringR name) c
     else if ((upperMethodName == "CASSINI") ||
              (upperMethodName == "CASSINI_SOLDNER") ||
              (upperMethodName == "CT_CASSINISOLDNER") ||
+             (upperMethodName == "HYPERBOLIC CASSINI-SOLDNER") ||
```
aruniverse: **High:** Hyperbolic Cassini-Soldner is a distinct projection method (different northing calculation), not an alias for ordinary Cassini-Soldner. GeoCoord/CS-Map has no corresponding hyperbolic implementation, so accepting this name as `pcvCassini` silently produces incorrect coordinates instead of reporting an unsupported method. Please remove this alias unless the hyperbolic equations are implemented, and add a test that rejects it (or validates a real implementation).

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -2013,6 +2253,10 @@ GeoCoordParseStatus SetParameterToCoordSys(Utf8StringR parameterName, Utf8String
                 if (SUCCESS != coordinateSystem.SetCentralPointLongitude(parameterValue * conversionToDegree))
                     return GeoCoordParse_InvalidParamForMethod;
                 break;
+            case BaseGCS::pcvCzechKrovak:
+            case BaseGCS::pcvCzechKrovakModified:
+                if (SUCCESS != coordinateSystem.SetPoint1Longitude(parameterValue * conversionToDegree))
```
aruniverse, line 2258: **High:** this branch applies every generic longitude-of-origin/center alias to Krovak `point1Longitude`, even though the oblique-pole longitude is a separate parameter handled below (`OBLIQUE POLE LONGITUDE`). It then falls through to `default` because there is no `break`, setting the same value as origin longitude too. Authority-less Krovak WKT can therefore overwrite the expected oblique-pole value (commonly 42.5°) with the origin/center longitude. Please let generic longitude aliases set only origin longitude for Krovak, reserve `point1Longitude` for the explicit oblique-pole alias, and add a parameter-level Krovak test.

aruniverse (commented):
Re-checked the current head. There have been no commits or thread replies since the previous review, so the four existing open findings remain unresolved. A fresh full-diff pass found five additional correctness issues, posted inline. The failing iTwinJS integration job appears to fail during hub setup before native behavior is exercised; native/platform builds passed.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -13645,6 +14314,16 @@ StatusInt BaseGCS::ToVerticalJson(BeJsValue jsonValue) const {
     if (!IsValid())
         return GEOCOORDERR_InvalidCoordSys;
 
+    VerticalDatumPtr verticalDatum = GetVerticalDatum();
+    if (verticalDatum.IsValid())
+        {
+        VerticalDatumInfoPtr verticalDatumInfo = verticalDatum->GetVerticalDatumInfo();
+        if (verticalDatumInfo.IsValid())
+            verticalDatumInfo->ToJson(jsonValue);
```
aruniverse: **High:** the full vertical CRS is written directly into the existing top-level `verticalCRS` object, but `FromVerticalJson` passes that inner object to `SetVerticalDatumFromJson`, while `VerticalDatumInfo::CreateFromJson` only parses an object containing another nested `verticalCRS` member (`10153-10156`). Round-tripping therefore fails full-definition parsing and falls back to legacy `id`, reducing specific datums such as `EGM96 height` to generic `GEOID`. Please make the reader accept the direct inner object (or make writer/reader shapes consistent). Also fix `VerticalDatumSerializationTests.cpp:351` to serialize `testGCS2`, not `testGCS1` twice, and assert the specific datum survives deserialization.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -26229,6 +25916,62 @@ StatusInt              Datum::GetGridFile (GridFileDefinition& gridFileDef, bool
...
+                        gridFileNames += gridFileDef.GetFileName();
+                        }
+                    }
+
+                    if (!cumulAllTransforms)
+                        return SUCCESS;
+                }
+            }
+
+            return SUCCESS;
```
aruniverse: **Medium:** this returns `SUCCESS` whenever a path exists, even if it contains no grid transform or the grid transform has zero definitions. The method also never clears `gridFileNames`, so callers can receive stale pre-existing content with a success status, contrary to the public contract. Please clear the output at entry, track whether at least one filename was appended, and return `ERROR` when none was found; add empty/no-grid/stale-output tests.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -5111,356 +5715,575 @@ GeoCoordParseStatus GetProjectionToCoordSys (Utf8StringR wkt,double conversionTo
...
-#endif
+    // CSMAP only supports prime meridian values other than Greenwich for
+    // lat/long GCS all other projections must use Greenwich.
+    if (BaseGCS::pcvUnity == coordinateSystem.GetProjectionCode() &&
+        (allowNonGreenwich ||
+         (name == "Ferro" && doubleSame(longitude, -17.666666666666)) ||
+         (name == "FerroPrecise" && doubleSame(longitude, -17.6665931666667))))
+        {
+        // Correcting an ancestral CSMAP typo error that uses 74.08175 instead of 74.08091666666667 as should be
+        if (longitude > -74.090 && longitude < -74.08)
```
aruniverse, line 5939: **Medium:** this broad 0.01° interval rewrites *any* permitted non-Greenwich prime meridian between -74.090° and -74.080° to the Bogota value, potentially shifting a legitimate custom meridian by hundreds of metres. The compatibility case appears to be the exact legacy Bogota value emitted by `CScs2Wkt.cpp` (`-74.08175`). Please gate the correction on the Bogota name/authority and a tight comparison to that known bad value instead of snapping every longitude in this range.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: That opened a can of worms. I had to go back to the original developer of the Vertical transformations. The fix removes the InitializeTransform()/ReleaseTransform() from the VerticalTransform class. Yes this implies that the transforms are only released when the dictionary is terminated but the allocated resources are extremely small so it will not be an issue.
The list of transforms and associated datum converters were moved to the class VerticalDatumConverter which cannot have source or target modified making the list of transforms applicable in all cases.
There may still be other issues but the original developer and myself are studying it.

aruniverse: Verified at latest head `269cf1ae`: the refactor fixes two important parts of this thread. `VerticalDatumConverter` now clones fixed source/target GCS definitions (current lines 11638-11640), owns the target-specific datum converters, and `ReleaseTransforms` destroys only those per-converter objects (11810-11821), so one converter no longer releases another converter's shared grid handle. This is still only partially addressed: transform-path selection depends on the first `latLong` at 11754-11758 but is then cached behind one boolean at 11729-11730, and dictionary-owned grid transforms are still lazily initialized shared mutable objects without synchronization. No regional-repeat or concurrency tests were added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified the additional compound-section fix at latest head `269cf1ae`: `GetCompound` now requires and consumes the outer `COMPD_CS` closing delimiter at current lines 5188-5191. The missing-final-delimiter portion is addressed. The top-level trailing-residue behavior remains intentionally unchanged, so the overall thread is still partial; no targeted parser regression tests were added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified the code regression is fixed at latest head `269cf1ae`: the missing-`UNIT` path now marks its helper section complete without consuming the parent delimiter at lines 6091-6094, and `GetProjected` consumes that delimiter at 5114-5117. The requested missing-`UNIT` regression test was not added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified the original cleanup leak path is structurally removed at latest head `269cf1ae`: required-datum-converter creation now occurs before any grid initialization, while grid initialization is lazy inside `GetElevation`; both grid subclasses release their handles from their destructors. The requested unavailable-converter regression test was not added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: I removed the removal of the PROJCS terminal ] in GetProjectionToCoordSys() the bracket will be there regardless of valid GetProjectionToCoordSys() exit condition

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: CSMAP code does not recognise any other values for prime meridians than specific ones. In fact CSMAP rounds the prime meridian value to an integer before selecting what to write in a WKT. There will not ever be another Prime Meridian in the range -75 to -74. I could have tested on the full degree range and it would have been valid but the CSMAP coded typo was within the specified range. Note that the fact this specific Prime Meridian has never been corrected is the result of the fact it is never used in real situation. I just fixed it to pass extensive unit tests.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: fixed

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: The fall through has been corrected. As for the rest the Krovak projection cannot be fully defined by the WKT parameters since some are assumed. For this projection we hope that the EPSG code is set at which time we will complete the projection parameters. In short support of this projection has always been crappy because it was designed too specific.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: Good find. I corrected. In PP we only used one of the paths so the error was never located. I added unit tests.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: gridFileNames is cleared. If there is no file in the list then SUCCESS is returned because this is what it is.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified at latest head `5544ad24`: the `HYPERBOLIC CASSINI-SOLDNER` alias has been removed from the ordinary Cassini mapping (current lines 1680-1684), so unsupported hyperbolic WKT will no longer silently select `pcvCassini`. No dedicated rejection test was added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified only the fall-through portion at latest head `5544ad24`: the Krovak branch now has a `break` at current line 2260. The generic longitude aliases still set `point1Longitude` for Krovak at 2256-2259 even though the explicit `OBLIQUE POLE LONGITUDE` branch separately owns that parameter, so the broader parameter-mapping concern remains partial.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified addressed at latest head `5544ad24`: `FromVerticalJson` now parses the direct inner vertical object, standalone setters keep the wrapper shape, extent objects are materialized before serialization, and parse failure without a legacy `id` is propagated. The previous test typo now serializes `testGCS2`, and focused wrapper/serialization tests were added.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified addressed at latest head `5544ad24`: `GetGridFileNames` clears the output at entry (current line 26131) and returns `ERROR` when no filename was appended (26173-26178), so stale output and false success are both fixed. I did not find a focused test for the empty/no-grid cases.

aruniverse (commented):
Follow-up review at `5544ad24`. The latest JSON/test corrections address the prior serialization and compile findings, but several additional correctness issues remain. The highest-value changed-line findings are inline. Additional findings that are not on currently commentable diff lines:

- **High — transform serialization errors are discarded:** `VerticalGeodetic3dTransform::ToJson` returns `ERROR` (`basegeocoord.cpp:9811-9817`), but `VerticalDatumInfo::ToJson` and `BaseGCS::ToVerticalJson` ignore child status and report success (`10265-10283`, `14533-14540`), producing truncated target-only JSON.
- **High — stored paths are not validated against their declared source/target:** `VerticalTransformPathInfo::CreateFromJson` accepts any path (`9980-10013`); resolution applies those edges as the owning datum's conversion (`10741-10756`).
- **Medium — path equality accepts a proper prefix:** `VerticalTransformPathInfo::operator==` never compares vector lengths (`9961-9978`).
- **Medium — malformed extents default to zero and inverted ranges pass:** `DictionaryValueExtentLatLong` accepts missing/non-numeric coordinates as `0.0`, while validation checks only outer bounds (`11017-11038`, `10161-10169`).
- **Medium — direct/predefined paths bypass point applicability:** direct and stored-path returns occur before the extent-filtered recursive search (`10724-10782`).

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
```diff
@@ -6237,37 +6227,43 @@ virtual VerticalTransformPtr CreateReverseCopy();
...
-* in preparation for a following GetElevation() call.
-* @return   SUCCESS, ERROR or GEOCOORDERR_***, will be defined by the subclass.
+* Returns the name of the base geodetic datum for the elevation transformation. This
+* geodetic datum is a latitude/longitude base to which latitude and longitude must be converted
+* prior to applying the transformation using GetElevation()
+* @return   The name of the base geodetic datum or an empty string if none is required.
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-virtual StatusInt InitializeTransform() = 0;
+virtual Utf8String GetRequiredHorizontalDatumBase() const {return "";}
```
aruniverse, line 6249: **Blocker — public API/ABI regression:** this hierarchy previously exposed virtual `InitializeTransform()` and `ReleaseTransform()` slots (and `VerticalDatum::GetElevation(...)`). The PR removes them, so existing subclasses using `override` stop compiling and subsequent vtable slots shift for already-built consumers. Lazy internal lifecycle can coexist with deprecated/default compatibility virtuals; please retain the old slots or explicitly version the ABI.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
```diff
@@ -6360,6 +6358,10 @@ class VerticalDatumInfo : public RefCountedBase
     BASEGEOCOORD_EXPORTED void GetRemarks(Utf8String& remarks) const;
     BASEGEOCOORD_EXPORTED void GetUnits(Utf8String& units) const;
     BASEGEOCOORD_EXPORTED void GetExtent(DRange2d& extent) const;
+    BASEGEOCOORD_EXPORTED bool IsDeprecated() const;
```
aruniverse: **High:** this exported `VerticalDatumInfo::IsDeprecated()` declaration has no implementation anywhere in GeoCoord. Any consumer that calls it gets an unresolved symbol at link time. Please add the missing definition and a direct API test.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
```diff
@@ -6538,7 +6540,7 @@ StatusInt AddVerticalDatumTransform(VerticalTransformPtr& transform);
 *               GEOCOORDERR_NoTransforms - no transforms were found
 * @bsimethod
 +---------------+---------------+---------------+---------------+---------------+------*/
-StatusInt GetVerticalDatumTransforms(bvector<VerticalTransformPtr>& transforms, const Utf8String& from, const Utf8String& to, const GeoPoint& latLong);
+StatusInt GetVerticalDatumTransforms(bvector<VerticalTransformPtr>& transforms, const Utf8String& from, const Utf8String& to, const GeoPoint* latLong);
```
aruniverse, line 6560: **Medium API compatibility:** changing the public parameter from `const GeoPoint&` to `const GeoPoint*` fixes the implementation mismatch but breaks existing source callers. The documentation also does not say `nullptr` is accepted or that it disables extent filtering. Please retain a reference overload and route both forms to an internal nullable helper, or document/version the breaking contract.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -9544,82 +10132,86 @@ VerticalDatumInfoPtr VerticalDatumInfo::CreateFromJson(BeJsConst jsonVerticalCRS
...
+            // transforms paths must have target string
+            if (transforms[i].isObject() && transforms[i].isMember("target"))
             {
-                VerticalTransformPathInfoPtr transformPath = VerticalTransformPathInfo::CreateFromJson(transformPaths[i], vdatumInfo->m_crsName);
-                if (transformPath.IsValid())
-                    vdatumInfo->m_transformPaths.push_back(transformPath);
+                Utf8String target = VerticalDatumDictionary::DictionaryValueString(transforms[i], "target");
+                VerticalTransformPtr transform = VerticalTransform::CreateFromJson(transforms[i], vdatumInfo->m_crsName, target);
+                if (transform.IsValid())
+                    VerticalDatumDictionary::Get()->AddVerticalDatumTransform(transform); // Note: even if we are not adding the item to the dictionary we need to save the transform
```
aruniverse: **High:** transform registration mutates the process-global dictionary even when `addToDictionary == false`, before the rest of the datum is validated, and the return status is ignored. Duplicate source/target pairs are first-wins, while runtime conversion resolves from that global dictionary rather than this object’s transform list. Two local JSON definitions with the same names but different grids/offsets can therefore silently use the first definition; a parse that later fails can still leave global state behind. Please make registration transactional/per-definition and propagate duplicate errors.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -9544,82 +10132,86 @@ VerticalDatumInfoPtr VerticalDatumInfo::CreateFromJson(BeJsConst jsonVerticalCRS
...
+
+                // keep a list of the transforms in the VerticalDatumInfo too so we can compare whether two VerticalDatums
+                // are equal which included if their transforms are equal, transforms are used from the VerticalDatumDictionary
+                // and not directly from here, ths is only for comparison
+                vdatumInfo->AddTransform(transform);
```
aruniverse, line 10269: **Blocker:** if a transform object contains `target` but its payload is unsupported/invalid, `VerticalTransform::CreateFromJson` returns an invalid pointer. This line still stores it. `GetTransformTargetNames` and serialization later dereference the pointer unconditionally, so JSON such as `{ "target": "WGS84", "bogus": {} }` can crash public APIs. Reject the datum immediately when transform creation fails and add a malformed-transform test.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
AlainRobertAtBentley: This is new API only used internally so changes have zero impact. This PR activate the use of vertical datum even though the code already existed.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: We are aware of the problem and are planning to fix it. In the meantime we are ready to live with this strange behavior since all vertical datum and transform definitions originate from the pre-defined dictionary.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
AlainRobertAtBentley: fixed

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
AlainRobertAtBentley: I know that this is a change of an API but currently it is only used internally. Documentation has been modified.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: fixed.

tcobbs-bentley:
@AlainRobertAtBentley I am going to update your branch to use RapidJson instead of jsoncpp now that jsoncpp has been removed from the main branch.

tcobbs-bentley:
@AlainRobertAtBentley I have pushed a [commit](https://github.com/iTwin/imodel-native/pull/1531/changes/d9cb9cb999e8e0c9d9a28cd1fe55bce98e1ac71c) to remove jsoncpp code in favor of RapidJson.

tcobbs-bentley:
Nit: the updated `basegeocoord.cpp` and `BaseGeoCoord.h` now have a mix of indented braces (old-school Bentley code style) and non-indented braces (typical code style). It's probably not worth normalizing, since doing so would update so many lines of code, but it does make the code jarring to read.

AlainRobertAtBentley:
> Nit: the updated `basegeocoord.cpp` and `BaseGeoCoord.h` now have a mix of indented braces (old-school Bentley code style) and non-indented braces (typical code style). It's probably not worth normalizing, since doing so would update so many lines of code, but it does make the code jarring to read.

The difference in styles between PowerPlatform and imodel-nativer already makes the merge process a nightmare so I hesitate diverging sources even more for a simple code-style difference. I may do it gradually instead

tcobbs-bentley (approved):

tcobbs-bentley:
@aruniverse Can you resolve your comments or indicate what is wrong with the fixes?
@pmconne @bbastings Can one of you review?

pmconne (approved):

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
aruniverse: Verified at `4c2110d4`: `VerticalDatumInfo::IsDeprecated()` now has the inline definition `return m_deprecated;` (`BaseGeoCoord.h:6375`). `VerticalDatumInfoGettersTest` supplies `deprecated: true` and calls the getter directly (`VerticalDatumUnitTests.cpp:414,463`). The missing-symbol issue is addressed; the current PR CI checks are green.

iModelCore/GeoCoord/PublicAPI/GeoCoord/BaseGeoCoord.h
aruniverse: Verified at `4c2110d4`: the declaration/definition and internal pointer/null call sites agree, and `BaseGeoCoord.h:6551-6554` now explicitly documents `nullptr`. Given the internal, non-exported use, I am dropping the earlier compatibility-overload objection. One documentation qualification remains: the supplied point filters recursive fallback search, whereas direct and predefined paths return without checking it (`basegeocoord.cpp:10836-10891`). Please narrow the claim that all returned transforms are location-filtered.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
aruniverse: Verified at `4c2110d4`: a failed `VerticalTransform::CreateFromJson` now returns a null datum with `GEOCOORDERR_InvalidTransform` before registration or insertion (`basegeocoord.cpp:10256-10269`), so the reported unsupported-payload null dereference is fixed. The new negative tests cover a missing target and a nonobject entry, but not the target-present/unsupported-payload branch; please add the original `{ "target": "WGS84", "bogus": {} }` case to `VerticalDatumInfoInvalidTransformTest`. No remaining null-insertion defect was found.

PR #1534, opened 2026-08-01
Update curl and c-ares to build using vcpkg

### LINKED ISSUE #1532: Use vcpkg to build curl
Build curl with vcpkg instead of bmake.

tcobbs-bentley:
imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/1148


iModelCore/libsrc/curl/triplets/arm64-android.cmake
```diff
@@ -0,0 +1,23 @@
...
+# invalidates the cached binaries and forces the library to be rebuilt.
+# CACHE_BUST = 0
+
+set(VCPKG_TARGET_ARCHITECTURE arm64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE static)
+set(VCPKG_CMAKE_SYSTEM_NAME Android)
+set(VCPKG_CMAKE_SYSTEM_VERSION 28)
+set(VCPKG_MAKE_BUILD_TRIPLET "--host=aarch64-linux-android")
+set(VCPKG_CMAKE_CONFIGURE_OPTIONS -DANDROID_ABI=arm64-v8a)
```
aruniverse, line 12: This drops the previous Android `CURL_CA_PATH="/system/etc/security/cacerts"` setting (`vendor/lib/config-android.h`). Because this is a cross-build, curl skips CA-path auto-detection; the vcpkg port falls back to iTwinOpenSSL's `/etc/ssl` defaults, while the default BeSQLite curl setup supplies no Android `CURLOPT_CAINFO`. HTTPS verification will therefore stop using Android's system trust directory. Please add `-DCURL_CA_PATH=/system/etc/security/cacerts` here and in the x64 Android triplet (and disable an auto-detected bundle if needed).

iModelCore/libsrc/curl/BeCurl.mke
```diff
@@ -1,697 +1,176 @@
...
+%else
+
+    #----------------------------------------------------------------------------------
+    #   Dynamic delivery (Windows only): generate an export .def from curl's public headers, then
+    #   re-link iTwinCurl.dll + import lib from the vcpkg static libcurl.lib. Reuses the standard
+    #   dlmlink machinery (via linkLibrary.mki) so versioning, signing, the import library, and
+    #   BuildContext delivery behave as before. link.exe pulls the exported objects out of the
+    #   static .lib as directed by the .def.
+    #----------------------------------------------------------------------------------
+    $(o)$(appName).def : $(curlDir)makeCurlDef.py
```
aruniverse: The generated export list only depends on the Python script even though its contents come from the installed curl headers. After an incremental curl version/feature update, vcpkg can install a new archive and headers while bmake keeps the old `.def`, silently omitting newly added DLL exports (or retaining removed ones until a clean build). Please regenerate this target unconditionally or add the vcpkg archive/header or install-version stamp, plus the manifest/configuration, as prerequisites.

iModelCore/libsrc/curl/vcpkg.json
```diff
@@ -0,0 +1,28 @@
...
+      "name": "curl",
+      "default-features": false,
+      "features": [
+        "c-ares"
+      ],
+      "platform": "osx | linux"
+    }
+  ],
+  "overrides": [
+    { "name": "curl", "version": "8.21.0#1" },
```
aruniverse, line 23: This changes the shipped curl component to 8.21.0, but `curl/CurlNugetLicense.json`—the `LicenseFile` declared by `BeCurl.PartFile.xml`—still reports Curl 7.74.0. It also has no c-ares entry even though the macOS/Linux `iTwinCurl` archive now statically merges c-ares 1.34.8. Please update the package sublicense metadata (and inspect a generated CurlNuget) so the delivered versions/components are accurately reported.

iModelCore/libsrc/curl/triplets/arm64-ios.cmake
```diff
@@ -0,0 +1,26 @@
...
+set(VCPKG_CMAKE_SYSTEM_NAME iOS)
+set(VCPKG_OSX_ARCHITECTURES arm64)
+set(VCPKG_OSX_SYSROOT iphoneos)
+set(VCPKG_OSX_DEPLOYMENT_TARGET 13.0)
+
+# curl's OpenSSL backend verifies the peer certificate chain against the iOS system trust store
+# (Keychain) via Apple SecTrust. vcpkg_cmake_configure forwards this triplet option to curl's
+# CMake, which then auto-links Security/CoreFoundation/CoreServices. The option is also passed
+# to curl's transitive cmake deps (zlib), where it is an unused, harmless define.
+set(VCPKG_CMAKE_CONFIGURE_OPTIONS "-DUSE_APPLE_SECTRUST=ON")
```
aruniverse, line 19: Does this intentionally retire the no-network-fetch safeguard from the previous Apple verifier? The retained `rustls-ffi/README.md` documents `set_network_fetch_allowed(false)` as required to prevent Security framework background dispatch threads, but curl 8.21's SecTrust path calls `SecTrustEvaluateWithError` without `SecTrustSetNetworkFetchAllowed(false)`. The iOS CI job only builds/packages, so it does not exercise HTTPS or shutdown behavior. Please either preserve that safeguard with an overlay patch and a focused runtime smoke, or confirm it is obsolete and update/remove the stale rustls documentation.

iModelCore/libsrc/curl/BeCurl.mke
```diff
@@ -1,697 +1,176 @@
...
+always:
+    !~@mkdir $(o)
+
+#--------------------------------------------------------------------------------------
+#   Resolve the vcpkg-produced static libraries. c-ares is only built (and only linked) on
+#   macOS and Linux; the other platforms use curl's threaded resolver, so no libcares there.
+#--------------------------------------------------------------------------------------
+# curl's CMakeLists.txt defaults CMAKE_DEBUG_POSTFIX to "-d" on every platform, so the debug
+# archive is libcurl-d (release stays libcurl). c-ares sets no postfix, so libcares is unchanged.
+%if defined (DEBUG)
```
aruniverse, line 69: The PR adds this DEBUG-only artifact path and includes a commit specifically fixing a Windows debug build, but the linked Windows CI stages run only the `production` configuration. Could you attach a clean DEBUG build/link/load result for the supported MSVC and clang-cl configurations, or add a targeted validation job? The current green checks do not execute this branch.

aruniverse (commented):
Adversarial review of the linked curl/c-ares vcpkg migration. I found the following runtime, packaging, incremental-build, and validation gaps.

iModelCore/libsrc/curl/triplets/arm64-osx.cmake
```diff
@@ -0,0 +1,24 @@
...
+set(VCPKG_LIBRARY_LINKAGE static)
+
+set(VCPKG_CMAKE_SYSTEM_NAME Darwin)
+set(VCPKG_OSX_ARCHITECTURES arm64)
+
+# curl's OpenSSL backend verifies the peer certificate chain against the macOS system trust
+# store (Keychain) via Apple SecTrust. vcpkg_cmake_configure forwards this triplet option to
+# curl's CMake, which then auto-links Security/CoreFoundation/CoreServices. The option is also
+# passed to curl's transitive cmake deps (zlib, c-ares), where it is an unused, harmless define.
+set(VCPKG_CMAKE_CONFIGURE_OPTIONS "-DUSE_APPLE_SECTRUST=ON")
```
tcobbs-bentley, line 17: As far as I can tell, this is working correctly. However, I am waiting for testing from @alexdunae to verify that locally installed system CA certificates work properly with a local MITM proxy.

iModelCore/libsrc/curl/triplets/arm64-android.cmake
tcobbs-bentley: Fixed.

iModelCore/libsrc/curl/BeCurl.mke
tcobbs-bentley: Fixed.

iModelCore/libsrc/curl/vcpkg.json
tcobbs-bentley: Fixed.

alexdunae (approved):
Looking good from my review on Mac with MITM certificates.  I see traffic flowing over curl to `gcsworkspaceprodeussa01.blob.core.windows.net` as expected.

```
< HTTP/1.1 200 Connection established
<
* CONNECT phase completed for HTTP proxy
* CONNECT tunnel established, response 200
* ALPN: curl offers http/1.1
* SSL Trust Anchors:
*   Native: Apple SecTrust
*   OpenSSL default paths (fallback)
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384 / x25519 / RSASSA-PSS
* ALPN: server did not agree on a protocol. Uses default.
* Server certificate:
*   subject: CN=*.blob.core.windows.net; O=Microsoft Corporation
*   start date: Aug  9 09:43:50 2026 GMT
*   expire date: Aug 11 09:43:50 2027 GMT
*   issuer: CN=mitmproxy; O=mitmproxy
*   Certificate level 0: Public key type RSA (2048/112 Bits/secBits), signed using sha256WithRSAEncryption
*   subjectAltName: "gcsworkspaceprodeussa01.blob.core.windows.net" matches cert's "*.blob.core.windows.net"
* OpenSSL verify result: 14
* SSL certificate verified via Apple SecTrust.
* Established connection to localhost (::1 port 8100) from ::1 port 64495
* using HTTP/1.x
> GET /gcs/manifest.bcv? HTTP/1.1
Host: gcsworkspaceprodeussa01.blob.core.windows.net
Accept: */*
x-ms-date:Tue, 11 Aug 2026 16:43:49 GMT
x-ms-version:2020-10-02
```

iModelCore/libsrc/curl/triplets/arm64-osx.cmake
tcobbs-bentley: Testing succeeded.

iModelCore/libsrc/curl/triplets/arm64-ios.cmake
tcobbs-bentley: Retiring that is intentional. That was a workaround to a bug that caused a deadlock during shutdown when using rustls with that feature enabled. I have tested using my locally built add-on in itwinjs-core on iOS and macOS to verify that the deadlock does not exist with the new setup. I have now removed the whole `libsrc/rustls-ffi` directory, since it is no longer used.

iModelCore/libsrc/curl/BeCurl.mke
```diff
@@ -1,697 +1,181 @@
...
+    #   dlmlink machinery (via linkLibrary.mki) so versioning, signing, the import library, and
+    #   BuildContext delivery behave as before. link.exe pulls the exported objects out of the
+    #   static .lib as directed by the .def.
+    #
+    #   The .def is regenerated on every build because its content comes from the vcpkg-installed
+    #   curl headers, whose timestamps are not a reliable prerequisite (a binary-cache hit can
+    #   restore older ones). makeCurlDef.py rewrites the file only when the export list actually
+    #   changes, so an unchanged .def does not force a re-link.
+    #----------------------------------------------------------------------------------
+    always:
```
aruniverse, line 140: Regenerating the `.def` in `always:` fixes the stale export file itself, but a changed `.def` still does not invalidate the final `iTwinCurl.dll`: the shared Windows DLL rule consumes `DLM_MODULE_DEF` as a linker setting without making it a prerequisite of the DLL target. If vcpkg restores an older-timestamp libcurl archive and the export set changes, BBMake can update the `.def` (and the MSVC import library) while leaving the existing DLL current; new imports can then link but fail at load time. Please make the final DLL depend on the generated `.def`, or invalidate only that DLL when `makeCurlDef.py` changes the export bytes.

aruniverse (commented):
Follow-up review of the latest fixes found one remaining incremental-build correctness issue.

iModelCore/libsrc/curl/triplets/arm64-android.cmake
aruniverse: Confirmed at current head `121ef4a`: both Android triplets now set `CURL_CA_PATH=/system/etc/security/cacerts` and disable bundle lookup with curl's documented `CURL_CA_BUNDLE=none` sentinel. This restores the previous hashed Android system-store behavior; no regression found in the fix. Thanks.

iModelCore/libsrc/curl/BeCurl.mke
aruniverse: Partially addressed at current head `121ef4a`: unconditional generation plus the byte comparison correctly refreshes a changed `.def` and preserves its timestamp when unchanged. However, the final DLL target does not depend on `DLM_MODULE_DEF`, so an older-timestamp cache restore can still leave `iTwinCurl.dll` stale after the export list changes. I posted the remaining issue here: https://github.com/iTwin/imodel-native/pull/1534#discussion_r3760476089

iModelCore/libsrc/curl/vcpkg.json
aruniverse: Confirmed at current head `121ef4a`: `CurlNugetLicense.json` now reports curl 8.21.0 and includes c-ares 1.34.8, matching the manifest pins. The JSON parses successfully and remains wired as the CurlNuget `LicenseFile`. Thanks.

iModelCore/libsrc/curl/triplets/arm64-ios.cmake
aruniverse: Confirmed at current head `121ef4a`: retiring the workaround is intentional, the replacement was exercised on iOS/macOS without the shutdown deadlock, the Mac MITM test verifies Apple SecTrust accepts a locally installed CA, and the now-unused `rustls-ffi` implementation and stale documentation have been removed. This addresses the concern. Thanks.

DanRod1999 (approved):

iModelCore/libsrc/curl/BeCurl.mke
tcobbs-bentley: I have successfully built local debug builds (both MSVC and clang) in Windows, but posting the whole output here is not practical. However, here is the end of the Windows MSVC debug build:
```
Built 123 of 123 items


========================================
  Build Succeeded
========================================


[Build] Tue Aug 11 2026 13:19:50, elapsed time: 00:07:26
```

Here is the end of the Windows clang debug build:

```
Built 123 of 123 items


========================================
  Build Succeeded
========================================


[Build] Tue Aug 11 2026 13:59:36, elapsed time: 00:19:28
```

iModelCore/libsrc/curl/makeCurlDef.py
```diff
@@ -0,0 +1,84 @@
...
+    # timestamp and forcing a needless re-link of iTwinCurl.dll.
+    if os.path.exists(out_def):
+        with open(out_def, "rb") as f:
+            if f.read() == data:
+                return 0
+
+    with open(out_def, "wb") as f:
+        f.write(data)
+
+    # The export list changed, so anything previously linked from the old .def is stale.
```
aruniverse: The invalidation needs to happen before committing the new `.def`. On Windows, deleting `iTwinCurl.dll` can fail while another process has it loaded. The script has already written the new `.def` at that point, so after the DLL is released the next build returns early because the export bytes match and never retries the deletion; the old DLL can remain current. Please delete the stale outputs first and only then replace/write the `.def`, so a failed deletion leaves the old export file and the next build retries.

aruniverse (commented):
Latest-head follow-up found one retry-path hole in the stale-DLL fix.

iModelCore/libsrc/curl/BeCurl.mke
aruniverse: Partially addressed at `baee6e9`: passing the output DLL to `makeCurlDef.py` closes the ordinary changed-export path by deleting the DLL before the link rules run. A retry hole remains when Windows refuses that deletion because the DLL is loaded: the new `.def` has already been written, so the next build returns before retrying deletion. I posted the remaining case here: https://github.com/iTwin/imodel-native/pull/1534#discussion_r3761465875

iModelCore/libsrc/curl/BeCurl.mke
aruniverse: The supplied MSVC and clang DEBUG build results verify that both configurations compile and link all 123 items successfully. They do not include the requested addon/DLL load check (or identify the tested head), so I consider the build/link portion addressed but the runtime-load evidence still outstanding. Thanks for adding the two build results.

iModelCore/libsrc/curl/BeCurl.mke
tcobbs-bentley: Fixed.

iModelCore/libsrc/curl/makeCurlDef.py
tcobbs-bentley: Fixed.

aruniverse: Confirmed at current head `ecd0dfb`: stale outputs are now deleted before the changed `.def` is written. A locked-DLL deletion failure returns with both the old DLL and old `.def` intact, so the next build still sees changed exports and retries. I verified the clean, unchanged-export, successful invalidation, simulated locked-DLL failure/retry, and partial-write recovery paths. This addresses the retry issue. Thanks.

iModelCore/libsrc/curl/BeCurl.mke
aruniverse: Fully addressed at current head `ecd0dfb`. The generator still runs on every build and preserves unchanged timestamps; when exports change it now invalidates the old DLL before committing the new `.def`, and a failed Windows deletion preserves the old `.def` so invalidation is retried. Focused tests covered clean, unchanged, changed-export, and locked-DLL retry paths. Thanks.

aruniverse (approved):

iModelCore/libsrc/curl/BeCurl.mke
tcobbs-bentley: Fixed.

PR #1535, opened 2026-08-03
vcpkg cleanup:

### LINKED ISSUE #1536: Windows debug builds failing
Windows debug builds are failing with the following error:

```
pugixml.lib(pugixml.cpp.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '2' doesn't match value '0' in BePugiXml.obj
pugixml.lib(pugixml.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MDd_DynamicDebug' doesn't match value 'MD_DynamicRelease' in BePugiXml.obj
```

tcobbs-bentley:
* Fix pugixml debug builds
* Stop using OutRoot for vcpkg downloads and registries

This pull request introduces important improvements to the vcpkg integration in the build system, focusing on two themes: (1) robust handling of Windows debug builds with release CRT, especially for libraries like `pugixml`, and (2) improved caching strategies for downloads and registries to make builds more reliable and efficient across platforms and architectures.

**Handling Windows Debug Builds with Release CRT:**

* Added detailed documentation in `.github/skills/vcpkg/SKILL.md` on the pitfall where Windows debug builds link the release CRT, including how to detect affected libraries and apply a two-part fix (force release-only triplets and always link the release archive) with concrete examples. [[1]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1R47-R53) [[2]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1R155-R175) [[3]](diffhunk://#diff-b5c6715da7d77c00c2fca8a96f4ed44d3547a8751e9175fb9990417e0862dbd1R202-R236)
* Updated `pugixml` build logic in `pugixml.mke` to always link the release archive on Windows, preventing CRT and `_ITERATOR_DEBUG_LEVEL` mismatches in debug builds.
* Modified `pugixml` Windows triplet files (`x64-windows-static.cmake` and `x64-windows-static-clang.cmake`) to force release-only builds by setting `VCPKG_BUILD_TYPE release`. [[1]](diffhunk://#diff-b2fb602cc9278f8dcf8009db06692295568b2dc7adefbca92fa4f30f3b07f8ccR24-R27) [[2]](diffhunk://#diff-e39e777240043548995303380179007ade8a73b0324ed7632129535f22be3ee0R13-R19)

**Improved Caching for Downloads and Registries:**

* Changed `vcpkg_run_install.bat` and `vcpkg_run_install.sh` to use persistent, per-platform (per-arch) downloads and registries caches under the vcpkg source root. This avoids repeated downloads and extraction across clean builds, and prevents race conditions when building for multiple architectures in parallel. [[1]](diffhunk://#diff-dd259756ffac3e4b2e88aa569c40ec7e3bce2b438ece026d65d9ce6add45115cL140-R177) [[2]](diffhunk://#diff-de01d0aa5c493a0807ef6b375beb8532ecc5dfcab33a962be0da8af9ec405c01L94-R123)

These changes make the build system more robust for Windows development and speed up multi-platform builds by improving cache persistence and isolation.


iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -137,22 +137,44 @@ if not exist "%VCPKG_EXE%" (
...
+rem      (e.g. x64-windows, arm64-android, x64-android), so parallel builds of different arches
+rem      (notably AndroidARM64 vs AndroidX64) get separate downloads\<platform>\tools trees and
+rem      cannot race on tool extraction. Triplet variants of the same platform (static / -md /
+rem      -veracode / -clang) and different configs (debug/release) intentionally share one
+rem      persistent cache; the sequential install chain (vcpkg.PartFile.xml) serializes
+rem      libraries within an arch, and once a tool is extracted no further extraction (hence no
+rem      race) occurs for that platform.
+rem The binary cache (VCPKG_DEFAULT_BINARY_CACHE) remains shared so compiled packages are not
+rem rebuilt redundantly.
+for /f "tokens=1,2 delims=-" %%a in ("%TRIPLET%") do set "VCPKG_PLATFORM_KEY=%%a-%%b"
```
aruniverse: The first-two-token key prevents the Android ARM64/x64 collision, but it merges every same-platform invocation (`x64-windows-static`, `-clang`, `-md`, `-veracode`, debug/release) into one mutable downloads/tools tree and registry Git repo. `vcpkg.PartFile.xml` serializes consumers only inside one BentleyBuild graph; two same-platform builds can still run concurrently. PR #1497 established that concurrent registry fetch/GC can corrupt this state. Please keep these caches per install root, add a build/invocation discriminator, or hold a cross-process lock around the shared-cache `vcpkg install`. The shell implementation has the same issue at lines 108-123.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -137,22 +137,44 @@ if not exist "%VCPKG_EXE%" (
...
+set "VCPKG_DOWNLOADS=%VCPKG_ROOT%\downloads\%VCPKG_PLATFORM_KEY%"
```
aruniverse: This newly requires `VCPKG_ROOT` to be writable. The wrapper can fall back to `%VCINSTALLDIR%\vcpkg` (normally under protected Program Files), and POSIX callers may use a shared read-only `IMODEL_VCPKG_ROOT`. It also overwrites/ignores the standard `VCPKG_DOWNLOADS` escape hatch (the shell wrapper always passes its computed root). Please honor caller-provided download/registry cache paths and, when unset, fall back to `INSTALL_ROOT` if the persistent directory cannot be created.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
```diff
@@ -10,3 +10,10 @@
 set(VCPKG_TARGET_ARCHITECTURE x64)
 set(VCPKG_CRT_LINKAGE dynamic)
 set(VCPKG_LIBRARY_LINKAGE static)
+
+# Force full release builds in vcpkg for this triplet.
+# Reason: bmake link settings in this pipeline use the release CRT even in DEBUG builds, so
+# pugixml must be produced as release artifacts to avoid CRT ('MDd_DynamicDebug' vs
+# 'MD_DynamicRelease') and _ITERATOR_DEBUG_LEVEL (2 vs 0) link mismatches against our
+# BePugiXml wrapper object.
+set(VCPKG_BUILD_TYPE release)
```
aruniverse: `VCPKG_BUILD_TYPE release` changes more than the CRT: the CMake Release configuration adds optimization and `NDEBUG`, stripping pugixml's internal `assert(...)` checks. Before the vcpkg migration, `pugixml.cpp` was compiled by bmake in DEBUG builds, so this silently removes library diagnostics from Windows DEBUG. Prefer a Debug archive configured with `/MD` and `_ITERATOR_DEBUG_LEVEL=0` but without `NDEBUG`; if release-only pugixml is an accepted tradeoff, document and explicitly validate that behavior instead of describing this solely as a CRT fix.

aruniverse (commented):
Adversarial review found two additional non-inline items:

- `iModelCore/libsrc/VCPKG.md:45-48` still says tools live under `$OutRoot/.../downloads/tools`, are repeatedly downloaded, and disappear on clean/TMR builds. That now directly contradicts this PR's persistent `VCPKG_ROOT/downloads/<arch-platform>` layout; please update the setup/troubleshooting guidance.
- The normal Windows CI jobs are production builds, so they do not exercise the new Windows `DEBUG` archive-selection/link path. Before merge, please record clean Windows DEBUG link validation for both MSVC and `WINDOWS_CLANG` (release archive present, no `debug/lib`, final PugiXml link succeeds).

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: The whole reason I stopped having these in the out directory ("install root") is because it was producing a bunch of needless extra work on every build. This output now goes to a single location for all builds by a given user, but I added a lock during install both here and in `vcpkg_run_install.sh` so that multiple jobs running on the same computer will wait for each other.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: See my response to the previous comment.

tcobbs-bentley: The user override procedure has now been documented in VCPKG.md.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
tcobbs-bentley: This was intentional, based on the amount of effort needed to support imodel-native's extremely unusual usage of release CRT during debug builds. Supporting this in any vcpkg-built library typically requires a great deal of extra effort for the marginal benefit of having a subset of debug functionality present in those third-party libraries.

tcobbs-bentley: The comments in the Windows cmake files have been updated.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Addressed for the original/default-cache case by 7b4e210f2c. The batch wrapper now holds `%VCPKG_CACHE_BASE%\%VCPKG_PLATFORM_KEY%.install.lock` across `vcpkg install` (lines 239-260), and the shell wrapper does the same at lines 188-203. Same-platform default-cache runs therefore serialize while different architectures retain separate caches. I found separate edge cases involving overridden cache paths and the macOS fallback lock and am commenting on those independently.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: The core concern is addressed by 7b4e210f2c/f794f280f0: defaults moved out of `VCPKG_ROOT` into a per-user cache, the wrappers honor `VCPKG_DOWNLOADS`, `X_VCPKG_REGISTRIES_CACHE`, and `VCPKG_DEFAULT_BINARY_CACHE`, and `VCPKG.md` documents those overrides. The latest fallback implementation still has a Windows expansion/writability edge case, so I am posting that as a focused follow-up rather than keeping this original concern open wholesale.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
aruniverse: Addressed as an explicit policy decision by 60d4f8fd3f: the triplet comments now acknowledge optimization, `NDEBUG`, and the loss of pugixml assertions in Windows DEBUG builds, and explain why release-only output was accepted. I found one overly absolute feasibility claim in the new rationale and am commenting on that separately. The previously requested clean Windows DEBUG link validation is also still not recorded.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +165,40 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+        --triplet "$TRIPLET" \
+        --downloads-root="$DOWNLOADS_ROOT" \
+        --x-install-root="$INSTALL_ROOT" \
+        --x-manifest-root="$MANIFEST_DIR" \
+        --x-buildtrees-root="$INSTALL_ROOT/buildtrees" \
+        --x-packages-root="$INSTALL_ROOT/packages" \
+        "${OVERLAY_ARGS[@]}"
+}
+
+VCPKG_LOCKFILE="$VCPKG_CACHE_BASE/$VCPKG_PLATFORM_KEY.install.lock"
```
aruniverse: The lock identity does not follow the documented cache overrides. For example, if `VCPKG_DOWNLOADS=/shared/vcpkg` is used for both `arm64-android` and `x64-android`, both processes mutate the same downloads/tools and derived registries directories but acquire different platform-keyed lock files. A common `X_VCPKG_REGISTRIES_CACHE` has the same problem, and different users sharing an override also have unrelated per-user locks. This recreates the registry fetch/GC race despite the new lock. Please lock the actual mutable downloads and registry directories (in deterministic order), rather than deriving synchronization only from the triplet and default cache base. The batch wrapper has the same issue at line 239.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +165,40 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+if command -v flock >/dev/null 2>&1; then
+    exec 9>"$VCPKG_LOCKFILE"
+    if ! flock 9; then
+        echo "Error: failed to acquire vcpkg install lock $VCPKG_LOCKFILE"
+        exit 1
+    fi
+    run_vcpkg_install
+else
+    VCPKG_LOCKDIR="$VCPKG_LOCKFILE.d"
+    while ! mkdir "$VCPKG_LOCKDIR" 2>/dev/null; do
```
aruniverse: Stock macOS has no `flock`, so it always takes this `mkdir` fallback. `EXIT`/`INT`/`TERM` traps cover normal termination, but SIGKILL, a crash, or power loss leaves the directory permanently. The unbounded loop then wedges every later same-platform install; a pre-created stale lock directory reproduces it. Please use an auto-releasing advisory lock on macOS too (for example a small `fcntl.flock` helper), or fail explicitly when no safe backend exists rather than using an unrecoverable directory lock.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -137,26 +137,52 @@ if not exist "%VCPKG_EXE%" (
...
+rem If the profile location is unavailable (LOCALAPPDATA unset or not creatable, e.g. a
+rem locked-down agent), fall back to a directory under INSTALL_ROOT so the build still works.
+rem --------------------------------------------------------------------------------------
+set "VCPKG_CACHE_BASE=%LOCALAPPDATA%\Bentley\vcpkg"
+if "%LOCALAPPDATA%"=="" set "VCPKG_CACHE_BASE=%INSTALL_ROOT%\vcpkg-cache"
+if not exist "%VCPKG_CACHE_BASE%" mkdir "%VCPKG_CACHE_BASE%" 2>nul
+if not exist "%VCPKG_CACHE_BASE%" (
+    echo vcpkg: persistent cache base "%VCPKG_CACHE_BASE%" unavailable; falling back to INSTALL_ROOT
+    set "VCPKG_CACHE_BASE=%INSTALL_ROOT%\vcpkg-cache"
+    if not exist "%VCPKG_CACHE_BASE%" mkdir "%VCPKG_CACHE_BASE%"
```
aruniverse: Because this is inside a parenthesized block, `%VCPKG_CACHE_BASE%` is expanded before line 162 changes it. This `mkdir` therefore retries the failed profile path instead of creating `%INSTALL_ROOT%\vcpkg-cache`. Default child-cache creation often masks the bug, but with cache variables overridden the fallback base can remain absent and the lock-open loop waits forever. Move fallback creation outside the block or use delayed expansion, verify it is writable, and report a bounded lock-open failure instead of treating every open error as contention.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
```diff
@@ -10,3 +10,22 @@
...
+# ('MDd_DynamicDebug' vs 'MD_DynamicRelease') and _ITERATOR_DEBUG_LEVEL (2 vs 0) link
+# mismatches against our release-compiled BePugiXml wrapper object.
+#
+# Accepted tradeoff: the Release CMake config also adds optimization and defines NDEBUG, which
+# strips pugixml's internal assert(...) checks. Before the vcpkg migration bmake compiled
+# pugixml.cpp with asserts active in DEBUG, so this drops those library diagnostics -- but only
+# on Windows DEBUG. The non-Windows triplets still build a debug archive with asserts enabled,
+# and pugixml's asserts guard internal invariants (tree/allocation/xpath consistency) rather
+# than acting as a correctness mechanism, so the lost coverage is minor and platform-local.
+# Reproducing a debug archive with /MD + _ITERATOR_DEBUG_LEVEL=0 but without NDEBUG is not
```
aruniverse: The accepted release-only policy is clear, but `not reachable through triplet variables` is too absolute. At the pinned baseline, `vcpkg_cmake_configure` appends `VCPKG_CMAKE_CONFIGURE_OPTIONS_DEBUG`, and pugixml honors cache-configurable `CMAKE_MSVC_RUNTIME_LIBRARY`; a triplet can therefore select `MultiThreadedDLL` for Debug (with an explicit iterator-level flag) without necessarily replacing the chainload toolchain. Please keep the policy decision but phrase this as a maintenance/complexity tradeoff rather than a technical impossibility.

aruniverse (commented):
Follow-up review: the original default-cache concurrency issue is addressed, but the new locking/override paths have the inline issues below. Two prior non-inline items also remain open:

- `iModelCore/libsrc/VCPKG.md:7` and the Setup note around line 47 still say caches/downloaded tools live under the vcpkg source tree or `$OutRoot`, are repeatedly downloaded, and disappear on clean/TMR builds. The new section at lines 185-221 contradicts those older paragraphs. The PR description likewise still says the caches are under the vcpkg source root. Please update all three descriptions.
- The now-green Windows checks are ordinary production builds; there is still no recorded clean Windows DEBUG PugiXml link for MSVC and `WINDOWS_CLANG`. Please record the targeted DEBUG validation requested in the prior review.

iModelCore/libsrc/vcpkg_run_install.sh
aruniverse: Partially addressed by 7cd5daec78. Both wrappers now place locks inside the actual downloads and registry directories; the shell implementation also canonicalizes, de-duplicates, and sorts those paths. The Windows implementation still acquires them in downloads-then-registry order and does not resolve junction aliases, leaving livelock cases; I posted a focused follow-up on the new code.

iModelCore/libsrc/vcpkg_run_install.sh
aruniverse: The reported stale-directory-lock problem is addressed by 7cd5daec78. Stock macOS now uses Perl `flock()`, Linux uses `flock(1)`, and the script fails explicitly when neither auto-releasing backend exists. I found a separate child-lifetime hole in the Perl wrapper and posted it on the new implementation.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Still open at c281ffdb5b. The fallback block at lines 160-164 is unchanged, so `%VCPKG_CACHE_BASE%` is still expanded before line 162 updates it. Directory-creation errors remain suppressed and lock-open errors still enter an unbounded retry. Moving lock files into the selected mutable directories removes one earlier consequence, but does not address the fallback/writability failure itself.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
aruniverse: Still open at c281ffdb5b: lines 27-30 are unchanged and continue to describe the alternative as technically unreachable rather than an intentionally rejected maintenance tradeoff.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +166,89 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+    perl -e '
+        my @locks;
+        while (@ARGV && $ARGV[0] ne "--") { push @locks, shift @ARGV }
+        shift @ARGV if @ARGV;
+        my @held;
+        for my $f (@locks) {
+            open(my $fh, ">>", $f) or die "vcpkg lock: cannot open $f: $!\n";
+            flock($fh, 2)          or die "vcpkg lock: cannot lock $f: $!\n";
+            push @held, $fh;
+        }
```
aruniverse, line 313: The Perl process owns the lock descriptors, but `system(@ARGV)` launches vcpkg as a separate child. If the Perl wrapper is killed or crashes while vcpkg remains alive, the kernel releases the locks and another install can enter the same caches while the first child is still mutating them. I reproduced this by killing Perl: the first fake-vcpkg child stayed alive and a second invocation immediately acquired the lock. Please make the locked process `exec @ARGV` after clearing close-on-exec on the held descriptors, or forward termination and guarantee the child has exited before releasing the locks.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -192,20 +219,75 @@ if exist "%OVERLAY_PORTS%" set OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports="%OVERL
...
+rem --------------------------------------------------------------------------------------
+for %%I in ("%VCPKG_DOWNLOADS%") do set "LOCK_DIR_1=%%~fI"
+for %%I in ("%X_VCPKG_REGISTRIES_CACHE%") do set "LOCK_DIR_2=%%~fI"
+if /i "%LOCK_DIR_1%"=="%LOCK_DIR_2%" set "LOCK_DIR_2="
+set "LOCK_FILE_1=%LOCK_DIR_1%\.vcpkg-install.lock"
+if defined LOCK_DIR_2 set "LOCK_FILE_2=%LOCK_DIR_2%\.vcpkg-install.lock"
+
+:acquire_lock
+set "GOT_LOCK="
+if defined LOCK_DIR_2 (
```
aruniverse: The two Windows locks are still acquired in caller-provided downloads-then-registry order. Two valid override configurations with reversed paths (A uses X then Y; B uses Y then X) can each acquire its first handle, fail the second, release, sleep for the same fixed interval, and livelock indefinitely. `%~fI` also does not resolve junction/symlink aliases, so two spellings of one directory can make a single process repeatedly contend with its own first handle. Please resolve final directory identities, de-duplicate them, and acquire in one deterministic case-insensitive order, matching the shell implementation; retain a bounded retry/error path as a backstop.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +166,89 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+# downloads tree, and locking the same file twice from one process would self-deadlock. Sorting
+# gives every process the same global acquire order, preventing AB/BA deadlock between runs whose
+# directory sets overlap.
+_canon_dirs=()
+for _d in "$DOWNLOADS_ROOT" "$X_VCPKG_REGISTRIES_CACHE"; do
+    _canon="$(cd "$_d" 2>/dev/null && pwd -P)" || _canon="$_d"
+    _canon_dirs+=("$_canon")
+done
+_oldifs="$IFS"; IFS=$'\n'
+LOCK_DIRS=($(printf '%s\n' "${_canon_dirs[@]}" | LC_ALL=C sort -u))
```
aruniverse: Changing `IFS` preserves spaces, but this unquoted command-substitution array assignment still performs pathname expansion. A literal cache directory such as `cache[x]` can be replaced by sibling `cachex`, causing the process to lock the wrong path; broad matches can create extra entries. Since there are only two canonical directories, compare/order the two quoted strings directly, or populate the array with a quoted `while IFS= read -r` loop so glob characters remain literal.

aruniverse (commented):
Follow-up review of `c281ffdb5b`:

- `c281ffdb5b` correctly updates the stale Setup note in `VCPKG.md`, but `VCPKG.md:7` and the PR description still say caches live under the vcpkg source tree. Please update those remaining descriptions.
- The requested clean Windows DEBUG PugiXml link validation for MSVC and `WINDOWS_CLANG` is still not recorded. The latest platform checks are currently pending and, in any case, the standard Windows jobs exercise production configuration rather than this DEBUG path.
- The existing Windows fallback/writability and pugixml feasibility threads remain open; I replied on those threads rather than duplicating comments.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: I never claimed that this was fixed.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
tcobbs-bentley: I never claimed that this was fixed.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: This should now be fixed.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
tcobbs-bentley: Using information from the review comment, this has now been updated to do a debug build with release CRT instead of release-only.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: This should now be fixed. (Now uses `exec` instead of `system` and clears the close-on-exec flag on the fd.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.sh
aruniverse: Addressed by 9dffba4a11. The Perl backend now clears `FD_CLOEXEC` on each held lock descriptor and `exec`s vcpkg in place, so the install process itself owns the locks for its full lifetime and its native exit status propagates. A killed outer wrapper can no longer release the locks while vcpkg continues.

iModelCore/libsrc/vcpkg_run_install.sh
aruniverse: Addressed by fd15fa4aa5. The unquoted command-substitution assignment was replaced with a quoted `while IFS= read -r` loop over the sorted entries, preserving spaces and literal glob characters without pathname expansion.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
aruniverse: Addressed substantively by 2af3f4376b. The release-only/impossibility rationale was removed: both Windows triplets now build Debug and Release, set the Debug target runtime to `MultiThreadedDLL`, and explicitly force `_ITERATOR_DEBUG_LEVEL=0`; `pugixml.mke` again selects `debug/lib/pugixml.lib` for DEBUG builds.

iModelCore/libsrc/pugixml/triplets/x64-windows-static.cmake
aruniverse: The diagnostics regression is now addressed in code by 2af3f4376b. Windows DEBUG builds once again produce and link the Debug pugixml archive, preserving assertions and unoptimized codegen, while the triplets override its runtime to `/MD` and `_ITERATOR_DEBUG_LEVEL=0` for compatibility with BentleyBuild.

aruniverse (commented):
The latest implementation fixes several prior findings. One new documentation inconsistency remains inline below.

.github/skills/vcpkg/SKILL.md
```diff
@@ -44,6 +44,13 @@ Under `iModelCore/libsrc/<mylib>/`:
 - `vcpkg-configuration.json` — copy from an existing consumer (e.g. `compress/`); update `baseline` if needed
 - `triplets/` — platform-specific triplet files if the defaults in `iModelCore/libsrc/` are not sufficient (see `compress/triplets/` for examples)
 
+> **Check whether the library needs release-only Windows triplets.** Some libraries fail to
```
aruniverse: The latest code no longer uses release-only pugixml triplets: `2af3f4376b` builds a Debug archive with `/MD` and `_ITERATOR_DEBUG_LEVEL=0`, and `pugixml.mke` links `debug/lib/pugixml.lib`. This guidance still says pugixml is a release-only reference, and the later sections at lines 155-174 and 223-233 still instruct agents to force `VCPKG_BUILD_TYPE release` and always link the release archive. Please document both supported strategies, use pugixml as the Debug-with-release-CRT example, and retain crashpad as the release-only example.

tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: Fixed using a new ps1 script to do the work that couldn't be done directly in the bat file.

tcobbs-bentley:
> * The requested clean Windows DEBUG PugiXml link validation for MSVC and `WINDOWS_CLANG` is still not recorded. The latest platform checks are currently pending and, in any case, the standard Windows jobs exercise production configuration rather than this DEBUG path.

I successfully performed all four Windows build combinations using the currently committed and pushed tree: MSVC release, MSVC debug, clang release, clang debug.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
```diff
@@ -0,0 +1,57 @@
...
+
+function Resolve-Final([string] $path) {
+    if ([string]::IsNullOrEmpty($path)) { return }
+    try { $path = [System.IO.Path]::GetFullPath($path) } catch { }
+
+    $FILE_SHARE_ALL = 7
+    $OPEN_EXISTING = 3
+    $FILE_FLAG_BACKUP_SEMANTICS = 0x02000000   # required to open a directory handle
+    $handle = [VcpkgLock.Native]::CreateFileW($path, 0, $FILE_SHARE_ALL, [System.IntPtr]::Zero, $OPEN_EXISTING, $FILE_FLAG_BACKUP_SEMANTICS, [System.IntPtr]::Zero)
+    if ($handle.ToInt64() -eq -1) { return $path }   # e.g. missing dir; the .bat re-checks existence
```
aruniverse: This fails open to the unresolved spelling when `CreateFileW` fails; line 43 does the same for `GetFinalPathNameByHandleW`, and `GetFullPath` errors are also suppressed. That restores the alias/order problem this helper is intended to eliminate—for example, valid restricted SMB paths can be openable/traversable while final-name normalization fails with access denied. Two aliases can then survive as different strings and self-contend or acquire in different orders. Please fail closed with the Win32 error (and have the batch wrapper check PowerShell's exit status) instead of returning the raw path; use a stable file identity if restricted shares must remain supported.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -192,20 +228,124 @@ if exist "%OVERLAY_PORTS%" set OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports="%OVERL
...
+rem de-dup and make a run self-deadlock (open one file on two handles -- the second open always
+rem fails and wedges the retry loop). vcpkg_resolve_lock_dirs.ps1 uses GetFinalPathNameByHandle to
+rem collapse those aliases; Sort-Object -Unique then de-duplicates case-insensitively and imposes a
+rem single deterministic order. Ordering matters even when the two dirs are distinct: two runs whose
+rem override paths list the same pair in opposite order would otherwise each grab their first handle,
+rem fail the second, release, sleep the same interval, and livelock forever. With one global order
+rem both runs contend on the same first handle, so one strictly wins. This mirrors the shell
+rem wrapper's sort -u.
+rem --------------------------------------------------------------------------------------
+set "LOCK_LIST=%VCPKG_CACHE_BASE%\.lockdirs-%RANDOM%%RANDOM%.txt"
```
aruniverse: This shared temporary filename is not atomically unique. Two concurrent wrappers can choose the same `%RANDOM%%RANDOM%` value, then delete/truncate/read each other's output; because one line is accepted as a valid de-duplicated result, a process can proceed while locking only one of two directories. Resolver exit status is also not checked before parsing. Prefer consuming validated stdout directly with a completion sentinel, or use an atomically created GUID/per-process file and require successful resolver completion plus exactly one or two records.

aruniverse (commented):
Follow-up review found the new Windows resolver substantially improves ordering, but the prior lock and skill threads are not fully addressed, so I did not reply to them. New issues are inline below.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
```diff
@@ -0,0 +1,57 @@
...
+    }
+    finally {
+        [void][VcpkgLock.Native]::CloseHandle($handle)
+    }
+}
+
+@($env:VCPKG_DOWNLOADS, $env:X_VCPKG_REGISTRIES_CACHE) |
+    ForEach-Object { Resolve-Final $_ } |
+    Where-Object { $_ } |
+    Sort-Object -Unique
```
aruniverse: `Sort-Object` uses the current user's culture by default. Since these override directories are explicitly supported across users, two users with different cultures can order the same Unicode path pair differently and reintroduce reversed acquisition. Sort and de-duplicate with one explicit `StringComparer.OrdinalIgnoreCase` comparer so the order is process- and locale-independent.

.github/skills/vcpkg/SKILL.md
```diff
@@ -171,6 +201,71 @@ may not run at all.
...
+set(VCPKG_C_FLAGS_DEBUG   "${VCPKG_C_FLAGS_DEBUG} -D_ITERATOR_DEBUG_LEVEL=0")
+```
+
+The Debug config keeps `NDEBUG` undefined, so `assert()`s survive; only the CRT/iterator level
+change to match our build. The consumer `.mke` keeps the normal `%if defined (DEBUG)` →
+`$(vcpkgDbgLibDir)` selection, since a `debug/lib/` archive is still produced. This works only if
+the library's CMake honors the cache-set `CMAKE_MSVC_RUNTIME_LIBRARY` (CMP0091) — most modern
+ports do — and it is more machinery: apply it to **every** Windows triplet and validate a real
+Windows DEBUG link on **both** the MSVC and clang-cl toolsets. `pugixml/triplets/x64-windows-static.cmake`
+documents this recipe in its header comment (pugixml itself ships Fix A as an accepted tradeoff).
```
aruniverse: This sentence contradicts both the preceding Fix B recipe and the current code: pugixml now ships Fix B, producing and linking a Debug archive with the release CRT; crashpad is the Fix A/release-only example. Please replace the parenthetical and the broad references below with explicit mappings: pugixml's Windows triplets + `pugixml.mke` implement Fix B, while crashpad's Windows triplets + `client.mke` implement Fix A.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
tcobbs-bentley: Fixed. Failures now trigger build failure.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: Fixed: temp file is no longer used, so no possible overlap.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
tcobbs-bentley: Fixed: now uses locale-independent ordering.

.github/skills/vcpkg/SKILL.md
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
aruniverse: Addressed by fb0178cf71. `Resolve-Final` now throws on `GetFullPath`, `CreateFileW`, or `GetFinalPathNameByHandleW` failure; the helper buffers all results, exits nonzero on error, and emits its completion sentinel only after every directory resolves. The batch wrapper rejects output without that sentinel, so unresolved aliases can no longer be accepted as lock identities.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Addressed by fb0178cf71. The shared `%RANDOM%` result file was removed; batch consumes resolver stdout directly and accepts it only when the final completion sentinel arrives with exactly one or two directory records.

iModelCore/libsrc/vcpkg_resolve_lock_dirs.ps1
aruniverse: Addressed by fb0178cf71. The helper now uses one `SortedSet[string]` with `StringComparer.OrdinalIgnoreCase`, providing locale-independent ordering and case-insensitive de-duplication across users.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: The original alias/reversed-order defect is addressed by the current resolver: it resolves final directory names through handles, de-duplicates with `OrdinalIgnoreCase`, emits one deterministic order, and batch acquires in that order with bounded retries. Separate transport and lock-file-lifecycle edge cases found in this follow-up are commented on the latest code.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -192,20 +229,174 @@ if exist "%OVERLAY_PORTS%" set OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports="%OVERL
...
+rem wrappers could otherwise pick the same %RANDOM% name and truncate/read each other's output,
+rem and a lone surviving line would look like a valid one-directory result while the second
+rem directory silently went unlocked. The resolver prints each canonical lock dir followed by a
+rem completion sentinel; we accept the result only when that sentinel arrives, which proves the
+rem resolver ran to the end (its stderr passes through to the console on failure).
+set "LOCK_DIR_1="
+set "LOCK_DIR_2="
+set "LOCK_OK="
+set "LOCK_COUNT=0"
+for /f "usebackq delims=" %%I in (`powershell -NoProfile -ExecutionPolicy Bypass -File "%~dp0vcpkg_resolve_lock_dirs.ps1"`) do (
```
aruniverse: Canonical paths cross a Windows PowerShell-to-`for /f` console pipe without establishing a Unicode-safe encoding. A `%LOCALAPPDATA%` or override path containing characters outside the active OEM code page can arrive as replacement characters while the ASCII completion sentinel still succeeds, so a valid directory is rejected or the wrong spelling is used. Please make producer and consumer use a matched UTF-8 transport (restoring the prior code page), encode records losslessly, or move lock opening into PowerShell so canonical paths never cross `for /f`. Add a non-ASCII user/cache-path test.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -192,20 +229,174 @@ if exist "%OVERLAY_PORTS%" set OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports="%OVERL
...
+    exit /b 1
+)
+
+rem Existence does not prove writability. A pre-existing cache dir (notably a VCPKG_DOWNLOADS /
+rem X_VCPKG_REGISTRIES_CACHE override, or a locked-down agent's profile) whose ACLs deny creating
+rem the lock file makes the redirection open below fail exactly like genuine contention, burning
+rem the full ~1h retry budget. Preflight each dir with a uniquely named probe file: a sharing
+rem violation is name-specific and transient, but an ACL denial fails for every name, so a probe
+rem that cannot be created after several distinct names is a permission problem -- report it now.
+call :probe_writable "%LOCK_DIR_1%"
```
aruniverse: `CALL` performs a second percent-expansion pass. Since `%` is legal in Windows directory names, a canonical path such as `C:\cache\%TEMP%\downloads` is re-expanded when passed as this argument, making the probe target a different path. Pass the directory through a dedicated environment variable and call the subroutine without a path argument; likewise avoid embedding the resolver path in nested command text if the checkout path can contain `%...%`.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -192,20 +229,174 @@ if exist "%OVERLAY_PORTS%" set OVERLAY_ARG=%OVERLAY_ARG% --overlay-ports="%OVERL
...
+rem an error instead of looping forever. ~5s per attempt; 720 attempts tolerates ~1h of genuine
+rem contention from a concurrent heavy install (e.g. crashpad) before giving up.
+set "LOCK_ATTEMPTS=0"
+set "LOCK_MAX_ATTEMPTS=720"
+
+:acquire_lock
+set "GOT_LOCK="
+if defined LOCK_DIR_2 (
+    (
+        9>>"%LOCK_FILE_1%" (
```
aruniverse: The random probe proves only that a new file can be created in the directory; it does not prove that the existing `.vcpkg-install.lock` can be opened for append. In a shared cache that file can be read-only or carry another user's ACL, causing every real open here to be suppressed and misclassified as contention for roughly an hour. Preflight the actual lock file with Win32/.NET append access and retry only sharing/lock violations; fail immediately on access denied, read-only, path, or other permanent errors.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +166,114 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+for _d in "$DOWNLOADS_ROOT" "$X_VCPKG_REGISTRIES_CACHE"; do
+    _canon="$(cd "$_d" 2>/dev/null && pwd -P)" || _canon="$_d"
+    _canon_dirs+=("$_canon")
+done
+
+LOCK_FILES=()
+_lock_keys=()
+for _d in "${_canon_dirs[@]}"; do
+    _f="$_d/.vcpkg-install.lock"
+    : >> "$_f"
```
aruniverse: The script creates/stats each lock file to choose a global inode order, then later reopens the pathname in the flock/Perl backend. If a writable shared cache cleans or replaces the file between stat and reopen, the acquired inode differs from the ordering key: overlapping runs can acquire opposite first locks, or one process can retain the old inode while another locks the replacement and enters the same cache concurrently. Open each lock once, derive its key with `fstat`, and retain/lock that same descriptor through exec; also detect or prevent replacement.

aruniverse (commented):
The latest commits fully address several prior Windows resolver findings. New path-transport and lock-file-lifecycle issues are inline below.

iModelCore/libsrc/vcpkg_run_install.bat
```diff
@@ -137,26 +137,63 @@ if not exist "%VCPKG_EXE%" (
...
-rem Isolate the git registries cache per install-root to prevent a race condition
-rem where parallel vcpkg processes (building different arches) collide on the shared
-rem global cache at %LOCALAPPDATA%\vcpkg\registries. Concurrent git fetch/GC
-rem operations on that bare repo cause transient "port does not exist" failures.
-set "X_VCPKG_REGISTRIES_CACHE=%INSTALL_ROOT%\registries"
-if not exist "%X_VCPKG_REGISTRIES_CACHE%" mkdir "%X_VCPKG_REGISTRIES_CACHE%"
-
-rem Use a persistent local binary cache by default to avoid rebuilding heavy ports
-rem (for example, crashpad) across builds. Allow callers to override.
+rem --------------------------------------------------------------------------------------
```
tcobbs-bentley: The interaction between this bat file and the ps1 script keeps getting more complicated, and much of the complication would be gone if the bat file were simply gone, so I'm working on replacing `vcpkg_run_install.bat` and `vcpkg_resolve_lock_dirs.ps1` with `vcpkg_run_install.ps1`.

tcobbs-bentley: The bat is now gone, completely replaced with a ps1 file.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: Fixing these problems kept increasing the complexity, so the .bat/.ps combination has been replaced with a single .ps1 file that is called from the mke files.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: Fixing these problems kept increasing the complexity, so the .bat/.ps combination has been replaced with a single .ps1 file that is called from the mke files.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: This should now be fixed.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: Fixing these problems kept increasing the complexity, so the .bat/.ps combination has been replaced with a single .ps1 file that is called from the mke files.

iModelCore/libsrc/vcpkg_run_install.bat
tcobbs-bentley: This is gone now, along with the bat file.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Addressed structurally by 038984d991. The batch/`for /f` bridge was removed; each Windows install caller now invokes `vcpkg_run_install.ps1` directly with separately quoted arguments, and canonical paths stay as .NET strings/wide Win32 paths throughout.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Addressed structurally by 038984d991. `vcpkg_run_install.bat`, `CALL`, and nested cmd reparsing are gone; the `.mke` files pass the script and three path/triplet arguments directly to PowerShell, so literal `%...%` path segments no longer receive a second batch expansion.

iModelCore/libsrc/vcpkg_run_install.bat
aruniverse: Addressed by the PowerShell rewrite. The wrapper now opens the actual `.vcpkg-install.lock` with `.NET File.Open`, retries only Win32 sharing/lock violations 32/33, and fails immediately for access-denied, read-only, path, or other permanent I/O errors.

iModelCore/libsrc/vcpkg_run_install.sh
aruniverse: The original stat/reopen race is addressed by 0267c9f10f/038984d991: the POSIX wrapper opens each directory once, derives device/inode identity from that retained descriptor, orders and locks those same descriptors, and verifies the pathname still resolves to the expected identity after acquisition. A separate post-validation retarget/replacement edge case found in this follow-up is commented on the latest code.

iModelCore/libsrc/vcpkg_run_install.ps1
```diff
@@ -0,0 +1,248 @@
...
+                    '--x-manifest-root', $ManifestDir,
+                    '--x-buildtrees-root', [System.IO.Path]::Combine($InstallRoot, 'buildtrees'),
+                    '--x-packages-root', [System.IO.Path]::Combine($InstallRoot, 'packages'),
+                    "--overlay-triplets=$overlayTriplets"
+                )
+                if ([System.IO.Directory]::Exists($overlayPorts)) {
+                    $arguments += "--overlay-ports=$overlayPorts"
+                }
+
+                & $vcpkgExe @arguments
```
aruniverse: The lock `FileStream`s are owned by `powershell.exe`, while this invocation starts `vcpkg.exe` as a separate native child with non-inheritable handles. If PowerShell is forcibly terminated or crashes while vcpkg survives, Windows closes the locks and another wrapper can enter the same caches while the first child is still mutating them—the same lifetime hole previously fixed in the Perl backend. Launch vcpkg with inheritance-confirmed lock handles, or place it in a kill-on-close Job Object before it can run so wrapper death either preserves the locks in vcpkg or terminates vcpkg before releasing them.

iModelCore/libsrc/vcpkg_run_install.ps1
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    if (-not $env:VCPKG_DEFAULT_BINARY_CACHE) {
+        $env:VCPKG_DEFAULT_BINARY_CACHE = [System.IO.Path]::Combine($cacheBase, 'archives')
+    }
+    if (-not (Ensure-Directory $env:VCPKG_DEFAULT_BINARY_CACHE)) {
+        throw "vcpkg binary cache '$($env:VCPKG_DEFAULT_BINARY_CACHE)' could not be created"
+    }
+    if (-not $env:VCPKG_BINARY_SOURCES) {
+        $env:VCPKG_BINARY_SOURCES = "clear;files,$($env:VCPKG_DEFAULT_BINARY_CACHE),readwrite"
```
aruniverse: This synthesizes a vcpkg binary-source expression from an unescaped filesystem path. Commas and semicolons are legal in Windows paths but are provider/source delimiters, so a cache such as `C:\cache,shared` is parsed as extra fields rather than the intended directory. When `VCPKG_BINARY_SOURCES` is unset, leave it unset and let vcpkg's default `default,readwrite` provider consume `VCPKG_DEFAULT_BINARY_CACHE` directly, or escape backticks/commas/semicolons according to vcpkg's source grammar.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +166,148 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+            : # BSD stat (macOS)
+        else
+            _path_key=""
+        fi
+        if [ "$_path_key" != "${_lock_keys[$_i]}" ]; then
+            echo "Error: vcpkg install lock directory was replaced while acquiring it: $_d" >&2
+            exit 1
+        fi
+    done
+    "${VCPKG_CMD[@]}"
```
aruniverse, line 281: The retained descriptor fixes the stat/reopen race, but vcpkg still receives the original `DOWNLOADS_ROOT` pathname. A configured symlink can be retargeted after this one-time identity check: the wrapper keeps the old directory locked while vcpkg follows the original alias into the replacement, whose lock can be acquired concurrently by another process. Directory rename/replacement has the same split-domain problem. Pass the canonicalized paths to vcpkg and keep lock objects in a stable namespace outside the replaceable cache directory (or otherwise make cache access descriptor-relative); validate symlink-retarget and directory-replacement cases.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -144,11 +166,148 @@ if [ -n "${CRASHPAD_USE_LLD:-}" ]; then
...
+    _tmp="${LOCK_FDS[0]}"; LOCK_FDS[0]="${LOCK_FDS[1]}"; LOCK_FDS[1]="$_tmp"
+    _tmp="${_lock_keys[0]}"; _lock_keys[0]="${_lock_keys[1]}"; _lock_keys[1]="$_tmp"
+fi
+
+if command -v flock >/dev/null 2>&1; then
+    # Lock the descriptors opened above; the kernel drops them on exit.
+    for _i in "${!LOCK_DIRS[@]}"; do
+        _d="${LOCK_DIRS[$_i]}"
+        _fd="${LOCK_FDS[$_i]}"
+        if ! flock "$_fd"; then
```
aruniverse: Exclusive `flock` on a read-only directory descriptor is not portable to NFS-backed Linux caches: Linux's NFS flock emulation requires the exclusive-lock target to be opened for writing, while a directory cannot be opened writable. An enterprise `$HOME/.cache` or explicit cache override on NFS can therefore fail before installation. Use a retained `O_RDWR` regular lock object in a stable lock namespace, or explicitly detect/reject and document remote filesystems if they are unsupported.

aruniverse (commented):
Replacing batch with a native PowerShell wrapper resolves the prior path-transport findings. The rewrite introduces several new lock-lifetime, filesystem, and cache-expression issues noted inline.

iModelCore/libsrc/crashpad/readme.md
```diff
@@ -22,7 +22,7 @@ when updating:
 
 - **Windows** (all triplets — both MSVC and clang): built from the **local vcpkg overlay
   port** at `iModelCore/libsrc/crashpad/ports/crashpad/`. The overlay is wired in by the
-  `--overlay-ports` handling in `iModelCore/libsrc/vcpkg_run_install.bat`, which passes any
+  `--overlay-ports` handling in `iModelCore/libsrc/vcpkg_run_install.ps1`, which passes any
```
aruniverse: The wrapper name is updated here, but the non-Windows guidance immediately below still says `vcpkg_run_install.sh` does not pass `--overlay-ports` and that Linux/macOS/Android use the upstream registry port. The shell wrapper currently adds `--overlay-ports` whenever `ports/` exists, and `crashpad/ports/crashpad/portfile.cmake` states that the local overlay is used on all platforms. Please update the source/version maintenance instructions so they point to the overlay actually built outside Windows too.

aruniverse (commented):
One additional documentation issue in the wrapper migration:

iModelCore/libsrc/vcpkg_run_install.ps1
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.ps1
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: Fixed.

iModelCore/libsrc/crashpad/readme.md
tcobbs-bentley: Fixed (here and in comment in cmake file)

aruniverse (approved):

PR #1537, opened 2026-08-05
Enable SQLite STAT4 query statistics

khanaffan:
## What it does

Turns on SQLite's STAT4 query-planner statistics (`-DSQLITE_ENABLE_STAT4=1` in `SqliteLib.mke`, `SqliteShell.mke`, `ITwinDaemon.mke`, and `bentley-sqlite.c`) so the planner can use per-index sample histograms instead of just the coarse `sqlite_stat1` averages.

## Why STAT4

`sqlite_stat1` stores only simple aggregate counts (rows per index, average rows per distinct key), so it is sensitive to data volume and goes stale as soon as row counts shift. `sqlite_stat4` instead captures the *data distribution* — sampled key prefixes with their frequencies — which stays representative even as data grows or changes, as long as the shape of the data is stable. In practice that means STAT4 needs to be recomputed (`ANALYZE`) far less often than STAT1, while giving the planner much better estimates for skewed data and range/equality constraints.

## Supporting changes

- **Db creation** (`BeSQLite.cpp`): after the initial `ANALYZE`, `sqlite_stat4` is now truncated alongside `sqlite_stat1`, so new dbs don't ship stats derived from an empty file.
- **Profile upgrade** (`BeSQLiteProfileManager.cpp`): the 3102 upgrader clears `sqlite_stat4` too.
- **Removed workarounds**: bloom-filter disabling on new dbs, and the `DisqualifyTypeIndexForBisCoreExternalSourceAspect` hack in `DgnDb::_OnDbOpened` (plus its test) — both were band-aids for bad plans that STAT4 addresses.

## Tests

- `sqlite_stat1` → `sqlite_stat1_and_stat4`: asserts `sqlite_stat4` now exists, is populated by `ANALYZE`, but is **not** captured by change tracking (upstream SQLite only fires the preupdate hook for `sqlite_stat1`). Added a `GetStatRowCountForTable` helper since the stat tables already contain rows for built-in `be_*` tables.
- Updated hard-coded db/serialized sizes (36864 → 40960, etc.) for the extra `sqlite_stat4` page.
- `FileFormatCompatibilityTests`: excludes `sqlite_stat%` from DDL comparison (build-dependent, not file-format).
- `MockHubApi`: updated the channel sqlite-schema SHA, which now differs from the briefcase one because the frozen seed asset predates STAT4.

## Pending Items
- [ ] Pending ECPresentation Performance Test Results
- [ ] STAT4 is not captured in changeset. Need to contact SQLite Dev. 
- [ ] Port back to imodel02


khanaffan:

PR #1538, opened 2026-08-07
Fix crash on null PropertyMap during ECSQL SELECT preparation

khanaffan:
Fixes iTwin/itwinjs-backlog#2311 (Sentry `IMODELJSNODEADDON-28M`, 648 events, "High", ongoing ~1 year).

## Problem

`ECSqlBinderFactory::CreateBinder` dereferenced `PropertyNameExp::GetPropertyMap()` with no null check when creating an Id binder:

```cpp
else
    return CreateIdBinder(ctx, *propNameExp.GetPropertyMap(), sysPropInfo, paramNameGen);
```

`GetPropertyMap()` can legitimately return `nullptr` — an exp referring to a **subquery or CTE alias** may be backed by an arbitrary expression (e.g. a `NULL` literal in one branch of a compound select) rather than a mapped property. The guards inside `GetPropertyMap()` were `BeAssert`s, which expand to `((void)0)` under `NDEBUG`, so shipping builds ran straight into the dereference.

### Why the fault address is `0x20`

`CreateIdBinder` builds `ECSqlTypeInfo(propMap)`, whose inline ctor calls `propertyMap.GetProperty()` → reads `PropertyMap::m_ecProperty`. In the x64 layout of `PropertyMap : RefCountedBase, ISupportsPropertyMapVisitor`:

| Offset | Member |
|---|---|
| `0x00` | vptr (`IRefCounted`) |
| `0x08` | `m_refCount` + pad |
| `0x10` | vptr (`ISupportsPropertyMapVisitor`) |
| `0x18` | `m_type` + pad |
| **`0x20`** | **`m_ecProperty`** |

With `this == nullptr` that reads `[0x00 + 0x20]` — exactly the reported `EXCEPTION_ACCESS_VIOLATION_READ / 0x20`, inlined into frame `00` (`CreateIdBinder`).

### How an Id-typed alias ends up without a property map

The branch is entered on `sysPropInfo.IsId() || propNameExp.GetTypeInfo().IsId()`. For a property ref, `m_sysPropInfo` stays `NoSystemProperty`, so it is the type-info arm that fires: `GetTypeInfoFromPropertyRef()` infers a compound-select alias's type by **borrowing it from a sibling `UNION` branch** when its own branch is `Null`/`Unset`. So the exp is typed as an `Id` while its own branch (`SELECT NULL AS eid`) has no backing property map.

> Note: the original issue attributed this to a failed `dynamic_cast` / use-after-free. That was a misread — the `VCRUNTIME140!__RTDynamicCast` frames sat mid-stack (indices 11–12), i.e. stack-scan artifacts from the missing Windows symbols. There is no `dynamic_cast` on the release-build path; `PropertyMap::GetAs<>` only uses one inside a `BeAssert`. The issue has been updated with the corrected analysis.

## Reproduction

```sql
SELECT * FROM (
    SELECT NULL AS eid FROM meta.ECClassDef
    UNION ALL
    SELECT ECInstanceId AS eid FROM meta.ECClassDef
) WHERE eid = ?
```

Before the fix this **crashed the test process** (SIGSEGV on macOS ARM64 debug, log truncated mid-fixture with no gtest summary), preceded by:

```
PropertyNameExp.cpp(561): ASSERTION FAILURE: propertyMap != nullptr && "...sub query ref..."
PropertyNameExp.cpp(588): ASSERTION FAILURE: propertyMap != nullptr && "PropertyNameExp's PropertyMap should never be nullptr."
```

## Changes

| File | Change |
|---|---|
| `ECSqlBinder.cpp` | Null-check `GetPropertyMap()`; fall back to `CreateIdBinderForQuery`. Also braced the outer `if`, which had a **dangling `else`** attached to a nested `if`. |
| `PropertyNameExp.cpp` | Removed the 3 `BeAssert`s claiming the property map is never null (they were factually wrong and would fail the new tests in debug); documented `nullptr` as a supported result; **kept** a targeted assert for the `ClassName` case where non-null genuinely is guaranteed, so real mapping bugs still get caught. |
| `ECSqlPreparer.cpp` | `RemovePropertyRefs`: `continue` on null (would have faulted at `+0x18` on `m_type`). |
| `ECSqlSelectPreparer.cpp` | `ExtractPropertyRefs`: skip null — **this frame appears in the Sentry stack**. |
| `ECSqlPropertyNameExpPreparer.cpp` | `Prepare`: assert + return `ECSqlStatus::Error` on null. |

`CreateIdBinderForQuery` is the correct fallback: this branch is only reachable in `SELECT`/`DELETE`/non-assignment `UPDATE` scopes (so `isNoop = false` is right), and `GetTypeInfo()` is guaranteed to be an Id type — that is the branch's entry condition.

All four new guards are no-ops when the pointer is non-null, so behaviour on every existing path is unchanged.

## Tests

New `CommonTableExpNullPropertyMapTestFixture` in `CommonTableExpTests.cpp` — one `TEST_F` per shape, since a native crash takes the whole process down and would otherwise mask the rest:

- `BindIdParam_SubqueryUnionNullAlias` — subquery + `UNION`
- `BindIdParam_CteWithoutColumnList` — CTE with no column list
- `BindIdParam_CteWithColumnList_Control` — **control**: `WITH cte(eid) AS ...` already took the guarded path and passed before the fix
- `BindIdParam_SubqueryUnionReversedBranchOrder` — exercises the type-resolution loop rather than branch ordering
- `BindIdParam_SubqueryUnionRelationshipEndId` — `TargetECInstanceId` variant

## Verification

Full ECDb suite, macOS ARM64 debug static:

```
[==========] 1645 tests from 85 test suites ran.
[  PASSED  ] 1644 tests.
```

All 5 new tests pass with **zero assertion failures** — the null property map is now produced and handled cleanly.

The single failure is `ConcurrentQueryFixture.DeadlockReproduction_MainStepVsWorkerPrepare`, which is **pre-existing and unrelated**:
- It fails intermittently (~1 in 8 runs) **with this fix applied**, and passes in isolation — a race, not a deterministic break.
- The assert is `rapidjson document.h:1523` → `RemoveAllMembers()` from `RowRender::ClearAndGetCachedJsonDocument` / `Reset` (`InstanceReaderImpl.cpp:167/177`) — unsynchronized shared `m_cachedJsonDoc` state on the `SELECT $` instance-reader path, which that test hammers from the main thread alongside 40 concurrent worker queries.
- Nothing here touches concurrency, `InstanceReader`, or JSON. Every query in that test resolves via the `Exp::Type::ClassName` path, where `GetPropertyMap()` is non-null, so all new guards are no-ops and the code path is identical to before.

Worth tracking separately as a latent thread-safety bug in `RowRender`.

## Regression history

- **#384** (`1e4c8592`, 2023-08-03) introduced the branch; `CreateIdBinder` took a `PropertyMap const*`.
- **#454** (`4d17925d`, 2023-09-15) changed it to a reference and introduced the unchecked `*GetPropertyMap()` — **the regression point**.
- **#1243** (`8d910710`, 2025-10-23) narrowed the CTE guard to "with columns", routing more CTE shapes into the unchecked `else`.

## Follow-ups (not in this PR)

- **Windows symbols are not uploaded to Sentry** — only Linux ELF DIFs are (`linux_process_all_symbols.py` → `sentry-cli upload-dif -t elf`); Windows PDBs go only to the Azure DevOps symbol server, which Sentry cannot read. This is why the crash never auto-symbolicated, and the resulting stack-scan unwind produced the misleading `dynamic_cast` frames. Recommend a matching `sentry-cli upload-dif -t pdb -t pe` step.
- The `RowRender` cached-document race noted above.

### DIFF
--- a/iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
+++ b/iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
@@ -4311,11 +4311,9 @@ TEST_F(CommonTableExpNullPropertyMapTestFixture, BindIdParam_SubqueryUnionNullAl
     )";
 
     ECSqlStatement stmt;
-    const ECSqlStatus prepareStatus = stmt.Prepare(m_ecdb, query); // must not crash
-    if (prepareStatus.IsSuccess()) {
-        ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
-        ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
-    }
+    ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, query)) << query;
+    ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
+    ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
 }
 
 //---------------------------------------------------------------------------------------
--- a/iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
+++ b/iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
@@ -4333,11 +4333,9 @@ TEST_F(CommonTableExpNullPropertyMapTestFixture, BindIdParam_CteWithoutColumnLis
     )";
 
     ECSqlStatement stmt;
-    const ECSqlStatus prepareStatus = stmt.Prepare(m_ecdb, query); // must not crash
-    if (prepareStatus.IsSuccess()) {
-        ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
-        ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
-    }
+    ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, query)) << query;
+    ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
+    ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
 }
 
 //---------------------------------------------------------------------------------------
@@ -4360,11 +4358,9 @@ TEST_F(CommonTableExpNullPropertyMapTestFixture, BindIdParam_CteWithColumnList_C
     )";
 
     ECSqlStatement stmt;
-    const ECSqlStatus prepareStatus = stmt.Prepare(m_ecdb, query); // must not crash
-    if (prepareStatus.IsSuccess()) {
-        ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
-        ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
-    }
+    ASSERT_EQ(ECSqlStatus::Success, stmt.Prepare(m_ecdb, query)) << query;
+    ASSERT_EQ(ECSqlStatus::Success, stmt.BindId(1, ECInstanceId((uint64_t) 1))) << query;
+    ASSERT_NE(BE_SQLITE_ERROR, stmt.Step()) << query;
 }
 
 //---------------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


rschili (approved):

rschili (approved):

iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
```diff
@@ -4268,4 +4268,166 @@ TEST_F(CommonTableExpTestFixture, Values_OnlyNull)
...
+    }
+}
+
+//---------------------------------------------------------------------------------------
+// Query 5: navigation / relationship end Id variant.
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+TEST_F(CommonTableExpNullPropertyMapTestFixture, BindIdParam_SubqueryUnionRelationshipEndId) {
+    ASSERT_EQ(BentleyStatus::SUCCESS, SetupECDb("issue2311_rel_end_id.ecdb", SchemaItem(R"xml(<?xml version='1.0' encoding='utf-8'?>
+        <ECSchema schemaName='TestSchema' alias='ts' version='1.0.0' xmlns='http://www.bentley.com/schemas/Bentley.ECXML.3.1'>
```
rschili: This test uses EC 3.1, new tests should always use 3.2 schemas

aruniverse:
@mergifyio backport release/5.12.x

iModelCore/ECDb/Tests/NonPublished/CommonTableExpTests.cpp
khanaffan: fixed

PR #1539, opened 2026-08-08
Schema Sync v2 "upstream" - update/upgrade split and update skips over extra sync db changes

rschili:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9604

For PR description please see the linked itwinjs-core PR.

Don't be discouraged by the size of this PR.

About 60% of this is tests - 3.6k lines across a new SchemaSyncImportTests.cpp and the MockHubApi harness. Of the ~2.4k product lines, half is SchemaSync.cpp itself and a fifth is one new file, DerivedDbStructures.cpp. The remaining 30 files average under 60 lines. (measured at time of writing, may be stale)

I split the tests into 2 suites, only ~30% of the schema sync tests run during a normal build now. To run the extra tests use:

`Product/ECDb-Gtest/ECDbTest.exe --gtest_filter=*ExtendedTests.*`

I will look into enabling those during our CI builds, just once on the fastest platform.


iModelCore/BeSQLite/BeSQLite.cpp
```diff
@@ -2669,6 +2669,22 @@ void DisableBloomFilter(SqlDbP db) {
...
+SuppressForeignKeyActions::SuppressForeignKeyActions(DbCR db) : m_db(db) {
+    m_db.GetStatementCache().Empty();
+    sqlite3_test_control(SQLITE_TESTCTRL_FK_NO_ACTION, m_db.GetSqlDb(), 1);
+}
+
+/*---------------------------------------------------------------------------------**//**
+* @bsimethod
++---------------+---------------+---------------+---------------+---------------+------*/
+SuppressForeignKeyActions::~SuppressForeignKeyActions() {
+    sqlite3_test_control(SQLITE_TESTCTRL_FK_NO_ACTION, m_db.GetSqlDb(), 0);
```
RohitPtnkr1996: I'm wondering if it is worth keeping track of the previous state instead of just a RAII switch?
True MirrorTables might never nest, but this is in the public api.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -2039,6 +2112,28 @@ BentleyStatus MainSchemaManager::CreateOrUpdateIndexesInDb(SchemaImportContext&
...
+                // There are no physical indexes to compare against here, so compare with what
+                // ec_Index already holds. Re-persisting an unchanged index would give it a new id on
+                // every import - PersistIndexDef always takes a fresh one - and the file would stop
+                // matching the briefcases that adopt from it.
+                if (IsIndexPersistedUnchanged(index))
+                    continue;
+
+                if (!ctx.IsSemanticRebasing())
+                    {
+                    if (BE_SQLITE_OK != m_ecdb.ExecuteSql(SqlPrintfString("DELETE FROM main." TABLE_Index " WHERE Name = '%s'", index.GetName().c_str())))
```
RohitPtnkr1996: Can we use a binder for this like in IsIndexPersistedUnchanged above ?

iModelCore/BeSQLite/BeSQLite.cpp
rschili: Valid concern, some relevant info:

There are two ways the connection flag is set:

- SuppressForeignKeyActions through SQLITE_TESTCTRL_FK_NO_ACTION in BeSQLite.cpp:2675.
- ChangeStream::ApplyChanges(... SetFkNoAction(true)) through SQLITE_CHANGESETAPPLY_FKNOACTION in ChangeSet.cpp:900,918.

SQLite’s changeset implementation reads and restores the actual bit in db->flags at sqlite3.c:239099 and 239287. The test-control interface only sets or clears it and provides no getter.

SQLite explicitly permits executing SQL from a changeset conflict callback. Such a callback can construct SuppressForeignKeyActions while the changeset apply owns the flag. A wrapper-local bool or nesting counter would miss that state and its destructor could clear the flag before the apply finishes.

The way we use this, it should not be a problem, but yes it's a poor API. We could request Sqlite make the flag gettable, that would improve the API, however, that would still not protect us against overlapping guards that remember different previous states. I lean towards keeping it as is. But I'm improving the documentation to state this.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
rschili: Good catch! Fixed, also fixed more occurrences across the entire file.

khanaffan (changes requested):
test schema-sync-off schema changeset application against legacy files; compare resulting FK/trigger/index DDL across authoring and receiving briefcases; and test interrupted/resumed multi-changeset pulls involving overflow-table creation.

We need ensure that when schema sync is off we have not broken any existing contract/behavior.

We need to verify changes not gated to schema sync is backward compatible. May try a an existing iModels with existing changesets and apply them from seed to tip. ECPresentation testing as very old iModels. The gated stuff can be fixed later but un-gated need to be tested well.

iModelCore/BeSQLite/BeSQLite.cpp
khanaffan: I do not think you should in any way use non-production api of sqlite. `sqlite3_test_control` is a testing api not meant for production. We previously actually used this api to suppress bloom filter which did not work great for us in ECPresentation. But it's always good to document you use case and send them and see what they think. Sometime they will catch edge case that we did not thought about our workflow.

iModelCore/ECDb/ECDb/SchemaSync.cpp
```diff
@@ -996,90 +897,1159 @@ SchemaSync::Status SchemaSync::UpdateDataVersion(SyncDbUri const& syncDbUri) {
...
+        m_conn.GetImpl().Issues().ReportV(
+            IssueSeverity::Error, IssueCategory::SchemaSync, IssueType::ECDbIssue, ECDbIssueId::ECDb_0630,
+            "Unable to attach sync db '%s' as '%s' to primary connection: %s",
+            effectiveSyncDbUri.GetUri().c_str(), SchemaSyncHelper::ALIAS_SYNC_DB, m_conn.GetLastError().c_str());
+        return Status::ERROR_UNABLE_TO_ATTACH;
+    }
+
+    const auto cleanup = [&](Status status) {
+        SchemaSyncUpstreamHelper::DropClosure(m_conn);
+        m_conn.DetachDb(SchemaSyncHelper::ALIAS_SYNC_DB);
```
RohitPtnkr1996: If detach fails for some reason, sync db will stay attached to the primary connection and still report a success.

rschili:
> test schema-sync-off schema changeset application against legacy files; compare resulting FK/trigger/index DDL across authoring and receiving briefcases; and test interrupted/resumed multi-changeset pulls involving overflow-table creation.
> 
> We need ensure that when schema sync is off we have not broken any existing contract/behavior.
> 
> We need to verify changes not gated to schema sync is backward compatible. May try a an existing iModels with existing changesets and apply them from seed to tip. ECPresentation testing as very old iModels. The gated stuff can be fixed later but un-gated need to be tested well.

@khanaffan good idea, I'll see if I can run against those presentation datasets.

iModelCore/BeSQLite/BeSQLite.cpp
RohitPtnkr1996: It seems SuppressForeignKeyActions was not actually working.
I kept getting a sqlite assertion when running the tests in debug.
It seems FK action triggers are compiled lazily and cached on the schema objects (FKey.apTrigger[]), not in prepared statements.
So just clearing the statement cache was not enough.
I was about to commit a test which would've failed in release, but the recent changes to QueryForeignKeyEnforcement have resolved it.

Marking this comment as resolved.

iModelCore/ECDb/ECDb/DbMapValidator.h
```diff
@@ -45,6 +45,8 @@ struct DbMapValidator final
         BentleyStatus ValidateOverflowPropertyMaps(ClassMap const& classMap) const;
         BentleyStatus CheckDuplicateDataPropertyMap() const;
         ECDbCR GetECDb() const { return m_schemaImportContext.GetECDb(); }
+        // True, if the database does not maintain the physical data layer, which we do e.g. for a schema sync db.
+        bool DbHasDataTables() const { return m_schemaImportContext.MaintainsDataTables(); }
```
soham-bentley: Is this method the single source of truth for checking if a db is a briefcase or a sync db? @rschili 
If so maybe we can name it better like `isSchemaSyncDb`?

README.md
```diff
@@ -78,6 +78,119 @@ imodel-native/
...
+| `BuildArchitecture` | `LinuxX64`, `Winx64`, `MacOSARM64`, the Android and iOS variants - `bb -h` lists them all |
+| `BB_PRIMARY_REPO` | `imodel-native-internal` |
+
+### Commands
+
+```sh
+bb pull                          # update every repository in the strategy
+bb build                         # incremental build of everything the strategy names
+bb build --tmrbuild --noprompt   # delete the whole output tree first, then build
+bb rebuild ECObjects:*test*      # force a rebuild of matching parts only
```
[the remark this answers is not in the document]
rschili: The order in the example is actually correct.

README.md
```diff
@@ -78,6 +78,119 @@ imodel-native/
...
+```
+
+`--tmr` deletes the output tree without building; `--tmrbuild` deletes and then builds. `--noprompt` skips the confirmation, which is what makes either usable from a script.
+
+Narrow a build to one product with `-p` (product), `-s` (strategy) and `-f` (part file, without the `.PartFile.xml` suffix):
+
+```sh
+bb -p ECDb-Gtest -s iModelCore -f iModelCore/ECDb/ECDb build
+bb -p Units-Gtest -s iModelCore -f iModelCore/Units/Units build
```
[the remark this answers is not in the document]
rschili: The commands work on my machine, they are correct.

iModelCore/ECDb/ECDb/SchemaSync.cpp
rschili: This was somehow marked resolved by you? But it's correct, Copilot further down just made almost the same remark, I'm fixing it.

iModelCore/ECDb/ECDb/SchemaSync.cpp
```diff
@@ -996,93 +915,1369 @@ SchemaSync::Status SchemaSync::UpdateDataVersion(SyncDbUri const& syncDbUri) {
...
+        return status;
```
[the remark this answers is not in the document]
rschili: - Cleanup now checks and logs DropClosure and DetachDb failures.
     - The success path propagates cleanup failure and no longer raises OnAfterSchemaChanges afterward.

iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
```diff
@@ -391,12 +386,21 @@ DbResult DgnDb::_AfterSchemaChangeSetApplied() const {
...
 }
 
+//--------------------------------------------------------------------------------------
+// Only half of the answer: whether this briefcase sits at the tip is an iModelHub fact, so the
+// caller has to have pulled first.
+// @bsimethod
+//--------------------------------------------------------------------------------------
+bool DgnDb::_IsLevelWithTimeline() {
+    return !IsBriefcase() || !Txns().HasPendingTxns();
+}
```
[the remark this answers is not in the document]
rschili: - _IsLevelWithTimeline() now uses Txns().HasLocalChanges(), covering saved and unsaved changes.

iModelCore/ECDb/ECDb/SchemaSync.cpp
RohitPtnkr1996: Ah yeah, I must've resolved it by mistake.

iModelCore/iModelPlatform/DgnCore/DgnDb.cpp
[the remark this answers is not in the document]
rschili: I had to revert this change it turns out this comes with a cost that is disproportional to the gain.

- initialization and repair reject pre-existing unsaved changes in their TypeScript preflight;
- initialization must stage its container property before native initialization because cloud access is discovered from that property;
- native sees operation-owned transient state at that point, so HasLocalChanges() is too broad;
- the existing implementation still rejects committed, unpushed transactions.

RohitPtnkr1996:
Carried out regression tests against 6 legacy iModels from iTwin Hub.
These iModels were of the sizes S, M, L and XL and spanned the years 2019-2024.

### 1. Legacy changeset replay (schema sync off)

- **Dataset:** 6 legacy iModels (base checkpoint + complete ordered changeset history
  exported from the hub; ordering and ids validated against the SHA1 changeset-id chain
  during replay).
- **Method:** Each changeset was replayed seed to tip without setting up schema sync,
  once on this branch and once on main.
- **Recorded state artifact after every changeset containing:**
  - `PRAGMA integrity_check`
  - `PRAGMA foreign_key_check`
  - Normalized `sqlite_master` DDL (all tables including FK clauses, triggers, indexes)
  - EC metadata/mapping fingerprints (`ec_Schema`/`Class`/`Property`/`Table`/`Column`/`PropertyMap`, per-table column maps)
  - Content fingerprints of `ec_cache_ClassHierarchy`/`ec_cache_ClassHasTables`
  - Trigger inventory
  - Row counts of every table
- **Result:** All artifacts are byte-identical between main and this branch for every
  iModel, at every changeset index. Integrity and FK checks are clean throughout.

### 2. Interrupted/resumed multi-changeset pulls

- **Dataset:** Same as above.
- **Method:** Closing and fully reopening the briefcase after every changeset, then
  comparing the tip state against a single-session replay.
- **Result:** Byte-identical tips in all cases, on both builds, and identical across builds.

khanaffan (approved):

khanaffan:
Just a note, the PR include changes to how we create tiggers and  fk constraints. it changes when they get created along with many little refactor and fixes that could have been in separate PR more like prep work for schema sync. SchemaSync PR should had only schema sync changes. Clubbing prep work with actual changes makes this PR more complex to review.

rschili:
> Just a note, the PR include changes to how we create tiggers and fk constraints. it changes when they get created along with many little refactor and fixes that could have been in separate PR more like prep work for schema sync. SchemaSync PR should had only schema sync changes. Clubbing prep work with actual changes makes this PR more complex to review.

Yes I agree, at first it seemed like only a few changes, but it got more and more over the course of the story and I probably should have separated this into two at some point.

iModelCore/ECDb/ECDb/DerivedDbStructures.cpp
```diff
@@ -0,0 +1,418 @@
...
+            BeAssert(false && "Link table map must have source and target ECInstanceId property maps");
+            return ERROR;
+            }
+
+        std::set<DbTable const*> constraintTables = GetConstraintPrimaryTables(manager, *end.constraint);
+        if (constraintTables.size() != 1)
+            continue; // the mapper refuses more than one; nothing to reference if there is none
+
+        DbTable const* constraintTable = *constraintTables.begin();
+        if (constraintTable->GetType() == DbTable::Type::Virtual || constraintTable->GetType() == DbTable::Type::Existing)
```
aruniverse, line 190: **P1 — Link tables lose foreign keys when an endpoint maps to `ExistingTable`.** The previous mapper still created the source/target FK when the referenced endpoint used an existing physical table; this condition now skips both `Virtual` and `Existing`. Newly materialized link tables can therefore lose endpoint enforcement and cascade deletion behavior. Please skip only virtual endpoint tables and add a `PRAGMA foreign_key_list` regression test for a link relationship targeting `ExistingTable`.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -1941,6 +2020,15 @@ BentleyStatus MainSchemaManager::CreateOrUpdateIndexesInDb(SchemaImportContext&
     if (LoadIndexesSQL(sqliteIndexes) != SUCCESS)
```
aruniverse: **P1 — Removing index metadata can leave the physical SQLite index active.** `LoadIndexesSQL()` discovers physical indexes by joining `sqlite_master` to current `ec_Index` rows. Once an `ec_Index` row is deleted, its physical index is absent from this reconciliation set, so it is never dropped. A removed unique index can continue rejecting valid data and different briefcases can retain different DDL. Please inventory managed indexes directly from `sqlite_master` and drop those no longer represented by `ec_Index`, excluding SQLite autoindexes and unmanaged tables.

iModelCore/ECDb/PublicAPI/ECDb/SchemaManager.h
```diff
@@ -20,6 +20,24 @@ enum class SchemaChangeType {
...
 using SchemaChangeEvent = BeEvent<ECDbCR,SchemaChangeType>;
 
+//=======================================================================================
+//! Schema import options. Not needed by regular callers. They are specific to certain
+//! exceptional workflows and therefore only used by them.
+//! Reachable as SchemaManager::SchemaImportOptions; it lives out here because SchemaSync, which
+//! is declared first, takes it.
+//! @ingroup ECDbGroup
+//+===============+===============+===============+===============+===============+======
+enum class SchemaImportOptions
```
aruniverse, line 30: **P1 if binary compatibility is required — moving this enum changes exported C++ symbols.** Existing exports previously used the nominal type `SchemaManager::SchemaImportOptions`; moving it to `BeSQLite::EC::SchemaImportOptions` and adding a source alias does not preserve the old mangled linker symbols. Previously compiled consumers can fail to load/link. Please retain the original nested exported type or provide ABI-compatible legacy overloads that convert internally.

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/TxnManager.h
```diff
@@ -703,16 +703,17 @@ struct TxnManager : BeSQLite::ChangeTracker {
 
     //! Reverse all changes back to a previously saved TxnId.
     //! @param[in] txnId a TxnId obtained from a previous call to GetCurrentTxnId.
-    //! @param[in] allowCrossSessions if No, don't reverse any Txns older than the beginning of the current session.
+    //! @param[in] allowCrossSessions also reverse Txns older than the beginning of the current session. A schema
+    //! Txn starts a new session, so this is the only way to back one out. Every Txn in the table is local and unpushed.
     //! @return DgnDbStatus::Success if the transactions were reversed, error status otherwise.
     //! @see  GetCurrentTxnId CancelTo
-    DGNPLATFORM_EXPORT DgnDbStatus ReverseTo(TxnId txnId);
+    DGNPLATFORM_EXPORT DgnDbStatus ReverseTo(TxnId txnId, bool allowCrossSessions = false);
```
aruniverse, line 710: **P1 if binary compatibility is required — a default argument does not preserve the old `ReverseTo`/`CancelTo` ABI.** The base exported one-argument functions; replacing them with two-argument functions removes those linker symbols because defaults are applied only at the caller. Please keep one-argument exported overloads forwarding to the new two-argument implementations. The same applies to `CancelTo` below.

aruniverse (commented):
Non-blocking review of the linked SchemaSync changes. These findings are submitted as a COMMENT review, not a request for changes.

**P1 — `upgradeSchemaStrings` discards its custom schema context** (`iModelJsNodeAddon/IModelJsNative.cpp:2507-2519`). `ImportXmlSchemas` reads `schemaLockHeld` and `schemaSyncDbUri`, but never reads `ecSchemaXmlContext`, although the TypeScript API accepts and forwards it and the file-based native import supports it. Serialized upgrades whose references exist only in that context therefore fail. Please assign the supplied context to the native import options and add a referenced-schema regression test. This hunk was not available as an inline anchor, so the finding is included in the review body.

tcobbs-bentley:
@rschili The new failures are due to the removal of jsoncpp. I wasn't aware that you were writing new jsoncpp code in this PR. For sure `MockHubApi.cpp` uses jsoncpp, and possibly other code does also. When imodel-native main was merged in removing jsoncpp, that code broke. My PR stack that removed jsoncpp converted the jsoncpp usage to RapidJson. The last PR in that stack is [here](https://github.com/iTwin/imodel-native/pull/1553). Contact me if you run into problems with the conversion, but my time zone is 2 hours behind EDT.

soham-bentley (approved):

iModelCore/ECDb/ECDb/DerivedDbStructures.cpp
rschili: ExistingTable is misleading, it means a table already exists, instead it means a table that is not managed by ECDb's data model. For those we do not manage foreign keys. So this is a false positive.

iModelCore/ECDb/PublicAPI/ECDb/SchemaManager.h
rschili: We make no guarantee for nagive binary compatibility. The only consumer is the N-API which we build together with this code.

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/TxnManager.h
rschili: Same as above, we do not make guarantees for binary compatibility, it is acceptable to break it in our environment.

iModelCore/ECDb/ECDb/DbMapValidator.h
rschili: Yeah valid concern. I named that `MaintainsDataTables()` in other spots, and changed this one to match that name now.
I did think about calling is "IsSchemaSyncDb" but I lean towards keeping it abstract.

rschili:
> **P1 — `upgradeSchemaStrings` discards its custom schema context** (`iModelJsNodeAddon/IModelJsNative.cpp:2507-2519`). `ImportXmlSchemas` reads `schemaLockHeld` and `schemaSyncDbUri`, but never reads `ecSchemaXmlContext`, although the TypeScript API accepts and forwards it and the file-based native import supports it. Serialized upgrades whose references exist only in that context therefore fail. Please assign the supplied context to the native import options and add a referenced-schema regression test. This hunk was not available as an inline anchor, so the finding is included in the review body.

NativeDgnDb::ImportXmlSchemas now validates and forwards ecSchemaXmlContext.
Added a regression test in iModelJsNodeAddon/api_package/ts/src/test/DgnDb.test.ts using a reference available only through the supplied context.

I am honestly not 100% sure if this is a good idea. Custom context is a weird mechanism that I'd like to drop at some point. Passing it along to two databases can have weird side effects, I think I left it out on purpose, but realized now that both choices are bad :-)

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
rschili: Replaced LoadIndexesSQL with LoadPhysicalIndexes, which inventories explicit indexes directly from sqlite_master.
As current mapped indexes are processed, their names are removed from the inventory. Remaining indexes are dropped.
SQLite autoindexes and indexes on ExistingTable tables are excluded.
Added UpdateDbSchemaDropsIndexRemovedFromMetadata in SchemaSyncTest.cpp.

aruniverse: Addressed at the latest head (`a1a2da88`). `LoadPhysicalIndexes` now inventories explicit managed-table indexes directly from `sqlite_master`; current mapped indexes are removed from that inventory while being processed, and the remaining orphaned indexes are dropped. SQLite autoindexes and indexes on `ExistingTable` tables are excluded. `UpdateDbSchemaDropsIndexRemovedFromMetadata` adds focused regression coverage.

iModelCore/ECDb/ECDb/DbMapValidator.h
soham-bentley: Sounds good

calebmshafer (approved):
Approving the high level readme changes.

tcobbs-bentley:
@Mergifyio backport release/5.13.x

PR #1544, opened 2026-08-12
Fix transformation issue where indirectly copied textures do not use the id remap list

ColinKerr:
Textures were the only element I could find that did not call through the remap path, textures also go through the remap path in other places but not here.


iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnImportContext_Test.cpp
```diff
@@ -573,3 +605,84 @@ TEST_F(ImportTest, OpenStatementsProblem)
...
+    RenderMaterialCPtr sourceMaterial = RenderMaterial::Get(*sourceDb, sourceMaterialId);
+    ASSERT_TRUE(sourceMaterial.IsValid());
+    ASSERT_EQ(sourceTextureId, sourceMaterial->GetRenderingAsset().GetPatternMap().GetTextureId());
+    sourceDb->SaveChanges();
+
+    //  Create the destination Db with an existing texture.
+    //  Texture has a different name to ensure the only way it's found is via the id remapping added to the import context. 
+    DgnDbPtr destDb = openCopyOfDb(L"3dMetricGeneralcc.bim");
+    ASSERT_TRUE(destDb.IsValid());
+    DgnTextureId seededTextureId = insertTexture(*destDb, "PreExistingTexture", L"");
```
aruniverse, line 631: Please force the source and seeded destination texture IDs to differ. Both databases are fresh copies of the same seed and each inserts its first texture, so these IDs can align. In that case the assertion below can still pass if relocation merely leaves the source ID unchanged, without consulting the seeded remap. Insert a throwaway destination texture before `seededTextureId` (or otherwise make the IDs differ), then add `ASSERT_NE(sourceTextureId, seededTextureId)` before importing.

aruniverse (commented):
Thanks for your contribution, AI review found one test-validity issue; details inline.

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnImportContext_Test.cpp
ColinKerr: Done. Dummy texture inserted before seed texture.

aruniverse (approved):

ColinKerr:
@hl662 Nam bot please tell Nam Colin would like him to take a look at this PR

pmconne (approved):
LGTM - thanks for the fix.

hl662 (approved):
> If this is your way of subtly telling me to start ramping up on native and C++ code... its working

PR #1545, opened 2026-08-12
Harden ECSQL against crash inputs

khanaffan:
## Summary

- Harden ECSQL parsing and Node addon lifetime handling against crashes.
- Add regression tests.


iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
```diff
@@ -93,8 +93,7 @@ static sal_Int32    parseString(yyscan_t yyscanner);
 #define YY_EXTRA_TYPE OSQLScanner*
 #define YY_INPUT(buf,result,max_size) \
 { \
-    int c = yyget_extra(yyscanner)->SQLyygetc(); \
-    result = (c == EOF) ? YY_NULL : (buf[0] = c, 1); \
+    result = (int) yyget_extra(yyscanner)->SQLyyread((buf), (max_size)); \
```
aruniverse, line 96: Block reads move `m_nCurrentPos` to the end of the prefetched chunk, but `SQLyyerror` still uses `m_nCurrentPos < m_sStatement.size()` before appending the offending/trailing text. For typical statements smaller than Flex's buffer, the first read sets `m_nCurrentPos` to the statement size even while the lexer is earlier in the input, so parser errors lose their diagnostic context. Can we use `GetScanPos()` in that guard, regenerate `SqlFlex.cpp`, and add an invalid-symbol regression assertion?

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
```diff
@@ -202,10 +202,18 @@ ECSqlTypeInfo PropertyNameExp::GetTypeInfoFromPropertyRef() const {
...
+
     ECSqlTypeInfo resolvedTypeInfo;
     for (auto stmtIdx = 0; stmtIdx < flatList.size(); ++stmtIdx) {
-        BeAssert(flatList[stmtIdx]->GetSelection()->GetChildren().Get<DerivedPropertyExp>(colIdx) != nullptr && "Programmer Error: All the select statements in the compound statement are expected to have the same number of columns in their select clause");
-        auto stmtTypeInfo = flatList[stmtIdx]->GetSelection()->GetChildren().Get<DerivedPropertyExp>(colIdx)->GetExpression()->GetTypeInfo();
+        // The select clauses of a compound statement may have a different number of columns. This is an error which is
+        // only reported later during preparation, so the column index must not be assumed to be valid here.
+        auto const* derivedProp = flatList[stmtIdx]->GetSelection()->GetChildren().Get<DerivedPropertyExp>((size_t) colIdx);
+        if (derivedProp == nullptr || derivedProp->GetExpression() == nullptr)
+            continue;
```
aruniverse, line 214: This safely avoids the out-of-bounds access, but it treats a structurally missing compound-select column as merely absent type information. Compound arity is validated later, and the existing fold-to-NULL path can discard the malformed subtree before that validation. The new `foldedToNullEcsqls` cases consequently ignore `Prepare` and allow invalid `UNION` shapes to succeed. Please validate branch arity before type-driven folding and assert those queries fail, rather than only asserting that they do not crash.

aruniverse (commented):
Adversarial review found four concrete crash/correctness issues; details are inline.

iModelJsNodeAddon/IModelJsNative.cpp
```diff
@@ -5674,7 +5750,14 @@ struct NativeECSqlStatement : BeObjectWrap<NativeECSqlStatement> {
         OPTIONAL_ARGUMENT_BOOL(2,logErrors, true);
         IssueListener listener(*ecdb);
 
+        InvalidateBinders();
```
aruniverse, line 5763: This invalidates binders and unregisters the statement before attempting the new prepare. However, native `ECSqlStatement::Impl::Prepare` returns an error immediately when the statement is already prepared and leaves the original statement prepared. A second `prepare()` therefore removes a still-live statement from `StatementRegistry` and clears its binder lifetime; a later database close can again force-finalize it without `OnDbClose`, and `getBinder()` receives a null lifetime. Please preserve the existing registration/lifetime on the already-prepared failure (or explicitly finalize if re-prepare is intended), and add a double-prepare followed by close/use regression test.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -491,7 +491,14 @@ Napi::Object JsInterop::ConcurrentQueryResetConfig(Napi::Env env, Napi::Object c
 void JsInterop::ConcurrentQueryExecute(ECDbCR ecdb, Napi::Object requestObj, Napi::Function callback) {
     ConcurrentQueryMgr::WithInstance(ecdb, [&](ConcurrentQueryMgr& mgr) -> void {
         BeJsValue beJsReq(requestObj);
-        auto request = QueryRequest::Deserialize(beJsReq);
+        QueryRequest::Ptr request;
+        try {
```
aruniverse: The exception boundary is still too narrow: `ConcurrentQueryMgr::WithInstance` is outside this `try`, and it throws `std::runtime_error` for a closed database. `NativeECDb::ConcurrentQueryExecute` passes `m_ecdb` without first requiring it to be open, so `closeDb(); concurrentQueryExecute(validRequest, cb)` can still let a C++ exception escape through N-API and terminate the process. Please cover the whole native operation (while preserving/rethrowing `Napi::Error`) or validate the open state before `WithInstance`, and add a closed-database regression test.

iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
khanaffan: Fixed. `SQLyyerror` now guards on `GetScanPos()` instead of `m_nCurrentPos`, and `SqlFlex.cpp` was regenerated to match `sqlflex.l`.

Added `InvalidSymbolReportsOffendingText`. It asserts on an invalid-symbol error (`#`), since `SQLyyerror` is only reached from `YY_FATAL_ERROR`, and it strips the echoed ECSQL out of the `ECDb_0471` message first, otherwise the assertion passes trivially.

One case left alone: an offending token that ends exactly at EOF is still not appended, because the guard is a strict `<`. That is pre-existing (the old one-char reads also left the position at the end), and relaxing the guard alone would print a dangling `": "`, since the message loop only assigns the buffer when it hits a space.

iModelCore/ECDb/ECDb/ECSql/PropertyNameExp.cpp
khanaffan: Fixed, and I took the option you suggested: validate arity before the type-driven folding rather than only asserting no crash.

Compound branch arity is now compared in `SelectStatementExp::_FinalizeParsing`, so it is caught while parsing instead of relying on `ValidateSelectClauseItems` during preparation, which the fold-to-NULL path could bypass. `_FinalizeParsing` became two-phase so the comparison runs after the children are finalized, i.e. after asterisks are expanded, and it defers when `GetDeferFinalize()` is set.

The `foldedToNullEcsqls` cases now assert `Prepare` fails. Also added a list of well formed compound selects, including the NULL folding shapes like `SELECT (SELECT b FROM (SELECT NULL a, NULL b))`, asserting they still prepare, so the new check cannot silently over-reject.

iModelJsNodeAddon/IModelJsNative.cpp
khanaffan: Fixed. I went with preserving the existing registration rather than finalizing, since re-prepare is not the intended behaviour here and silently finalizing would change what a double `prepare()` does.

When `m_stmt.IsPrepared()`, the prepare is now attempted without first calling `InvalidateBinders()`/`Unregister()`, so the still-live statement keeps its registry entry and binder lifetime and is still finalized by `OnDbClose`.

Added a JS regression test: prepare, prepare again (fails), use the binders and step, then close the db and check the binder and statement throw instead of crashing.

iModelJsNodeAddon/JsInterop.cpp
khanaffan: Fixed, and I did both things you mentioned rather than picking one.

`NativeECDb::ConcurrentQueryExecute` now uses `GetOpenedDb(info)`, matching the `NativeDgnDb` overload, so a closed db is rejected before `WithInstance`. On top of that the whole native operation is wrapped, rethrowing `Napi::Error` untouched so a real JS exception keeps propagating, and converting other `std::exception`s.

One extra fix while in here: the `ThreadSafeFunction` callback runs later through N-API's C callback, outside that try, so an exception from `JSON.parse` or the response conversion could still terminate the process. That lambda is now guarded too and reports via `ThrowAsJavaScriptException()`.

Added closed-db regression tests for both the ECDb and DgnDb paths.

iModelCore/ECDb/ECDb/ECSql/Parser/SqlFlex.cpp
```diff
@@ -5439,6 +5438,36 @@ sal_Int32 OSQLScanner::SQLyygetc(void)
...
+    struct yyguts_t* yyg = (struct yyguts_t*) yyscanner;
+    if (nullptr == yyg || nullptr == yyg->yy_buffer_stack || nullptr == YY_CURRENT_BUFFER_LVALUE)
+        return m_nCurrentPos;
+
+    // characters delivered to flex that it has not scanned past yet
+    const sal_Int32 unscanned = (sal_Int32) ((YY_CURRENT_BUFFER_LVALUE->yy_ch_buf + yyg->yy_n_chars) - yyg->yy_c_buf_p);
+    return m_nCurrentPos - unscanned;
+    }
 //------------------------------------------------------------------------------
 void OSQLScanner::SQLyyerror(const char *fmt)
```
rschili, line 5461: `SQLyyerror`'s buffer growth corrupts the heap.
The resize loop advances the file-static `Buffer` pointer while copying, so `s` then writes past the
new allocation and the trailing `delete[]` frees an interior pointer. Triggered by any lexer fatal
error followed by 256+ non-space chars, e.g. `"SELECT [" + string(300, 'A')`. Pre-existing, but it's
in the function you're editing and it's the same bug class the PR targets.

iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
rschili: IN_SQLyyerror` and `Buffer` are file-statics, ECSQL parses on
ConcurrentQuery worker threads.** The `OSQLScanner` constructor also resets `IN_SQLyyerror`, so a
scanner built on one thread clears the reentrancy flag for another thread already inside `SQLyyerror`.

iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
```diff
@@ -586,8 +585,8 @@ sal_Int32 gatherNamePre(yyscan_t yyscanner, const sal_Char* text)
             {
             Utf8String sStmt = yyget_extra(yyscanner)->getStatement();
             size_t nLength = strlen(text);
-            size_t nPos = yyget_extra(yyscanner)->GetCurrentPos() - nLength - 2;
-            if (sStmt[nPos] == ':')
+            const sal_Int32 nPos = yyget_extra(yyscanner)->GetScanPos() - (sal_Int32) nLength - 1;
```
rschili, line 588: `SQLyygetc` and `GetCurrentPos()` have no callers left. Should probably be removed.

rschili (approved):
Approved but left 3 comments you should look at first.

iModelCore/ECDb/ECDb/ECSql/Parser/SqlFlex.cpp
khanaffan: Fixed

iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
khanaffan: Fixed

iModelCore/ECDb/ECDb/ECSql/Parser/sqlflex.l
khanaffan: Fixed

aruniverse (approved):

khanaffan:
@pmconne can you review this PR.

pmconne (approved):

PR #1547, opened 2026-08-13
Set up infrastructure so that vcpkg-built libraries will be included in mend scans

tcobbs-bentley:
### TL;DR:

Makes vcpkg library sources visible to the Mend scan, which previously saw nothing because normal
builds restore from the binary cache.

**Note:** this won't actually affect Mend scans until the pipeline that runs them is updated to run the new script.

---

- `vcpkg-mend.json` in each of the 6 consumers lists the triplets whose source graph Mend must see.
- `vcpkg_download_mend_sources.ps1` finds those consumers, downloads their sources into a scan root,
  and verifies every declared and pinned port actually materialized. The Mend pipeline will run it.
- `vcpkg_run_install.{ps1,sh}` gain a `--mend-scan` mode (download-only, no binary cache, no
  compiler tracking) and now refuse to build a consumer whose `vcpkg-mend.json` is missing or empty.
- `vcpkg_validate_mend` part runs the script's checks on Windows builds, so mistakes fail a PR. As a local-only check, this runs very fast.
- `VCPKG.md` and the vcpkg skill document all of the above plus a platform-coverage audit.


iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -6,20 +6,57 @@
...
+fi
+
+# Checked on every build so a consumer that omits it fails here, not in the Mend pipeline.
+MEND_CONFIG="$MANIFEST_DIR/vcpkg-mend.json"
+if [ ! -f "$MEND_CONFIG" ]; then
+    echo "Error: '$MEND_CONFIG' is required so the Mend scan can materialize this library's sources" >&2
+    exit 1
+fi
+# Whitespace-stripped match keeps this dependency-free; the Mend script does the full validation.
+if ! tr -d ' \t\r\n' < "$MEND_CONFIG" | grep -q '"triplets":\["'; then
```
aruniverse: **Medium — metadata validation is only a text heuristic**

This accepts malformed JSON such as `{"triplets":["x64-linux"],"broken":}` because the text pattern still matches. It also differs from the PowerShell path, which accepts a scalar `"triplets":"x64-linux"` via `@(...)`. As a result, non-Windows builds can pass invalid Mend metadata that the Windows validation rejects or interprets differently. Please parse the JSON and explicitly require an array of non-empty strings in both wrappers (with regression cases for malformed JSON, scalar, empty array, and empty entries).

iModelCore/libsrc/vcpkg_run_install.ps1
```diff
@@ -165,12 +168,22 @@ function Invoke-NativeProcessInKillOnCloseJob([string] $exePath, [string[]] $arg
...
 
     $ManifestDir = Normalize-Path $ManifestDir
     $InstallRoot = Normalize-Path $InstallRoot
 
+    # Checked on every build so a consumer that omits it fails here, not in the Mend pipeline.
+    $mendConfigPath = [System.IO.Path]::Combine($ManifestDir, 'vcpkg-mend.json')
+    if (-not [System.IO.File]::Exists($mendConfigPath)) {
+        throw "'$mendConfigPath' is required so the Mend scan can materialize this library's sources"
+    }
+    $mendTriplets = @((Get-Content -LiteralPath $mendConfigPath -Raw | ConvertFrom-Json).triplets)
```
aruniverse: **Medium — `triplets` is not verified to be an array**

`@((ConvertFrom-Json ...).triplets)` turns a scalar value such as `"triplets":"x64-linux"` into a one-item PowerShell array, so this check passes despite the error text requiring a non-empty array. The shell wrapper rejects that shape, making validation platform-dependent. Please check that the parsed property is an actual array and that every element is a non-empty string.

iModelCore/libsrc/vcpkg_run_install.sh
```diff
@@ -210,6 +264,12 @@ VCPKG_CMD=("$VCPKG_EXE" install
     --x-packages-root="$INSTALL_ROOT/packages"
     "${OVERLAY_ARGS[@]}")
 
+if [ "$MEND_SCAN" -eq 1 ]; then
+    VCPKG_CMD+=(--only-downloads --binarysource=clear)
```
aruniverse, line 299: **Medium — documented cache fallback is bypassed for existing read-only caches**

The new Mend path still uses the cache setup above. `mkdir -p "$VCPKG_CACHE_BASE"` succeeds when the directory already exists, even if it is read-only; the later `mkdir -p "$VCPKG_DEFAULT_BINARY_CACHE"` then aborts under `set -e` instead of falling back to `INSTALL_ROOT/vcpkg-cache` as documented at lines 120–121. Please validate writability of an existing cache base (for example by creating a temporary child) before selecting it.

aruniverse:
**Additional low-severity finding (not inline-resolvable):** `iModelCore/libsrc/vcpkg_install_crashpad.mke:6` still says crashpad runs after `vcpkg_install_openssl`, while this PR changes `vcpkg.PartFile.xml` so the chain is `... -> vcpkg_install_curl -> vcpkg_install_crashpad`. Please update the stale comment; it can mislead future chain-order debugging.

aruniverse:
**Additional medium-severity finding:** The shell wrapper claims at `iModelCore/libsrc/vcpkg_run_install.sh:157-158` that shared `VCPKG_DOWNLOADS` / `X_VCPKG_REGISTRIES_CACHE` overrides are serialized, but lock files are created under `$VCPKG_CACHE_BASE/locks` at line 286. If two agents/users point those overrides at the same shared directory while using different `$HOME`/`XDG_CACHE_HOME`, they acquire different lock files and can concurrently mutate the same downloads or registry tree. Please derive the lock-file directory from the canonical shared cache path (or use a shared lock location) when overrides are set, and add a cross-user/shared-cache concurrency test.

tcobbs-bentley:
> **Additional medium-severity finding:** The shell wrapper claims at `iModelCore/libsrc/vcpkg_run_install.sh:157-158` that shared `VCPKG_DOWNLOADS` / `X_VCPKG_REGISTRIES_CACHE` overrides are serialized, but lock files are created under `$VCPKG_CACHE_BASE/locks` at line 286. If two agents/users point those overrides at the same shared directory while using different `$HOME`/`XDG_CACHE_HOME`, they acquire different lock files and can concurrently mutate the same downloads or registry tree. Please derive the lock-file directory from the canonical shared cache path (or use a shared lock location) when overrides are set, and add a cross-user/shared-cache concurrency test.

Unless I'm misunderstanding, this has nothing to do with the changes made in this PR. If this is something that needs to be fixed, let me know, and I can create a separate issue to track it. I don't personally believe that the stated reproduction requirements will ever be met.

tcobbs-bentley:
> **Additional low-severity finding (not inline-resolvable):** `iModelCore/libsrc/vcpkg_install_crashpad.mke:6` still says crashpad runs after `vcpkg_install_openssl`, while this PR changes `vcpkg.PartFile.xml` so the chain is `... -> vcpkg_install_curl -> vcpkg_install_crashpad`. Please update the stale comment; it can mislead future chain-order debugging.

The `vcpkg_install_*` mke files have been updated to simply note that they are part of a sequence found in the part file. Saying which part came before just required extra unnecessary maintenance, and the partfile lists the sequence.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.ps1
tcobbs-bentley: Fixed.

iModelCore/libsrc/vcpkg_run_install.sh
tcobbs-bentley: The behavior that is being called out here is preexisting before this PR, even though the highlighted lines of code happen to be new ones in this PR. The same problem exists in the ps1 script (not mentioned). Once again, if this is something that needs to be fixed, let me know, and I'll create an issue.

aruniverse (approved):

PR #1548, opened 2026-08-14
Resolve benign dgn_Domain conflicts instead of aborting the merge

### LINKED ISSUE #9603: Multi-changeset pullChanges aborts with "PRIMARY KEY INSERT CONFLICT - rejecting this changeset" on dgn_Domain rows; same changesets apply cleanly when pulled individually
### Brief summary

When a single `BriefcaseDb.pullChanges()` call merges multiple changesets, and a later changeset inserts `dgn_Domain` rows that collide with rows auto-created by the schema-upgrade DDL patch, the native conflict handler aborts the apply with `PRIMARY KEY INSERT CONFLICT - rejecting this changeset` (BE_SQLITE_ABORT, errorNumber 4). Pulling the exact same changesets one at a time — or downloading a fresh briefcase at latest — succeeds, resolving the same conflict by Replace.

### Expected vs. actual behavior

Expected: A PK INSERT conflict on `dgn_Domain` rows created by the DDL patch is benign (the incoming row matches the auto-created row) and should be resolved by Replace regardless of how many changesets are being merged in one pull — the outcome should not depend on pull batching.

Actual: The conflict is resolved by Replace only when there are no pending/in-flight txns. During a multi-changeset pull-merge, the earlier changesets already applied within the same merge count as in-flight, so the handler takes the Abort branch and the whole pull fails with:

```
PRIMARY KEY INSERT CONFLICT - rejecting this changeset
```

`errorNumber: 4` (BE_SQLITE_ABORT), not `ChangeSetStatus.ApplyError`.

This makes the failure deterministic and unrecoverable for the cached briefcase: every retry of the pull fails the same way. In OpenSite+ this surfaced in production as repeated import failures (Sentry: 76+ events, 9 users on one issue alone).

### Reproduction steps

The failing sequence involves three changesets produced by an OpenSite+ import using the Civil (Dgn) Connector:

1. cs1 `a33557ce0b207f1a92c9ca097a68d4763e29fa57` — app changeset ("Geolocate iModel")
2. cs2 `30eec208cfece7b69a7a5277704522f4f9d8725f` — connector schema changeset (BisCustomAttributes)
3. cs3 `e5905cb057b8a394270dec64f97bd478e32cf7ad` — connector "Domain schema upgrade": 8 `ec_Schema` rows + 8 matching `dgn_Domain` rows (Functional, DocumentMetadata, …)

Repro matrix (deterministic, verified on both 5.11.2 and 5.12.0):

| Scenario | Result |
|---|---|
| Briefcase at cs1 → single `pullChanges()` spanning cs2+cs3 | ❌ `PRIMARY KEY INSERT CONFLICT - rejecting this changeset` (BE_SQLITE_ABORT) |
| Same briefcase → pull cs2, then cs3 in separate `pullChanges()` calls | ✅ conflict logged as "resolved by replacing the existing row" |
| Fresh `BriefcaseManager.downloadBriefcase` at latest | ✅ succeeds |

Steps:
1. Create an iModel, push an app changeset (cs1), then run the Civil Connector so it pushes its schema changeset (cs2) and a "Domain schema upgrade" changeset (cs3) that inserts `dgn_Domain` rows.
2. Open a briefcase that is at cs1 (i.e. downloaded before the connector ran).
3. Call `pullChanges()` once so it merges cs2 and cs3 in a single merge.
4. The apply of cs3 aborts: the DDL patch applied during the merge auto-creates the `dgn_Domain` rows, cs3's inserts collide, and because cs2 is in-flight within the same merge the conflict handler aborts instead of replacing.

Notes from instrumentation:
- The abort decision is made inside the native addon (`imodeljs.node`); the JS `onChangesetConflict` handler in `core-backend` `IModelDb` is never invoked for this path (verified by patching it with logging — it did not fire).
- `txns.hasPendingTxns` is `false` before the pull — no local user txns are involved; it is purely the in-flight changesets within the same multi-changeset merge that trip the abort branch.

### Minimal reproducible example

The failing hub iModel (Bentley-internal, **production** environment): iTwin `02406360-9717-435b-85df-9b68565255f1`, iModel `468f8e01-1d2c-4a0c-b2d3-e95a29318289` (SurveyData.dgn). A sibling iModel with the same failure: `b9108683-feb5-45ff-9403-cb0d78b66cbd` (SportsField.dgn). A ~40-line Node script that downloads a briefcase pinned at cs1 and calls `pullChanges()` reproduces it deterministically; happy to share it or the briefcase.

### Stack traces and logs

```
[PATCH] Applying DDL patch
PRIMARY KEY INSERT CONFLICT - rejecting this changeset
Error applying changeset ... errorNumber: 4 (BE_SQLITE_ABORT)
```

The conflicting rows are the 8 `dgn_Domain` inserts (Functional, DocumentMetadata, etc.) in cs3, which duplicate rows the DDL patch just auto-created.

### Environment

iTwin.js Version: 5.11.2 and 5.12.0 (both reproduce)
Runtime (e.g., Node.js 20, Electron 28): Node.js 22 (also Electron in production)
Operating System: macOS (repro) / Windows (production reports)
Browser (if applicable): N/A (backend)

### Priority / urgency

High for OpenSite+: this deterministically breaks file import for any cached briefcase that is behind a connector Domain-schema-upgrade changeset (dozens of production Sentry events across multiple users). We have shipped an app-side workaround (delete the poisoned local briefcase and re-download fresh), but that discards the briefcase and re-downloads the full iModel, so a proper fix in the native conflict handling (treat in-flight changesets of the same merge like the no-pending-txns case for this benign conflict) is preferred.

### AI acknowledgment

- [x] I used AI to generate this issue report.
- [x] (If the above is checked) I have reviewed the AI-generated content before submitting.

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9624

Fixes [#9603](https://github.com/iTwin/itwinjs-core/issues/9603)

A `dgn_Domain` row is created automatically by `DgnDomains::SyncWithSchemas()` as soon as a domain's schema is imported, so merging a changeset that registers a domain conflicts with the row that was just created locally. Both rows describe the same domain, so the conflict is benign and the incoming row is authoritative.

The conflict was only resolved by `Replace` when `HasPendingTxns()` was false, which made the outcome depend on how the pull was batched. Merging several changesets in one session aborted with `PRIMARY KEY INSERT CONFLICT - rejecting this changeset`, while applying the same changesets one at a time succeeded.

## Changes

`ChangesetFileReader::_OnConflict` now resolves a duplicate `dgn_Domain` INSERT (`ConflictCause::Conflict`) by `Replace` regardless of pending txns. UPDATE/DELETE before-value mismatches (`ConflictCause::Data`) are left on the existing path; they are not treated as benign.

The companion iTwin.js PR fixes the JavaScript handler that takes precedence whenever the addon is hosted by JavaScript. This change covers hosts that install no JavaScript handler, such as checkpoint creation and change history, so that both handlers agree.

Reported in https://github.com/iTwin/itwinjs-core/issues/9603.


iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
```diff
@@ -421,6 +439,12 @@ const auto jsIModelDb = m_dgndb->GetJsIModelDb();
 // From the SQLite docs: "CHANGESET_CONFLICT is passed as the second argument to the conflict handler while processing an INSERT change if the operation would result in duplicate primary key values."
         // This is always a fatal error - it can happen only if the app started with a briefcase that is behind the tip and then uses the same primary key values (e.g., ElementIds)
         // that have already been used by some other app using the SAME briefcase ID that recently pushed changes. That can happen only if the app makes changes without first pulling and acquiring locks.
+        if (IsBenignDomainTableConflict(iter)) {
```
aruniverse: As a standalone native PR, this does not fix the reported BriefcaseDb/pullChanges path when a JavaScript `onChangesetConflict` handler is installed: that callback runs first at lines 377-388 and can return `Abort` before this native guard is reached. The linked issue says the normal JS-host path is the failing path, and PR 9624 is a required companion change. Please keep the native and JS guards version-coordinated and add/enable an end-to-end test against the addon containing this commit so the callback-precedence path is actually verified.

aruniverse:
Adversarial review summary: the Conflict/INSERT fallback appears to address the native no-JS path in issue #9603, but the PR is not safe or complete as written. (1) The new Data branch treats every dgn_Domain UPDATE/DELETE conflict as benign and returns Replace; please narrow it to the duplicate Conflict+INSERT case. (2) The JS callback takes precedence, so the normal BriefcaseDb path also requires the companion itwinjs-core#9624 change. (3) No native regression test was added; the linked test change should cover multi-changeset pull, native/default-handler execution, and negative dgn_Domain UPDATE/DELETE plus non-domain duplicate cases. Existing CI success does not verify these semantics.

iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
```diff
@@ -393,6 +405,12 @@ const auto jsIModelDb = m_dgndb->GetJsIModelDb();
         * is passed ApplyChangeset(). The flag will disable CASCADE action and treat
         * them as CASCADE NONE resulting in conflict handler been called.
         */
+        if (IsBenignDomainTableConflict(iter)) {
```
aruniverse: This special case is broader than the reported benign conflict. SQLite reports `ConflictCause::Data` for UPDATE/DELETE before-value (or cascade) conflicts, as the comment immediately above explains; it is not the duplicate INSERT case. Matching only `dgn_Domain` means a divergent domain update or delete is silently accepted with `Replace`, even when pending local work exists. Please restrict this guard to the duplicate insert (`cause == ConflictCause::Conflict` and an INSERT opcode/table), and leave Data conflicts on the existing path. The companion PR 9624 already narrows its JS-side guard similarly.

iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
khanaffan: Not it does not contribute to fix. But we keep same rules in native side so when we merge back into imodel02 the same rules are applied to applyChangeset() as does in itwinjs-core.

iModelCore/iModelPlatform/DgnCore/ChangesetTxns.cpp
khanaffan: Done. Dropped the `Data` special case. The native guard is now only `ConflictCause::Conflict` on an `INSERT` into `dgn_Domain`, matching itwinjs-core#9624. UPDATE/DELETE before-value mismatches stay on the existing path.

Added native tests for the pending-txn INSERT `Replace` case and for negative `dgn_Domain` UPDATE/DELETE plus non-domain duplicate INSERT.

pmconne (approved):

PR #1553, opened 2026-08-19
Remove jsoncpp 6/6: decouple the build and make the Json::Value bridge opt-in

tcobbs-bentley:
Final PR in the stack that removes all remaining **usage** of JsonCpp from this repository. PRs 1-5 migrated the source; this one decouples the build and flips the `Json::Value` bridge to opt-in.

The JsonCpp library itself **stays in the tree** at `iModelCore/libsrc/jsoncpp/` and remains standalone-buildable and deliverable, so downstream repositories that still use it are not forced to migrate.

### The stack

| | PR | Scope | Size |
|---|---|---|---|
| 6 | **#1553 (this one)** | Build decoupling + `USE_JSONCPP` opt-in | 28 files |

Split out of the original 177-file PR. The tree was byte-identical to the fully-built monolithic branch at the time of the original split; it has since diverged by the review fixes and `main` merges made during review, so the verification below was re-run from scratch against this tip (`4043dd0e1a`) rather than inherited.

> [!IMPORTANT]
> **Merge the stack as a unit, bottom-up, then the companion PR, then this one.** Only #1554 is independently safe. PRs 2-5 change public signatures that imodel-native-internal consumes, so once any of them lands, that repo's `main` will not build until #1157 lands - and #1157 cannot land before #1558, because it calls APIs that only exist in migrated form afterwards. This PR must land **after** #1157, because it is what removes JsonCpp's availability from that repo.

## Breaking change for downstream

Two things change for anyone outside these two repositories:

1. **`Bentley:BeRapidJson` no longer subparts `BeJsonCpp`.** Consumers that relied on getting JsonCpp headers transitively must now take their own `SubPart` on `iModelCore/libsrc/jsoncpp/BeJsonCpp`.
2. **The `BeJsValue` <-> `Json::Value` bridge is now opt-IN.** `BeJsValue.h` previously included `<json/json.h>` unconditionally; it now does so only when a part defines **`USE_JSONCPP`**.

A consumer that needs the bridge must do both. A consumer that doesn't use it needs to do nothing. Both failure modes are loud, never silent: missing the SubPart gives `cannot open include file: 'json/json.h'`, and missing the macro makes the `Json::Value` overloads disappear so call sites fail to compile.

### Why opt-in rather than opt-out

The first attempt was opt-out (`NO_JSONCPP` via a shared `NoJsonCpp.mki`). That put the burden on *every* part - roughly 200 - to remember to opt out, and a miss was only discoverable by building. It also created a real ODR hazard: `BeJsValue.h` is a public header with inline definitions, so two parts that disagreed about the macro got different `BeJsConst`/`BeJsValue` definitions.

In practice we were still finding missed parts one build at a time after 34 opt-out sites, and the include graph cannot be resolved statically - it uses bmake variables that only expand at build time. Inverting the default makes the correct state automatic, drops the required opt-out count to **zero**, and eliminates the ODR hazard by construction rather than by vigilance. `NoJsonCpp.mki` and all 34 of its includes are deleted.

The tradeoff: the bridge code is now compiled by nothing in these repositories, so it can rot silently. It exists solely for downstream consumers, who do compile it.

## Verification

Full clean builds **and** full test suites, green on both platforms, re-run against this tip (`f14777ee00`) using the composed strategy `iModelNative+iModelJsNodeAddon`, starting part `iModelCore:iModelCorePRG`:

| | macOS ARM64 (static) | Windows x64 (shared) |
|---|---|---|
| items built (final pass) | 114 / 114 | 121 / 121 |
| test products | 11 | 12 |
| tests | **7,865**, 0 failures | **9,406**, 0 failures |

JsonCpp's remaining footprint, split by graph:

| | production graph | test graph |
|---|---|---|
| parts linking `iTwinJsonCpp` | 0 | 1 - `BeJsonCpp:RunGtest` |
| compile targets defining `USE_JSONCPP` | 0 | 1 - `BeJsonCpp:UnitTests-NonPublished` |
| parts defining `NO_JSONCPP` | 0 | 0 |

Verified at the artifact level, not just by grepping build logs:

- Exactly one delivered product contains `iTwinJsonCpp`: `BeJsonCpp-Tests`. On Windows, `iTwinJsonCpp.lib` appears only under `build\iTwinJsonCpp\`, `BuildContexts\BeJsonCpp\`, and `Product\BeJsonCpp-Tests\Libs\` - it is never a link input to anything else, including `iModelJsNative-Dynamic`.
- The shipped `libiTwinPlatform.a` (UPack and NuGet) contains **0** JsonCpp object files and **0** `Json::` symbols; **0** `Json::` symbols across *every* archive in the shipped lib directories; **0** JsonCpp headers in either package.

The `USE_JSONCPP` row counts *compile targets*, not log occurrences, because that define is not observable the same way on both platforms: MSVC compiles the part through a response file (`cl @...BuildTestsOutOfDateCppSources1.opt`), so it never appears in the Windows log even though it is set.

Both platforms were necessary and neither was sufficient:

- **macOS** builds `[static]`, so it never reaches a shared-library link step and surfaces compile errors first. It also skips GeoCoord's `RunGtest` entirely (`OnlyPlatforms` excludes MacOS), so 1,477 GeoCoord tests never run there.
- **Windows** builds shared, so it is the only platform that exercises `LINKER_LIBRARIES` / `DLM_OBJECT_FILES`, and therefore the only one that catches a stale `iTwinJsonCpp.lib` reference.

A third build matters too: the **TilesetPublisher** in imodel-native-internal is built by **cmake**, not BentleyBuild, and compiles sources neither run above covers. That is where a missing `BeJsValue::operator=(uint64_t)` surfaced (fixed in #1554). Those sources were subsequently compiled locally against the built product to confirm they are clean.

The production-graph counts and the artifact checks are the real evidence that decoupling is complete - a green build alone is not, since a part that is simply no longer in the graph will not fail. An earlier "Build Succeeded" run silently excluded half the change. Note that the test graph now deliberately builds `BeJsonCpp` and defines `USE_JSONCPP` in order to keep the bridge and its tests compiling, so "zero occurrences anywhere" is no longer the expected result - "zero in the production graph" is.

## Notes for reviewers

- The whole stack is **EOL-clean**: `git diff` and `git diff --ignore-cr-at-eol` agree exactly on every PR, so every changed line is a real content change. This repo has mixed CRLF/LF files - some with mixed endings *within* a single file - so please avoid tooling that normalizes line endings.
- `BeJsonUtilities.h` stays in the JsonCpp tree. Its `IJsonLocalState`/`JsonLocalState` are a *different*, unused hierarchy from the `ECPresentation::` types of the same name, which were already `BeJsValue`-based.

## Known follow-ups (deliberately out of scope)

- Several pre-existing raw-RapidJson reads pass `<0>` / `kParseNoFlags` and so bypass the central full-precision fix: `ECJsonUtilities.cpp`, `SchemaReader.cpp`, `ECSqlRowAdaptor.cpp`, `ChangesetValue.cpp`, `SchemaPersistenceHelper.cpp`, `ArrayECSqlField.cpp`.
- `ChangeSummaryTestFixture.ElementChangesFromCurrentTransaction` is a pre-existing latent flake unaffected by this work: it depends on `julianday('now')` advancing between statements, so it can fail on a fast Windows run. Its sibling `ElementChildRelationshipChanges` is already in `ignore_list.txt`.
- `libsrc/compress/{lzma,lzma19,snappy}/*.mke` carry a copy-pasted comment mentioning JsonCpp that is pre-existing and unrelated.


aruniverse:
This PR is way too large, split it up. Use Stacks if you have to

iModelCore/Bentley/Tests/NonPublished/LegacyJsonNumberReadTests.cpp
```diff
@@ -0,0 +1,106 @@
...
+USING_NAMESPACE_BENTLEY
+
+// Existing iModels hold numbers written by JsonCpp's "%#.17g" formatting, and nothing rewrites
+// them. RapidJson must therefore keep recovering the identical double indefinitely, long after
+// JsonCpp itself is gone. These tests deliberately depend only on RapidJson.
+//
+// They are the regression test for the kParseFullPrecisionFlag on BeJsDocument::Parse. Several
+// entries below are values that RapidJson's DEFAULT (fast-path) number parser gets 1 ULP wrong;
+// drop the flag and this file fails.
+struct LegacyJsonNumberReadTests : ::testing::Test {};
```
aruniverse, line 19: **High:** These are the only focused regression tests for the load-bearing `kParseFullPrecisionFlag`, but this PR places them in the BeJsonCpp test product and also removes that product's `RunGtest` from `iModelCoreTesting.PartFile.xml`. The verification section explicitly says BeJsonCpp was not built, so these tests were not part of the claimed full suites and will not protect the new RapidJSON behavior. Please move them into an always-built BeRapidJson/BeJsValue test target before removing the BeJsonCpp run.

iModelCore/libsrc/jsoncpp/tests/NonPublished/BeJsonUtilitiesTests.cpp
```diff
@@ -227,6 +227,8 @@ TEST_F(StringifyTests, DoubleFormatting)
     EXPECT_EQ(str, Utf8String("1736345970855116.8\n")) << str.c_str();
     }
 
+#ifdef USE_JSONCPP
```
aruniverse, line 230: **Medium:** No in-repository target defines `USE_JSONCPP`, so this guard removes the only bridge-specific tests from every normal build at the same time the bridge semantics change. The opt-in compatibility path can now stop compiling unnoticed. Please add one small test/compile target that defines `USE_JSONCPP`, depends explicitly on both BeRapidJson and BeJsonCpp, and exercises the bridge in both header include orders.

aruniverse (commented):
Two validation gaps are inline. Also, the PR body's “two things change for anyone outside these repositories” understates the stack's exported API changes: earlier layers delete `GeomJsonWireFormat/JsonUtils.h` and change exported GeomLibs, Units, ECDb, and Visualization signatures from JsonCpp types to BeJs types. Please either document this as a coordinated breaking release with a downstream-consumer census/migration guide, or retain the compatibility overloads needed for a transition.

iModelCore/Bentley/Tests/NonPublished/LegacyJsonNumberReadTests.cpp
tcobbs-bentley: Good catch — fixed in f361cb0117.

`LegacyJsonNumberReadTests.cpp` tests `BeJsValue`, not JsonCpp, so I moved it to `iModelCore/Bentley/Tests/NonPublished/`, which is part of the always-built `Bentley-Gtest`. No mke change was needed since `DetectAndCompileUnitTests.mki` globs `$(baseDir)$(TestDir)/`, but `Bentley`'s `UnitTests-NonPublished` part now takes an explicit `<SubPart PartName="BeRapidJson" />` so `<BeRapidJson/BeJsValue.h>` resolves in that build context.

Verified locally on macosarm64: all 3 tests now run inside `BentleyTest` (425 tests, all passing).

iModelCore/libsrc/jsoncpp/tests/NonPublished/BeJsonUtilitiesTests.cpp
tcobbs-bentley: Agreed, and it was worse than "not compiled in CI" — after the decoupling *nothing* in the build defined `USE_JSONCPP`, so the bridge was not compiled anywhere and the existing `#ifdef USE_JSONCPP` tests in `BeJsonUtilitiesTests.cpp` were compiling to nothing.

Fixed in f361cb0117:

- Restored the `BeJsonCpp` `RunGtest` SubPart in `iModelCoreTesting.PartFile.xml` and the two `BeJsonCppStaticLibDir` `ProductDirectory` entries in `BeGTest.PartFile.xml`.
- `iModelCore/libsrc/jsoncpp/tests/BuildTests.mke` now does `nameToDefine = USE_JSONCPP` + `%include cdefapnd.mki`, so that target is the one place that builds the bridge.
- That part now has an explicit `<SubPart PartName="BeRapidJson" PartFile="iModelCore/Bentley/Bentley" />` alongside its existing `PublicAPI` dependency, so it genuinely depends on both sides.
- Added `BridgeIncludeOrderJsonFirstTests.cpp` and `BridgeIncludeOrderBeJsValueFirstTests.cpp`, pinning both header orders. Each also `#error`s if `USE_JSONCPP` is not defined, so the target can't silently lose the define again.

Verified locally on macosarm64: `BeJsonCppTest` runs 19 tests, all passing, including both include-order suites and the previously-dead `BeJsonUtilitiesTests` bridge tests.

iModelCore/iModelCoreTesting.PartFile.xml
```diff
@@ -24,6 +24,8 @@
         <SubPart PartName="RunGtest" PartFile="iModelCore/ECPresentation/ECPresentation" />
         <SubPart PartName="RunGtest" PartFile="iModelCore/GeomLibs/geomlibs" />
         <SubPart PartName="RunGtest" PartFile="iModelCore/Units/Units" />
+        <!-- BeJsonCpp is no longer linked into the product, but its tests are the only coverage of the
```
aruniverse, line 27: **High (validation):** Keeping this bridge test in the aggregate is the right fix, but it invalidates the PR's current verification evidence. The body still says the tested tree is byte-identical to this tip and reports `BeJsonCpp / iTwinJsonCpp = 0 / 0` and `NO_JSONCPP / USE_JSONCPP = 0 / 0`; the current full test graph now intentionally builds BeJsonCpp and defines `USE_JSONCPP`, and this tip also contains the substantive fixes added after that validation. The current #1553 checks are skipped/neutral. Please rerun the full composed macOS/Windows builds and suites at the current tree, then update the table to distinguish the production dependency graph from this test-only bridge target.

aruniverse (commented):
The prior code findings are addressed, but the current tip needs fresh validation after those fixes; see inline comment.

iModelCore/iModelCoreTesting.PartFile.xml
tcobbs-bentley: Good catch - the evidence was stale, and re-running it surfaced that the old table's premise no longer holds. Both full composed builds have been re-run from scratch at this tip (`efe9947dd4`), and the body is updated.

**macOS ARM64 (static):** 114/114 items, 11 test products, **7,865 tests, 0 failures**
**Windows x64 (shared):** 121/121 items, 12 test products, **9,406 tests, 0 failures**

The table now separates the two graphs, which is the distinction you asked for:

| | production graph | test graph |
|---|---|---|
| parts linking `iTwinJsonCpp` | 0 | 1 - `BeJsonCpp:RunGtest` |
| compile targets defining `USE_JSONCPP` | 0 | 1 - `BeJsonCpp:UnitTests-NonPublished` |
| parts defining `NO_JSONCPP` | 0 | 0 |

You were also right that the old "0 occurrences in log" metric is no longer meaningful, so I replaced it with artifact-level checks rather than log greps:

- Exactly one delivered product contains `iTwinJsonCpp`: `BeJsonCpp-Tests`. On Windows, `iTwinJsonCpp.lib` appears only under `build\iTwinJsonCpp\`, `BuildContexts\BeJsonCpp\`, and `Product\BeJsonCpp-Tests\Libs\` - never as a link input to anything else, including `iModelJsNative-Dynamic`.
- The shipped `libiTwinPlatform.a` (UPack and NuGet) contains 0 JsonCpp object files and 0 `Json::` symbols; 0 `Json::` symbols across every archive in the shipped lib dirs; 0 JsonCpp headers in either package.

One methodological note worth recording: a raw `USE_JSONCPP` log-occurrence count is **not** a sound metric across platforms. It greps 2 on macOS and 0 on Windows - not a build difference, but because MSVC compiles that part via a response file (`cl @...BuildTestsOutOfDateCppSources1.opt`), so the define never reaches the log. Confirmed by inspecting the `.opt` directly: `-DUSE_JSONCPP` is present exactly once, scoped to the three bridge TUs. The table therefore counts compile targets, not log lines.

Note that CI on this PR still reports `skipping` for `imodel-native` and `Profile-Version-Validation` - that's a trigger filter (`pr.branches.include: [main, release/*]` in `build/azure-pipelines.yml`), not a staleness artifact, so it applies to every PR in the stack except #1554, which is the only one targeting `main`. That is why the macOS/Windows evidence above comes from local composed builds at this tip.

aruniverse (approved):

iModelCore/iModelCoreTesting.PartFile.xml
aruniverse: Resolved after verifying the update. The author re-ran both full composed builds from scratch at the current tip `efe9947dd4`: macOS ARM64 completed 114/114 items with 7,865 tests passing, and Windows x64 completed 121/121 items with 9,406 tests passing. The PR body now makes the production/test-graph distinction explicit: the production graph has no `iTwinJsonCpp` link or `USE_JSONCPP` target, while the test graph intentionally has one of each for `BeJsonCpp:RunGtest` / `BeJsonCpp:UnitTests-NonPublished`. It also replaces the stale inherited byte-identical/zero-log-count evidence with validation tied to this exact tip.

tcobbs-bentley:
This has a bunch of code owners. Could I get one reviewer from each row below? Thanks.

Edit: done, thanks.

| Directory | Reviewers |
| --- | --- |
| ✅ `iModelCore/Bentley` | @pmconne @bbastings @khanaffan |
| ✅ `iModelCore/ECDb` | @khanaffan @pmconne @calebmshafer @rschili @RohitPtnkr1996 |
| ✅ `iModelCore/ecobjects` | @calebmshafer @cammac1803 @rschili @RohitPtnkr1996 |
| ✅ `iModelCore/ECPresentation` | @saskliutas @grigasp @MartynasStrazdas @JonasDov |
| ✅ `iModelCore/GeoCoord` | @pmconne @AlainRobertAtBentley @bbastings |
| ✅ `iModelCore/GeomLibs` | @bbastings @dassaf4 @saeeedtorabi |
| ✅ `iModelCore/iModelPlatform/` | @pmconne @khanaffan @bbastings |
| ✅ `iModelJsNodeAddon` | @pmconne @khanaffan @calebmshafer @bbastings |

pmconne (approved):

dassaf4 (approved):
Go for Geomlibs.

grigasp (approved):

RohitPtnkr1996 (approved):

PR #1554, opened 2026-08-19
Remove jsoncpp 1/6: add the BeJsValue capabilities the migration needs

tcobbs-bentley:
Adds the `BeJsValue` capabilities the rest of the stack depends on. **Purely additive** - no call sites change and the `Json::Value` bridge is untouched, so this one is safe to merge on its own.

### The stack

| | PR | Scope | Size |
|---|---|---|---|

### What's here

- **`Parse` with `kParseFullPrecisionFlag`.** JsonCpp parsed doubles with `strtod`; RapidJson's default parser can be 1 ULP off. Load-bearing, and carries a DO-NOT-REMOVE comment.
- **`IntegralFromDouble<T>`** behind `GetInt`/`GetUInt`/`GetInt64`, which previously read the raw union when the value happened to be stored as a double.
- **`StringifyLegacy()`**, reproducing JsonCpp's sorted keys and `%#.17g` double formatting, for the few sites whose exact bytes are persisted.
- **`SetString()`**, so assigning a string can be spelled unambiguously - `BeJsDocument doc = str` parses, while `BeJsValue val = str` assigns, with no diagnostic to tell them apart.
- **`operator=(uint64_t)`.** `BeJsValue` could *read* a uint64 (`asUInt64`) but had no way to *write* one; JsonCpp had `Json::Value(UInt64)`. Without it the assignment is ambiguous across five overloads. Covered by a new `BeJsValueTest.AssignUInt64`, since nothing in this repo's own builds exercises it.
- **Explicit STL includes** that MSVC required once the JsonCpp header stops supplying them transitively.

> [!IMPORTANT]

### DIFF
--- a/iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
+++ b/iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
@@ -18,9 +18,19 @@
 #include <Bentley/Base64Utilities.h>
 #include <BeRapidJson/BeRapidJson.h>
 #include <json/json.h>
+#include <algorithm>
+#include <cstdio>
+#include <cstring>
 #include <limits>
 #include <cmath>
+#include <functional>
 #include <optional>
+// These were previously reaching this header transitively through <json/json.h>. libc++ still
+// leaks them in via the headers above, but the MSVC STL does not, so declare them explicitly.
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
 
 namespace Napi {
 class Value;
@@ -35,6 +45,29 @@ enum StringifyFormat {
     Default,
     Indented,
 };
+
+//=======================================================================================
+// A name with static storage duration. Implementations may store the pointer rather than
+// copying the characters, so the referenced string must outlive every value that uses it.
+// @bsiclass
+//=======================================================================================
+struct BeJsStaticString {
+private:
+    Utf8CP m_str;
+
+public:
+    constexpr explicit BeJsStaticString(Utf8CP str) : m_str(str) {}
+    constexpr operator Utf8CP() const { return m_str; }
+    constexpr Utf8CP c_str() const { return m_str; }
+};
+
+// Declares a `json_<name>()` accessor returning the member name as a static string.
+// JsonCpp declares a Json::StaticString-based version of this macro. Redefine it here so a
+// translation unit gets the same expansion no matter which header it included first;
+// BeJsStaticString converts to Utf8CP, so JsonCpp call sites keep compiling.
+#undef BE_JSON_NAME
+#define BE_JSON_NAME(__val__) static constexpr BentleyApi::BeJsStaticString json_##__val__() {return BentleyApi::BeJsStaticString(#__val__);}
+
 //=======================================================================================
 // @internal
 // @bsiclass
@@ -85,6 +118,9 @@ protected:
     virtual void operator=(int32_t value) = 0;
     virtual void operator=(uint32_t value) = 0;
     virtual void operator=(int64_t value) = 0;
+    // Not pure virtual: an out-of-tree implementation of this interface would otherwise fail to
+    // compile. The default is lossy above INT64_MAX; in-tree implementations override it to be exact.
+    virtual void operator=(uint64_t value) { *this = static_cast<int64_t>(value); }
     virtual void operator=(Utf8CP value) = 0;
     virtual double GetDouble(double defVal = 0) const = 0;
     virtual bool GetBoolean(bool defVal) const = 0;
@@ -344,6 +380,19 @@ public:
     // Stringify this value and all its children.
     // @note a null value returns an empty string, not "null". This differs from rapidjson's api.
     Utf8String Stringify(StringifyFormat format = StringifyFormat::Default) const { return m_val->isNull() ? "" : m_val->Stringify(format); }
+    // Stringify this value and all its children so that the result is BYTE-IDENTICAL to what
+    // Bentley's JsonCpp fork produced for the same data (Json::Value::ToString()).
+    // Two things differ between JsonCpp and rapidjson and both are reproduced here:
+    //   1. object members are emitted in alphabetical order, not insertion order;
+    //   2. doubles are spelled with "%#.17g" and trailing zeros trimmed to one (so 0.3 becomes
+    //      "0.29999999999999999" and 1.5 becomes "1.50"), not rapidjson's shortest round-trip form.
+    // @note This exists ONLY for JSON that is PERSISTED and later hashed or compared as text
+    // (e.g. ec_Format.NumericSpec/CompositeSpec, ec_Enumeration.EnumValues). Those bytes are
+    // covered by PRAGMA checksum(ecdb_schema), which SchemaSync compares ACROSS briefcases, so a
+    // briefcase written by old code and one written by new code must agree exactly.
+    // Anything that merely round-trips through a parser should use Stringify() instead: this is
+    // slower, uglier, and is not needed for correctness.
+    Utf8String StringifyLegacy() const;
     // compare this value to another for equality, using default tolerances
     bool operator==(BeJsConst other) const { return isAlmostEqual(other); }
     // compare this value to another for inequality, using default tolerances
@@ -396,6 +445,7 @@ public:
     // get the value of a member of an object using a static string. If it doesn't exist, it is created.
     // @note this value must be an object
     // @note this can be less expensive for some implementations since a reference to the string can be stored in the returned object.
+    BeJsValue operator[](BeJsStaticString const& key) { return BeJsValue(m_val->GetMember(key, true)); }
     BeJsValue operator[](Json::StaticString const& key) { return BeJsValue(m_val->GetMember(key, true)); }
     // get the value of a member of an object. If it doesn't exist, it is created.
     // @note this value must be an object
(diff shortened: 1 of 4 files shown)


pmconne (approved):

khanaffan (approved):

khanaffan:
Note we use RapidJson which is outstanding for parsing file. But we do not really have large xml to parse so RapidJson isn't really very useful beside its very old library.

I think at point we should just abandon this old libraries and move to modern header only https://github.com/nlohmann/json . Its is clean modern.

rschili (approved):

tcobbs-bentley:
> Note we use RapidJson which is outstanding for parsing file. But we do not really have large xml to parse so RapidJson isn't really very useful beside its very old library.
> 
> I think at point we should just abandon this old libraries and move to modern header only https://github.com/nlohmann/json . Its is clean modern.

The point of this PR stack is to update all the remaining jsoncpp usage to instead use RapidJson. Are you saying we should **not** do that but instead create a completely different PR stack that replaces all RapidJson and jsoncpp usage with that other json library?

rschili:
> The point of this PR stack is to update all the remaining jsoncpp usage to instead use RapidJson. Are you saying we should **not** do that but instead create a completely different PR stack that replaces all RapidJson and jsoncpp usage with that other json library?

I could swear we discussed this about a year ago in our team and filed a story with potential modernizations. For example pugixml which some of native already uses.
But I just searched through our entire backlog, and I cannot find an issue about it.
We DID research alternative json libraries, which is why we noticed RapidJson was basically unmaintained since 2016.

Switching all our code to a new library seems like a lot of work though. We do have the abstraction layer, but still.

The one Affan mentioned seems to be the slowest of the modern bunch, other alternatives: yyjson, simdjson, Glaze - probably needs some research into which one fits our existing patterns best.

pmconne:
> I think at point we should just abandon this old libraries and move to modern header only https://github.com/nlohmann/json

>> Are you saying we should not do that but instead create a completely different PR stack that replaces all RapidJson and jsoncpp usage with that other json library?

I am all but certain that he meant "at **some** point (in the future)" we should consider making that change. Your changes are good and necessary and will reduce the amount of work in migrating to a different JSON library when/if we ever decide to do so.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -565,20 +641,20 @@ struct BeJsValue : BeJsConst {
         }
         if (other.isNumeric()) {
             double val = other.GetDouble();
-            if (val > Json::Value::minInt64() && Json::Value::IsIntegral(val)) {
-                if (val < Json::Value::minInt()) {
+            if (val > s_minSafeInteger && IsIntegral(val)) {
```
aruniverse, line 677: **High:** `FromOther` still routes every numeric value through `double`. A `.From()` copy of an `int64_t`/`uint64_t` above `2^53` silently changes the value and its JSON representation; later stack layers add new `.From()` call sites, including ECDb result copying. Please preserve the stored integer kind instead of going through `GetDouble()`, and add copy/serialize tests for `INT64_MAX`, `(1ULL << 53) + 1`, and `UINT64_MAX`.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -838,17 +917,54 @@ struct BeRapidJsonValue : JsValueRef {
...
+        if (!isNumeric())
+            return defVal;
+        return m_value->IsInt() ? m_value->GetInt() : IntegralFromDouble<int32_t>(m_value->GetDouble());
+    }
+    virtual uint32_t GetUInt(uint32_t defVal) const override {
+        if (!isNumeric())
+            return defVal;
+        return m_value->IsUint() ? m_value->GetUint() : IntegralFromDouble<uint32_t>(m_value->GetDouble());
+    }
+    virtual int64_t GetInt64(int64_t defVal) const override {
```
aruniverse, line 988: **High:** This clamping implementation is bypassed by the public `BeJsConst::GetInt64` wrapper above: its numeric path still does `int64_t(GetDouble(...))` directly. That leaves NaN and out-of-range doubles reaching an undefined floating-to-integer conversion, despite this PR's stated fix. Please retain the string special case and delegate numeric values to `m_val->GetInt64(defVal)`; add NaN and boundary tests through the public API.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -838,17 +917,54 @@ struct BeRapidJsonValue : JsValueRef {
...
     virtual uint64_t GetUInt64(uint64_t defVal) const override {
+        if (!isNumeric())
+            return defVal;
+        if (m_value->IsUint64()) {
+            return m_value->GetUint64();
+        }
+        // Unlike the accessors above, a non-integral double deliberately yields defVal rather than a
+        // truncated value: this accessor is used to round-trip 64-bit ids, where silently losing the
+        // fractional part would hide a real error.
         if (m_value->IsDouble() && IsLosslessUint64(m_value->GetDouble())) {
```
aruniverse, line 1002: **High:** `IsLosslessUint64` compares against `static_cast<double>(UINT64_MAX)`, which rounds up to exactly `2^64`. A double equal to `2^64` therefore passes the current `>` check and is cast to `uint64_t`, which is out of range and undefined. Please use an exactly representable exclusive `2^64` bound (`d >= 0x1p64`) before casting, and cover that boundary in a test.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
```diff
@@ -966,6 +1109,104 @@ struct BeJsDocument : BeJsValue {
...
+        out.append("null");
+        return;
+    }
+    if (in.isBool()) {
+        out.append(in.asBool() ? "true" : "false");
+        return;
+    }
+    if (in.isNumeric()) {
+        // rapidjson always spells a double with a '.' or an exponent, and an integer with neither.
+        Utf8String text = in.Stringify();
```
aruniverse, line 1254: **Medium:** This calls RapidJSON `Stringify()` before handling non-finite doubles. The normal RapidJSON writer rejects NaN/Infinity, so `text` can be empty and `StringifyLegacy()` can emit invalid JSON such as `{"x":}` instead of JsonCpp-compatible `null`. Please check `std::isfinite(in.asDouble())` first and append `null` for non-finite values; nested object/array tests would catch this.

aruniverse (commented):
Adversarial pass on the current stack found four correctness issues in the BeJsValue foundation; see inline comments. Separately, the current head (`db7fbf4664`) still fails all eight native build legs in Azure build 5589321, even though this layer is described as independently safe. I would not start the merge sequence until those failures are diagnosed and this head is green.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
tcobbs-bentley: Fixed in 87558d3de2. Added `JsValueRef::GetNumericKind()` (non-pure virtual, so out-of-tree implementers keep compiling) and `FromOther` now copies values outside the JS safe-integer range through the exact integer accessors instead of `GetDouble()`. Values inside the safe range keep their previous narrowing behaviour, so existing type expectations are unchanged. Added `FromPreservesLargeIntegers` covering copy *and* serialization of `INT64_MAX`, `(1<<53)+1` and `UINT64_MAX`, plus a copy-of-a-copy stability check.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
tcobbs-bentley: Fixed in 87558d3de2. `BeJsConst::GetInt64` now keeps only the string special case and delegates everything else to `m_val->GetInt64(defVal)`, which clamps. Added `GetInt64ClampsAndRejectsNonFinite` covering NaN, ±Infinity, ±1e300, truncation toward zero, and the decimal/hex string paths.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
tcobbs-bentley: Fixed in 87558d3de2. Confirmed your analysis empirically: `static_cast<double>(UINT64_MAX) == 0x1p64` is true, so the old `>` test let 2^64 through. Now uses the exclusive bound `d >= 0x1p64` (and `!(d >= 0.0)`, which also rejects NaN); same for the `float` overload. Added `GetUInt64RejectsTwoToThe64` covering exactly 2^64, the largest double below it, negative, NaN and fractional inputs.

iModelCore/Bentley/BeRapidJson/PublicAPI/BeRapidJson/BeJsValue.h
tcobbs-bentley: Fixed in 87558d3de2. `BeJsLegacyStringify` now checks `std::isfinite` before calling `Stringify()` and appends `null` for non-finite values. Added `StringifyLegacyWritesNullForNonFinite` with NaN/±Infinity both at top level and nested inside an object and an array.

aruniverse:
Latest follow-up: the current head `05f50f48c7` is still red on all eight native build legs in Azure build 5595538 (10 errors total). This PR is the independently mergeable base of the stack, so the later skipped/neutral stacked checks do not substitute for a green run here. Please diagnose or link the common failure and rerun this head before starting the merge sequence: https://dev.azure.com/bentleycs/2c48216e-e72f-48b4-a4eb-40ff1c04e8e4/_build/results?buildId=5595538

PR #1555, opened 2026-08-19
Remove jsoncpp 2/6: GeomLibs

tcobbs-bentley:
Replaces JsonCpp usage in GeomLibs with `BeJsValue`, removing the `Json::Value`-based helpers in `GeomJsonWireFormat/JsonUtils.h` and migrating the BeCG serializer and the geometry tests.

`BentleyGeometryJson::TryJsonStringToGeometry` deliberately keeps lenient parse flags - it is the one place that legitimately accepted the comments and trailing commas JsonCpp allowed.

imodel-native-internal: https://github.com/iTwin/imodel-native-internal/pull/1157

### The stack

| | PR | Scope | Size |
|---|---|---|---|

> [!IMPORTANT]


iModelCore/GeomLibs/PublicAPI/GeomJsonWireFormat/BeJsGeomUtils.h
```diff
@@ -137,7 +137,7 @@ struct BeJsGeomUtils {
     static DRange3d ToDRange3d(BeJsConst inValue) {
         DRange3d range = DRange3d::NullRange();
         if (inValue.isArray()) { // if it's an array, just extend range by all points.
-            for (Json::ArrayIndex i = 0; i < inValue.size(); ++i)
+            for (BeJsConst::ArrayIndex i = 0; i < inValue.size(); ++i)
```
dassaf4: Should this be `BeJsValue` like the other change in this file?

iModelCore/GeomLibs/serialization/src/BeJsonToCG.cpp
```diff
@@ -152,49 +153,50 @@ GEOMAPI_VIRTUAL bool ReadTag_AnyICurvePrimitive (TSource const &source, CharCP n
...
     }
 
 
-struct BeJsonToCGReaderImplementation : BeDirectReader<Json::Value>
+struct BeJsonToCGReaderImplementation : BeDirectReader<BeJsConst>
 {
 Transform m_placement;
 
-typedef bool (BeDirectReader<Json::Value>::*ParseMethod)(Json::Value const &, IGeometryPtr &);
+typedef bool (BeDirectReader<BeJsConst>::*ParseMethod)(BeJsConst const &, IGeometryPtr &);
```
dassaf4, line 199: There are 43 instances of `BeJsConst const &` which I'm pretty sure should just be `BeJsConst` in this file.  Maybe other files too...

dassaf4 (approved):

iModelCore/GeomLibs/serialization/src/BeJsonToCG.cpp
tcobbs-bentley: These are due to the `BeDirectReader` using `TSource const &`. Our subclass of that class uses `BeJsConst` as the `TSource`, but we have to retain the declared `const &` modifiers in our instantiations. (It's not obvious in this function pointer typedef, but even this is locked in by the templated superclass.) It is true that these wouldn't normally be appropriate with `BeJsConst`, but they don't cause problems, and they are required for the code to work.

PR #1556, opened 2026-08-19
Remove jsoncpp 3/6: Units and ecobjects

tcobbs-bentley:
Replaces JsonCpp usage in Units and ecobjects with `BeJsValue`.

The notable behaviour to preserve here is JsonCpp's **string-to-number coercion**: Bentley's fork patched `asInt`/`asDouble`/etc. with `case stringValue: sscanf(...)`, and Units depended on it. That tolerance is now explicit, in the helpers in `FormattingParsing.h`, which clamp before the double-to-int cast to avoid undefined behaviour.

One deliberate behaviour change: a negative string such as `"-5"` now clamps to 0 for unsigned rather than producing the garbage JsonCpp's `sscanf("%u")` returned.

### The stack

| | PR | Scope | Size |
|---|---|---|---|

> [!IMPORTANT]


tcobbs-bentley:
@rschili @RohitPtnkr1996 Could I get a review of this PR, please?

rschili (approved):

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
```diff
@@ -236,4 +240,66 @@ struct FormattingScannerCursor
...
+    {
+    if (!val.isString())
+        return false;
+    Utf8String str = val.asString();
+    if (str.empty())
+        return false;
+    // JsonCpp used sscanf("%lf"/"%d"/"%u"); strtod matches its locale-dependent behavior while
+    // also letting us detect the "no digits consumed" case that sscanf reported via its return.
+    char* end = nullptr;
+    double parsed = strtod(str.c_str(), &end);
```
aruniverse, line 264: **High:** `strtod` accepts `"nan"` as a successful parse. In `JsonToUInt`/`JsonToInt64`, both bounds comparisons are then false and NaN is cast to an integer, which is undefined behavior. Please reject non-finite parsed values before integer conversion (returning `defaultVal`) and add regression cases for `nan` and infinities.

aruniverse (commented):
One undefined-behavior edge case remains in the new string-number compatibility helper; see inline comment.

iModelCore/Units/PrivateAPI/Formatting/FormattingParsing.h
tcobbs-bentley: Fixed in 2bb4cbe5fc. `JsonToUInt`/`JsonToInt64` now reject non-finite parsed values and return `defaultVal` before any integer conversion. Verified that `strtod` does report success for `"nan"`, `"-NaN"`, `"inf"` and `"-Infinity"`. Added `JsonStringCoercionRejectsNonFinite` for all four, plus `JsonStringCoercionClampsOutOfRange` to pin the existing clamping behaviour. `JsonToDouble` is deliberately unchanged, since returning the parsed non-finite value matches what the JsonCpp fork did.

PR #1557, opened 2026-08-19
Remove jsoncpp 4/6: ECDb

tcobbs-bentley:
Replaces JsonCpp usage in ECDb with `BeJsValue`.

**Persisted data:** `SchemaPersistenceHelper.cpp` and `ProfileUpgrader.cpp` write JSON whose exact bytes are persisted and checksummed by SchemaSync (`PRAGMA checksum`), so they use `StringifyLegacy()` to keep producing JsonCpp's sorted-key, `%#.17g` output. Without it the stored text changes and the checksum moves.

Also fixes test data that only ever parsed because JsonCpp was permissive - leading zeros in numbers, and trailing content after the first complete value, which JsonCpp silently ignored.

### The stack

| | PR | Scope | Size |
|---|---|---|---|

> [!IMPORTANT]


khanaffan (approved):

rschili (approved):

iModelCore/ECDb/Tests/BackDoor/TestInfoHolders.cpp
```diff
@@ -17,74 +18,70 @@ BEGIN_ECDBUNITTESTS_NAMESPACE
...
+        {
+        if (!rhs.isNumeric())
+            return false;
 
-    if (m_value.isUInt())
-        return rhs.m_value.isConvertibleTo(Json::uintValue) && m_value.asUInt64() == rhs.m_value.asUInt64();
+        double l = lhs.asDouble();
+        double r = rhs.asDouble();
+        if (l == std::trunc(l) && r == std::trunc(r))
+            return lhs.asInt64() == rhs.asInt64();
```
aruniverse, line 82: **Medium:** This comment promises exact integral comparison above `2^53`, but every integral value is converted through signed `asInt64()`. Distinct `uint64_t` values above `INT64_MAX` can therefore undergo out-of-range conversion and compare equal, masking migration regressions. Please compare non-negative integral values via `asUInt64()` and negative values via `asInt64()`, with a case containing two distinct values above `INT64_MAX`.

aruniverse (commented):
One test-comparator issue can mask the large-integer regressions this migration is intended to catch; see inline comment.

iModelCore/ECDb/Tests/BackDoor/TestInfoHolders.cpp
tcobbs-bentley: Fixed in e88894086c. Non-negative integral values now compare via `asUInt64()` and negative ones via `asInt64()`; magnitudes outside the exactly representable integer range fall through to the tolerant double comparison rather than reporting a bogus match. Updated the misleading comment too. Added `JsonValueHelper.LargeUnsignedComparisons` with `UINT64_MAX` vs `UINT64_MAX - 1`, both bare and nested in an object.

tcobbs-bentley: Follow-up: my first fix here was itself wrong, corrected in 300b6c8b05.

I had gated the exact-integer path on `asDouble()` staying inside `[-0x1p63, 0x1p64)`, which is lossy at exactly the range this thread is about: `UINT64_MAX` and `UINT64_MAX-1` both round to exactly `0x1p64`, so the guard rejected them, they fell through to the tolerant double comparison, and they compared equal anyway. The regression test I added caught it on the first full build.

It now keys off `BeJsConst::GetNumericKind()` rather than round-tripping through `double` at all. When neither side is stored as a double, the integers are compared directly — via `asUInt64` when either side only fits in a `uint64_t` (rejecting a negative counterpart first), `asInt64` otherwise. Genuine floating point values keep the `TestUtilities::Equals` tolerance, so `1` still equals `1.0`.

Verified locally on macosarm64: `JsonValueHelper.LargeUnsignedComparisons` passes and the full `ECDbTest` suite is green (1653 tests).

PR #1558, opened 2026-08-19
Remove jsoncpp 5/6: iModelPlatform, ECPresentation, GeoCoord, node addon

tcobbs-bentley:
Replaces JsonCpp usage in iModelPlatform, ECPresentation, GeoCoord and the node addon with `BeJsValue`.

**Project extents** (`DgnCore/DgnUnits.cpp`) are persisted but deliberately do *not* use `StringifyLegacy()`: tile content Ids hash the recovered `DRange3d` rather than the text, and reads go through `kParseFullPrecisionFlag`, so the doubles round-trip bit-identically from either form. Documented in place.

Also fixes `GCSUnitTests.cpp`, where 34 entries had trailing content after the first complete JSON value that JsonCpp had silently ignored.

### The stack

| | PR | Scope | Size |
|---|---|---|---|

> [!IMPORTANT]


iModelCore/ECPresentation/Source/TaskScheduler.cpp
```diff
@@ -7,6 +7,7 @@
 #include <folly/executors/InlineExecutor.h>
 #include "RulesEngineTypes.h"
 #include "TaskScheduler.h"
+#include <deque>
```
grigasp: Could we add `<string>` and `<deque>` to `ECPresentationPch.h` instead of adding them to each source file?

pmconne (approved):
Code changes look good (I didn't look at ECPresentation). I could nitpick that littering the code with verbose comments about divergence from JsonCpp will just confuse anyone who doesn't know/care that we used to use JsonCpp - you might consider moving those explanations into comments on this PR instead - that's where people generally look for "why was this change made". But it's your call, unless others have strong opinions.

iModelCore/ECPresentation/Source/TaskScheduler.cpp
tcobbs-bentley: The intent was to only include these where they are actually needed, but I don't mind doing what you suggest.

tcobbs-bentley: Fixed for ECPresentation and its tests.

tcobbs-bentley:
> Code changes look good (I didn't look at ECPresentation). I could nitpick that littering the code with verbose comments about divergence from JsonCpp will just confuse anyone who doesn't know/care that we used to use JsonCpp - you might consider moving those explanations into comments on this PR instead - that's where people generally look for "why was this change made". But it's your call, unless others have strong opinions.

I have done this for the comments that aren't warnings to future maintainers. The changes span 5 levels of the 6-level the PR stack.

grigasp (approved):

rschili (approved):

tcobbs-bentley:
@grigasp @saskliutas @MartynasStrazdas @JonasDov could I get somebody to review the changes in `iModelCore/ECPresentation`?

@bbastings @dassaf4 @saeeedtorabi could I get somebody to review the changes in `iModelCore/GeomLibs`?

dassaf4 (approved):
Go for Geomlibs.

tcobbs-bentley:
Apparently I missed the ECPresentation approval, so this PR now shows approved.

PR #1561, opened 2026-08-20
Preserve nullable geometric placement data during serialization/deserialization

hl662:
## TL;DR

Geometric elements (both 2D and 3D) with nullable placement fields were serialized with fabricated default values, so transformation changed Yaw/Pitch/Roll to 0 and BBoxLow/BBoxHigh to DBL_MAX (when it should be just NULL). This PR adds tracking which placement fields were actually present and preserves that information through JSON, SQL binding, and native clone/import.

3D elements with invalid or NULL bounding boxes are also no longer added to the spatial index. (new as of Aug 21)

> I spent lot of time and effort learning C++ through this and handwriting all the code. Macros were confusing @@

## Why

`Placement2d` and `Placement3d` use default in-memory values when database columns are `NULL`, so the placement object alone cannot distinguish an absent value from a real zero or default value. Native serialization then emitted those defaults, and clone/import persisted them into the target iModel.

## Shape

```text
SQL row with nullable placement
    ↓ _ReadSelectParams()
placement value + presence flags (these presence flags are new!)
    ↓ _ToJson(), _BindWriteParams(), clone/import
original NULL/value representation preserved (thanks to the presence flags)
```

Instead of making 3 separate bool check properties in `GeometricElement` and having to worry about ABI (Application Binary Interface), I opted to reuse the `DgnFlag` struct available in `DgnElement`, which already has 32 bits allocated to our existing elements. But only 3 bits were used, so I added 3 new bits for `m_placementData` and used bit masking to keep track of whether a DgnElement had NULL placement properties or not. Placement data isn't part of all DgnElement, so I exposed helper methods to read and set the flags in **only `GeometricElement`**

An omitted placement remains NULL, an origin-only placement remains origin-only, and a complete placement remains complete.

## What

| Asset | Why it exists |
|---|---|
| `iModelCore/iModelPlatform/PublicAPI/DgnPlatform/DgnElement.h` | Default placement values lose SQL nullability — Adds a 3 bit flag to an existing DgnFlag struct, storing three placement-presence bits and initializes/copies them with the element state. |
| `iModelCore/iModelPlatform/DgnCore/DgnElement.cpp` | Native read, JSON, SQL-binding, property-accessor, and clone/import paths previously treated default placement values as present — now they preserve only the placement groups that actually exist. |
| `iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnElement_Test.cpp` | Adds focused tests for absent, origin-only, and complete placements. A test snapshot was updated to reflect this |

## Tests

<details>
<summary>New and updated test coverage</summary>

| Test | What it covers |
|---|---|
| `DgnElementTests.SpatialLocationNullPlacementToJson` | A spatial location with no placement does not serialize a fabricated placement object. |
| `DgnElementTests.OriginOnlyPlacementToJson` | An origin-only placement remains origin-only in JSON without acquiring angles or a bbox. |
| `DgnElementTests.PlacementJsonPreservesNulls` | Complete and origin-only 3D placements preserve JSON, raw SQL NULLs, reload behavior, native import/clone behavior, and the equivalent 2D representation. |

</details>

---
*Nam (Nambot just made the first draft of this description)*


iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
```diff
@@ -2962,43 +2968,52 @@ void dgn_ElementHandler::Geometric3d::_RegisterPropertyAccessors(ECSqlClassInfo&
...
-#define SETGEOMPLCPROP(PTYPE, EXPR) [](DgnElement& elIn, ECN::ECValueCR valueIn)\
+#define SETGEOMPLCPROP(PTYPE, FLAGS, EXPR) [](DgnElement& elIn, ECN::ECValueCR valueIn)\
             {                                                                          \
             if (valueIn.IsNull() || valueIn.IsBoolean() || !valueIn.IsPrimitive())       \
                 return DgnDbStatus::BadArg;                                              \
             ECN::ECValue value(valueIn);                                                 \
             if (!value.ConvertToPrimitiveType(PTYPE))                                    \
                 return DgnDbStatus::BadArg;                                              \
             GeometricElement3d& el = (GeometricElement3d&)elIn;                          \
+            uint8_t placementDataFlags = el.GetPlacementDataFlags();                    \
```
tcobbs-bentley: Nit: even though they were partially broken before, these EOL backslashes should line up on all the lines except the new one that is too long.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
```diff
@@ -3075,34 +3090,43 @@ void dgn_ElementHandler::Geometric2d::_RegisterPropertyAccessors(ECSqlClassInfo&
...
-#define SETGEOMPLCPROP(PTYPE, EXPR) [](DgnElement& elIn, ECN::ECValueCR valueIn)\
+#define SETGEOMPLCPROP(PTYPE, FLAGS, EXPR) [](DgnElement& elIn, ECN::ECValueCR valueIn)\
             {                                                                          \
             if (valueIn.IsNull() || valueIn.IsBoolean() || !valueIn.IsPrimitive())       \
                 return DgnDbStatus::BadArg;                                              \
             ECN::ECValue value(valueIn);                                                 \
             if (!value.ConvertToPrimitiveType(PTYPE))                                    \
                 return DgnDbStatus::BadArg;                                              \
             GeometricElement2d& el = (GeometricElement2d&)elIn;                          \
+            uint8_t placementDataFlags = el.GetPlacementDataFlags();                    \
```
tcobbs-bentley: Same as above for EOL backslashes.

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnElement_Test.cpp
```diff
@@ -250,6 +250,172 @@ TEST_F(DgnElementTests, GenericDomainElements)
...
+    {
+        SetupSeedProject();
+
+        PhysicalModelPtr model = GetDefaultPhysicalModel();
+        GenericSpatialLocationPtr element = GenericSpatialLocation::Create(*model, GetDefaultCategoryId());
+        ASSERT_TRUE(element.IsValid());
+
+        BeJsDocument json;
+        element->ToJson(json);
+        EXPECT_FALSE(json.hasMember("placement"));
```
tcobbs-bentley: This has an extra indent.

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnElement_Test.cpp
```diff
@@ -250,6 +250,172 @@ TEST_F(DgnElementTests, GenericDomainElements)
...
+        element->FromJson(placementProps);
+
+        BeJsDocument json;
+        element->ToJson(json);
+        EXPECT_TRUE(json.hasMember(GeometricElement::json_placement()));
+        EXPECT_TRUE(json[GeometricElement::json_placement()].hasMember(Placement3d::json_origin()));
+        EXPECT_EQ(expectedOrigin, BeJsGeomUtils::ToDPoint3d(json[GeometricElement::json_placement()][Placement3d::json_origin()]));
+
+        EXPECT_FALSE(json[GeometricElement::json_placement()].hasMember(Placement3d::json_angles()));
+        EXPECT_FALSE(json[GeometricElement::json_placement()].hasMember(Placement3d::json_bbox()));
```
tcobbs-bentley: This has an extra indent.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
hl662: is there like no formatter to prettify our stuff 🥲  anws fixed

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
hl662: tks fixed

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnElement_Test.cpp
hl662: tks fixed

iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnElement_Test.cpp
hl662: tks fixed

hl662:
So... this PR got a bit bigger, because it surfaced an existing correctness bug that our CI tests using itwinjs-core caught.

We caught three failures in the field-driven TextAnnotation workflow, TextAnnotation persistence, and the placement round-trip with undefined geometry. They all surfaced as `Error inserting element [write error]`, with SQLite reporting `rtree constraint failed: dgn_SpatialIndex.(MinX<=MaxX)`.

This is because `GeometricElement3d::_BindWriteParams` was marking every element in a spatially located model as `InSpatialIndex=1`, even when the element had no valid bounding box. The existing SQL trigger (is this right use of the term?) then tried to insert an invalid NULL range into `dgn_SpatialIndex`, SQLite’s R-tree index for accelerating spatial queries.

I (okay this part AI did) fixed this by setting `InSpatialIndex=1` only when the 3d element has a complete, valid placement with a usable bounding box. 2d elements don't need this logic, afaik we don't use spatial indexing for 2d elements (or the bbox is alr excluded I dunno).

Partial placements still persist their supplied fields, but they are now not indexed. The UPDATE triggers now respond when`InSpatialIndex` changes and remove the corresponding R-tree row, and the native update path also removes stale rows for existing iModels that still have the legacy trigger definitions.

So with this, we are reducing the spatial-index footprint: elements without a valid spatial extent no longer consume R-tree rows, while valid elements remain indexed and invalid elements become indexed again when/if a valid bounding box is restored. This change happens to affected elements; this change does not perform a database-wide R-tree rebuild.

I also created a new regression test.

iModelCore/ECDb/ECDb/IntegrityChecker.cpp
```diff
@@ -348,8 +348,8 @@ DbResult IntegrityChecker::CheckProfileTablesAndIndexes4002AndLater(std::functio
...
-		{"dgn_rtree_upd1", "CREATE TRIGGER dgn_rtree_upd1 AFTER UPDATE OF Origin_X,Origin_Y,Origin_Z,Yaw,Pitch,Roll,BBoxLow_X,BBoxLow_Y,BBoxLow_Z,BBoxHigh_X,BBoxHigh_Y,BBoxHigh_Z ON bis_GeometricElement3d WHEN OLD.Origin_X IS NOT NULL AND NEW.Origin_X IS NULL BEGIN DELETE FROM dgn_SpatialIndex WHERE ElementId=OLD.ElementId;END"},
+		{"dgn_rtree_upd", "CREATE TRIGGER dgn_rtree_upd AFTER UPDATE OF InSpatialIndex,Origin_X,Origin_Y,Origin_Z,Yaw,Pitch,Roll,BBoxLow_X,BBoxLow_Y,BBoxLow_Z,BBoxHigh_X,BBoxHigh_Y,BBoxHigh_Z ON bis_GeometricElement3d WHEN new.Origin_X IS NOT NULL AND 1 = new.InSpatialIndex BEGIN INSERT OR REPLACE INTO dgn_SpatialIndex(ElementId,minx,maxx,miny,maxy,minz,maxz) SELECT new.ElementId,DGN_bbox_value(bb,0),DGN_bbox_value(bb,3),DGN_bbox_value(bb,1),DGN_bbox_value(bb,4),DGN_bbox_value(bb,2),DGN_bbox_value(bb,5) FROM (SELECT DGN_placement_aabb(DGN_placement(DGN_point(NEW.Origin_X,NEW.Origin_Y,NEW.Origin_Z),DGN_angles(NEW.Yaw,NEW.Pitch,NEW.Roll),DGN_bbox(NEW.BBoxLow_X,NEW.BBoxLow_Y,NEW.BBoxLow_Z,NEW.BBoxHigh_X,NEW.BBoxHigh_Y,NEW.BBoxHigh_Z))) as bb);END"},
```
rschili, line 352: The expected trigger SQL for `dgn_rtree_upd`/`dgn_rtree_upd1` changed, but the triggers are only ever created in `DgnDbSchema.cpp:365-373` at file creation - I don't think they ever affect existing files.
This may lead to pre-existing iModel now reporting a DDL mismatch during `check_ec_profile`.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
```diff
@@ -3984,6 +4107,18 @@ void GeometricElement3d::_FromJson(BeJsConst props)
             {
             // NOTE: Use the existing bounding box when the GeometryStream is cloned as binary or no geometry exists to calculate from
             m_placement.FromJson(placementJson);
+
+            uint8_t placementDataFlags = PlacementData_None;
+            auto originJson = placementJson[Placement3d::json_origin()];
+            if (placementJson.hasMember(Placement3d::json_origin()) && !originJson.isNull())
+                placementDataFlags |= PlacementData_Origin;
+            // The wire format represents zero angles as null; require a real bbox to distinguish them from a default placement.
```
rschili: Unconfirmed, my copilot raised this:
`!anglesJson.isNull()` is true for anything that is not a JSON null.
`GeometricElement3d.toJSON()` in itwinjs-core (`core/backend/src/Element.ts:829`) hands over a live `YawPitchRollAngles`, which is an empty object for zero angles, not null. So an origin-only element via `iModel.elements.getElement()`, Yaw/Pitch/Roll come back as 0 instead of NULL.

hl662: yeah its legit, I've added a check, so that an empty angle object would lead to NULL now. **Except**, if there is a valid bounding box that has an empty angle object, in that case yaw pitch and roll would still come back as 0 as before

iModelCore/ECDb/ECDb/IntegrityChecker.cpp
hl662: Hmm yeah... the `IntegrityChecker` would report a DDL mismatch. I added a helper method for the 2 dgn_rtree checks for our existing Imodels - so unexpected trigger definitions not part of `dgn_rtree_upd/upd_1` will still be reported, while we skip any rtree_upd mismatch flagged by the checker.

`GeometricElement::_UpdateInDb()` will still handle the updates of the index and the triggers after that check above

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/DgnElement.h
```diff
@@ -2307,6 +2308,31 @@ struct EXPORT_VTABLE_ATTRIBUTE GeometricElement : DgnElement {
...
+
+    uint8_t GetPlacementDataFlags() const
+        {
+        return static_cast<uint8_t>(m_flags.m_placementData);
+        }
+
+    void SetPlacementDataFlags(uint8_t flags)
+        {
+        BeAssert((flags & ~PlacementData_All) == 0); // Accept only flags defined in PlacementDataFlags
+        m_flags.m_placementData = flags;
```
dassaf4, line 2336: The assert is fine, but you still might want to mask off the non-placement bits before assigning `flags`.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
```diff
@@ -2962,43 +2968,52 @@ void dgn_ElementHandler::Geometric3d::_RegisterPropertyAccessors(ECSqlClassInfo&
...
             GeometricElement3d& el = (GeometricElement3d&)elIn;                          \
+            uint8_t placementDataFlags = el.GetPlacementDataFlags();                     \
             Placement3d plc = el.GetPlacement();                                         \
             EXPR;                                                                        \
-            return el.SetPlacement(plc);                                                 \
+            auto status = el.SetPlacement(plc);                                          \
+            if (DgnDbStatus::Success == status)                                          \
+                {                                                                        \
+                uint8_t newPlacementDataFlags = placementDataFlags | FLAGS;              \
+                if (FLAGS == GeometricElement::PlacementData_Bbox && !plc.GetElementBox().IsValid()) \
```
dassaf4: Curious why you aren't calling `Placement2/3d::IsValid` in the SETGEOMPLCPROP macros. The `Placement2/3d::IsValid` methods do more validation on bbox, and they also validate the origin, which is missing here.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
```diff
@@ -3940,21 +4038,41 @@ DgnDbStatus GeometricElement3d::_ReadSelectParams(ECSqlStatement& stmt, ECSqlCla
...
-
-    double yaw      = stmt.GetValueDouble(params.GetSelectIndex(prop_Yaw())),
-           pitch    = stmt.GetValueDouble(params.GetSelectIndex(prop_Pitch())),
-           roll     = stmt.GetValueDouble(params.GetSelectIndex(prop_Roll()));
+    uint8_t placementDataFlags = PlacementData_Origin;
+
+    auto yawIndex = params.GetSelectIndex(prop_Yaw());
+    auto pitchIndex = params.GetSelectIndex(prop_Pitch());
+    auto rollIndex = params.GetSelectIndex(prop_Roll());
+    bool hasAngles = !stmt.IsValueNull(yawIndex) && !stmt.IsValueNull(pitchIndex) && !stmt.IsValueNull(rollIndex);
```
dassaf4, line 4062: Does it ever happen that just one or two angles are null?  If so, this code ignores all three angles. Do we want that?

If we interpret as 0 a null angle that is accompanied by another non-null angle, this would change the semantics of the `PlacementData_Angles` flag from "no angle is null" to "at least one angle is not null and any null angles are understood to be zero".

dassaf4 (approved):
Rob asked me to review the geom bits...

iModelCore/iModelPlatform/PublicAPI/DgnPlatform/DgnElement.h
hl662: Ehh.... `m_placementData` is a 3-bit field, so `GetPlacementDataFlags()` cannot contain non-placement bits. `SetPlacementDataFlags()` also asserts that only PlacementData_All bits are supplied, so an additional mask is not needed here, I tried my best to make sure all call sites and usage of the new placementData flag goes through the helper methods I created

dassaf4: Ok, my only concern was that `BeAssert` not be used to flag a state that needs to be addressed in production. But you're right, only bits 3-5 of `flags` are set in the bitfield; everything else is ignored.

rschili (approved):

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
hl662: `Placement2d/3d::IsValid()` validates the **complete** placement, including requiring a valid BBox.

The macro updates one nullable placement field at a time, so using it here would incorrectly discard legitimate Origin-only or other partial placement states...

So I agree we want to still validate the origin, so I extracted the origin-range logic into `PlacementOnEarth::IsValidOrigin()`. The 2D/3D property setters where you left this comment now validate the origin without calling full Placement::IsValid().

The existing Placement2d/3d::IsValid() methods also delegate to the same new helper, and the new tests cover both dimensions and verify rejected origins do not modify the existing value.

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
hl662: You’re right that the current implementation treats Yaw, Pitch, and Roll as an atomic nullable group at the database boundary. If any of the three persisted angle columns is NULL, `_ReadSelectParams` does not set `PlacementData_Angles`, so the values are omitted from JSON and all three are written back as NULL.

The reported issue has all three angle columns set to NULL, so I’m intentionally keeping independently nullable angles out of this PR. 

Supporting that state would require separate presence tracking and an explicit compatibility decision about how missing angles should be represented, rather than silently converting them to zero or normalizing them to all NULL.

I will say though... that problem is strictly with the data layer. I got this nice table from pi (I can say pi now cuz we just approved pi for harness use :D )

image

So in memory, and over on TS side, we will return the angles as zeroes, not undefined, to the users if thats a concern

dassaf4: Agree on handling individually null YPR angles in another PR. Make it so, Nambot!

iModelCore/iModelPlatform/DgnCore/DgnElement.cpp
dassaf4: I see the validation of origin coord size now. Good! However, bbox _coord size_ is still not validated in the macros. If you want to perform this extra validation, maybe you can break out helpers to fully validate a bbox, to be called here instead of calling `GetElementBox().IsValid()`.

hl662: Good point, I've now added PlacementOnEarth::IsValidBoundingBox() overloads, depending on the type of arg (DRange2d and DRange3d in this case). 

Placement2d::IsValid(), Placement3d::IsValid(), and both SETGEOMPLCPROP macros now use the new shared validation.

The validation here checks for empty ranges and bounding-box spans exceeding the allowed circumference

hl662:
@khanaffan @RohitPtnkr1996

 I added a DgnDb profile migration/forward updater as part of the PR, since you've highlighted that this is a profile change. It's odd that I didn't find previous profile upgrades related to DgnDb, just ECSQL and ECDb....

I followed an approach from `rschili/dgnprofile2008` branch (https://github.com/iTwin/imodel-native/commit/51c7ba05bf10361726910989829157bf13f4f1db). It removes the dormant upgrader registry and introduces a private DgnDb::UpgradeToProfile2_0_0_8 method, explicitly noting that DgnDb currently has only one profile upgrader, and I guess there would be an opportunity to refactor into an actual migration registry once we find more use cases for it?

The migration is gated on profile version 2.0.0.8, replaces the legacy spatial-index update triggers, and advances the stored profile version only after the migration succeeds. I am deliberately not activating WHEN_FIRST_UPGRADER because there is currently no second concrete DgnDb migration that requires ordered execution. If one is added later, we can introduce the generic framework with dedicated ordering, failure, and retry tests.

I added a test to make sure its compatible as well, but I want you 2's opinions on if this is correct...? Or would it warrant a EC profile upgrade too... or maybe fold into https://github.com/iTwin/imodel-native/pull/1488 if we do?

RohitPtnkr1996:
> I added a DgnDb profile migration/forward updater as part of the PR, since you've highlighted that this is a profile change. It's odd that I didn't find previous profile upgrades related to DgnDb, just ECSQL and ECDb....

The DgnDb profile version needs to stay in lockstep with the ECDb version to maintain the whole file format version consistency.
Historically, all our profile version upgrades have been to the ECDb profile version, with the DgnDb version bump as an extra additional change for said consistency.
The imodel evolution tests check this and fail if DgnDb is not incremented alongside.

I don't think we need an ECDb profile upgrade as the triggers here are purely DgnDb-owned artifacts.

That said, I defer to Affan's opinion.
@khanaffan

RohitPtnkr1996 (approved):

aruniverse:
> That said, I defer to Affan's opinion. @khanaffan

@khanaffan plz take a look

JulijaRamoskiene:
@Arun, @khanaffan, @hl662   This is a blocking issue for @nitinjain01 team.

khanaffan:
> @Arun, @khanaffan, @hl662 This is a blocking issue for @nitinjain01 team.

Since we never standalone does DgnDb profile, I am testing out our File Compatibility and SchemaEvolution test. It seem we have to push out artifact to 2.0.0.7 before we can run any test for 2.0.0.8.

khanaffan (approved):

khanaffan:
Note Evolution test were failing for few reasons. PR fixing it is here https://github.com/iTwin/imodel-native-internal/pull/1168

hl662:

PR #1566, opened 2026-08-26
Add native ECDb CSV import support

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9664

## Description

Simulation and analysis engines can generate millions of result rows that must be persisted to ECDb with minimal per-row overhead. This change adds native `NativeECDb.importCSVData` and `NativeECDb.importCSVFile` paths for the corresponding beta iTwin.js APIs.

The implementation:

- streams CSV files through a bounded native buffer;
- decodes V8-serialized in-memory rows without per-row JavaScript/native calls;
- reuses one ECSQL insert statement and rolls back the complete import on failure;
- caches ECSQL binder parameter indexes, removing repeated SQLite parameter-name resolution from the import loop.

### ECSQL binder improvements

Profiling showed repeated named-parameter resolution consuming approximately 28% of active import samples. The following prepared-statement improvements remove that work from repeated insert loops:

- `PrimitiveECSqlBinder` resolves its mapped SQLite parameter index once and reuses it for subsequent bindings.
- `IdECSqlBinder` likewise caches its SQLite parameter index instead of looking it up by name for every row.
- `ECSqlInsertPreparedStatement` caches the generated ECInstanceId binder instead of searching its parameter map on every `Step`.

These caches are scoped to the prepared statement, for which binder identities and parameter indexes remain fixed. In the million-row benchmark, the combined changes reduced insertion time by approximately 40% for `importCSVData` and 42% for `importCSVFile`.

## Performance

Measured with a 119 MB CSV containing 1,000,142 rows:

| Approach | TS read/parse | `v8.serialize()` | Insert | Insert rows/s | End-to-end rows/s |
|---|---:|---:|---:|---:|---:|
| Raw SQLite | 0.739s | — | 0.889s | 1,125,519 | 614,648 |
| `importCSVFile` | — | — | 1.048s | 954,372 | 954,372 |
| `importCSVData` | 0.597s | 0.255s | 1.139s | 877,706 | 502,135 |
| `ECSqlWriteStatement` | 0.733s | — | 16.871s | 59,282 | 56,812 |

## Testing

Added coverage for both import paths, headers, null handling, mapping validation, rollback behavior, and missing files.

Supports iTwin/itwinjs-core#9651


iModelCore/libsrc/v8serial/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 affank
```
tcobbs-bentley: @khanaffan Was this created by you in your personal time? If not, shouldn't this be Bentley?

khanaffan: It's AI generated. Though the idea was mine and had it years ago but did not had time to ever test it. It's based on v8 engine code. But limited support. It's header only. I was not sure where to put the header files. I need to use this api for marshaling concurrent query result and also use for element CRUD. If you can tell me some common place where I can put it I can copy it over there.

tcobbs-bentley: It feels like it's in the right place, but that the MIT license should be Bentley Systems Inc. instead of affank.

khanaffan: Will move it to Bentley lib. It does not need to be third party.

khanaffan: Done move to Bentley repo.

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparedStatement.h
```diff
@@ -429,6 +429,7 @@ struct ECSqlInsertPreparedStatement final : CompoundECSqlPreparedStatement
 
 
         ECInstanceKeyHelper m_ecInstanceKeyHelper;
+        ECSqlBinder* m_generatedIdBinder = nullptr;
```
tcobbs-bentley, line 433: It seems like this is leaking.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -1040,6 +1049,498 @@ Napi::Value JsInterop::DeleteInstance(ECDbR db, NapiInfoCR info) {
     return Napi::Value::From(info.Env(), db.GetModifiedRowCount() > 0);;
 }
 
+namespace {
```
tcobbs-bentley, line 1048: I think this is just continuing to work like we always have, but it seems to me that `JsInterop.cpp` should be kept to code that takes care of the JS interop. In other words, all the work should be done in separate `.cpp`/`.h` files that are appropriate for their task.

Note: I'm not just talking about this namespace, but also about the large function bodies for `ImportCSVFile` and `ImportCSVData`.

If the rest of this file is full of JS interop functions that do lots of work here instead of being split out into more manageable chunks, feel free to ignore and resolve this comment.

iModelCore/ECDb/ECDb/ECSql/ECSqlPreparedStatement.h
khanaffan: This is a non-owning pointer; the primary table statement’s parameter map owns the binder. Added a comment clarifying the ownership.

iModelJsNodeAddon/JsInterop.cpp
khanaffan: Moved the CSV import logic into CsvImporter.cpp/.h. JsInterop.cpp now handles argument decoding and error translation only.

iModelJsNodeAddon/api_package/ts/src/test/ImportCsv.test.ts
```diff
@@ -0,0 +1,354 @@
...
+import { expect } from "chai";
+import * as fs from "fs-extra";
+import * as path from "path";
+import * as v8 from "v8";
+import { DbResult, Guid } from "@itwin/core-bentley";
+import { IModelJsNative } from "../NativeLibrary";
+import { getOutputDir, iModelJsNative } from "./utils";
+
+const testSchemaXml =
+  `<ECSchema schemaName="Test" alias="test" version="01.00.00" xmlns="http://www.bentley.com/schemas/Bentley.ECXML.3.1">
```
rschili: Should be an EC 3.2 schema

tcobbs-bentley (approved):

iModelCore/Bentley/v8serial/PublicAPI/v8serial/writer.hpp
```diff
@@ -0,0 +1,394 @@
...
+#include "v8serial/detail/simd.hpp"
+#include "v8serial/types.hpp"
+
+namespace v8serial {
+
+/// Writes one value using the supported subset of V8 wire format version 15.
+///
+/// The writer owns its byte buffer, has no V8 or N-API dependency, and may be
+/// used on any thread. A single instance is not safe for concurrent access.
+class Writer {
```
rschili, line 25: I'd remove the writer from this PR.

- No production code includes or uses the writer.
- No tests include or exercise it.
- The CSV tests serialize with Node’s v8.serialize(), so they exercise only the new reader.
- No separate library test suite is included.

roughly 500 lines of unused, untested code

iModelJsNodeAddon/CsvImporter.cpp
```diff
@@ -0,0 +1,306 @@
...
+}
+
+uint64_t CsvImporter::ImportData(ECDbR db, Utf8StringCR className, uint8_t const* bytes, size_t byteCount,
+    bvector<CsvImportMapping> const& mapping, CsvImportOptions const& options) {
+    ECSqlStatement statement;
+    const auto plan = prepareCsvImport(statement, db, className, mapping, "Failed to prepare CSV data import ECSQL");
+    const auto bindings = createCsvImportBindings(plan, statement);
+    const Utf8CP nullValue = options.m_nullValue ? options.m_nullValue->c_str() : nullptr;
+
+    v8serial::DecodedValue serializedRowsValue;
```
rschili, line 192: The in-memory path retains a complete native value tree.
A million twelve-column rows require approximately 1.87 GB for row/cell objects alone on 64-bit libstdc++, before strings, the V8 buffer, original JavaScript rows, and SQLite memory.
Can we try to stream rows while preserving repeated-reference support?
Rerun the benchmark with peak memory included.

iModelJsNodeAddon/CsvImporter.cpp
```diff
@@ -0,0 +1,306 @@
...
+                return binding.m_binder->BindBoolean(true);
+            if (value.Equals("false") || value.Equals("0"))
+                return binding.m_binder->BindBoolean(false);
+            return ECSqlStatus::Error;
+        case PRIMITIVETYPE_Double: {
+            if (value.empty())
+                return ECSqlStatus::Error;
+            char* parsedEnd = nullptr;
+            errno = 0;
+            const double parsed = std::strtod(begin, &parsedEnd);
```
rschili, line 164: `strtod` can set `ERANGE` while returning a valid nonzero subnormal, so values such as `1e-310` and `String(Number.MIN_VALUE)` make either CSV API roll back even though ECDb stores those doubles correctly.
We should accept representable subnormal results. Test recommended.

rschili (approved):
Approved but I left some comments that should be handled before merging

iModelCore/Bentley/v8serial/PublicAPI/v8serial/writer.hpp
khanaffan: I need this in for near future. I like to concurrent query to use this api. It will send more data with less memory bandwidth and more performantly. If that all work out then we can use it for other api that send bulk data from native to ts.

khanaffan: It's not untested, it thoroughly tested for compliance with v8. It's copied from another [repo](https://github.com/khanaffan/v8serial) with test. I can bring in test for it as well but its very static library running test all the time will just wast time. 

The issue for me from start was where to place. If we place it as third party then it should be a public repo on itwin like imodel-native. Which I could not create and also not sure if we should create. So I pushed to my profile but then attribution was the problem . So I copied the source code instead. Its at the end really AI extracting a serializer/deserializer with there test from v8 source code. Not sure who takes the credit for doing it. Its not new code.

iModelJsNodeAddon/CsvImporter.cpp
khanaffan: This is know by the caller. They would batch it on there end.

iModelJsNodeAddon/api_package/ts/src/test/ImportCsv.test.ts
khanaffan: Updated the test schema to Bentley.ECXML.3.2.

iModelJsNodeAddon/CsvImporter.cpp
khanaffan: Fixed. We still reject non-finite values and underflow-to-zero, but accept representable subnormals even when strtod sets ERANGE. Added tests for `1e-310` and `String(Number.MIN_VALUE)` on both import APIs.

iModelJsNodeAddon/CsvImporter.cpp
```diff
@@ -0,0 +1,311 @@
...
+        return binding.m_binder->BindNull();
+
+    const auto begin = value.data();
+    const auto end = begin + value.size();
+    switch (binding.m_primitiveType) {
+        case PRIMITIVETYPE_Boolean:
+            if (value.Equals("true") || value.Equals("1"))
+                return binding.m_binder->BindBoolean(true);
+            if (value.Equals("false") || value.Equals("0"))
+                return binding.m_binder->BindBoolean(false);
```
RohitPtnkr1996, line 157: The matching here is case sensitive.
CSV data might contain TRUE/True and FALSE/False.

iModelJsNodeAddon/CsvImporter.cpp
```diff
@@ -0,0 +1,311 @@
...
+uint64_t CsvImporter::ImportData(ECDbR db, Utf8StringCR className, uint8_t const* bytes, size_t byteCount,
+    bvector<CsvImportMapping> const& mapping, CsvImportOptions const& options) {
+    ECSqlStatement statement;
+    const auto plan = prepareCsvImport(statement, db, className, mapping, "Failed to prepare CSV data import ECSQL");
+    const auto bindings = createCsvImportBindings(plan, statement);
+    const Utf8CP nullValue = options.m_nullValue ? options.m_nullValue->c_str() : nullptr;
+
+    v8serial::DecodedValue serializedRowsValue;
+    try {
+        serializedRowsValue = v8serial::Reader(bytes, byteCount).read();
```
RohitPtnkr1996, line 194: With ref to [Rob's comment above](https://github.com/iTwin/imodel-native/pull/1566/changes#r3971859316), I saw the v8serial also provides readRows() to stream the data being read.
Why are we not using that ?

iModelJsNodeAddon/IModelJsNative.mke
```diff
@@ -50,6 +50,9 @@ $(o)imodeljs-nodeaddonapi.package.version.h: $(baseDir)/imodeljs-nodeaddonapi.pa
 
 cIncs + -I$(o)
 
+dirToSearch = $(BuildContext)VendorAPI/
+%include cincapnd.mki
+
```
RohitPtnkr1996, line 55: We no longer need this right?
It's pointing to the whole directory.

iModelCore/BeSQLite/BeSQLite.PartFile.xml
```diff
@@ -6,6 +6,7 @@
     <Part Name="PublicAPI" BentleyBuildMakeFile="BeSQLite.prewire.mke">
         <SubPart PartName="BeRapidJson" PartFile="iModelCore/Bentley/Bentley" />
         <SubPart PartName="PublicAPI"  PartFile="iModelCore/libsrc/google_re2/google_re2" />
+        <SubPart PartName="v8serial" PartFile="iModelCore/Bentley/Bentley" />
```
RohitPtnkr1996, line 9: For now, this should only be in the iModelNodeAddon partfile?

iModelJsNodeAddon/IModelJsNative.mke
khanaffan: We would like to use v8serial for concurrent query to send data from a thread to Node.js.

iModelJsNodeAddon/CsvImporter.cpp
khanaffan: Updated boolean parsing to match true/false case-insensitively, while retaining support for 1/0. Added coverage for TRUE, True, FALSE, and False for both importCSVData and importCSVFile.

iModelCore/BeSQLite/BeSQLite.PartFile.xml
khanaffan: We would like to use v8serial for concurrent query to send data from a thread to Node.js.

iModelJsNodeAddon/CsvImporter.cpp
khanaffan: Updated importCSVData to use readRows() and bind values while decoding, so it no longer materializes the complete native value tree. I also extended readRows() to replay completed dense and sparse row references, preserving the existing repeated-row behavior while retaining only lightweight reference metadata.

pmconne (approved):

PR #1572, opened 2026-08-31
Fix Stack Overflow Sec Issue

DanRod1999:
Based on sec issue found in bug bounty. This is just an attempt from copilot based on the findings of the issue.
 
 # Description:
Fixes unsafe SHX glyph decoding that could allow a crafted embedded font to exhaust the process stack during text layout.

 DecodeSubShape  allocated up to 65,535 attacker-controlled bytes with  _alloca  and recursively decoded referenced subshapes without cycle or depth limits. Source tracing confirmed that the allocation size comes directly from the SHX glyph table and that self-referencing or deeply nested subshapes are reachable through normal glyph layout. Additional review found another recursive path through chained  VERTICAL  opcodes and several malformed-record cases where partial data or missing operands could be consumed.

Changes

• Replaced the variable-sized stack allocation with bounded  ScopedArray  storage.
• Validated locale and Unicode glyph-table entries before caching them.
• Required complete glyph-data reads instead of processing partially initialized buffers.
• Added bounds validation for fixed-size and variable-length opcode operands.
• Rejected active-path subshape cycles while preserving repeated noncyclic references.
• Limited active subshape nesting to 64 levels.
• Limited cumulative decoding to 1 MiB per glyph conversion.
• Reworked  VERTICAL  command processing to consume the following command iteratively rather than recursively.
• Preserved valid glyphs preceding malformed trailing records and applied safe default metrics when font metadata is invalid.
• Added regression coverage for self-references, mutual cycles, excessive nesting, repeated valid subshapes, truncated records and opcodes, unterminated variable-length commands, and long  VERTICAL  chains.

Changes generated with GPT-sol 5.6


iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
```diff
@@ -96,10 +124,14 @@ void ShxFont::LoadNonUnicodeGlyphFPosCacheAndMetrics() {
...
     uint32_t pBufSize = zeroGlyphFPos->second.m_dataSize;
+    if (0 == pBufSize)
+        return;
+
     ScopedArray<Byte> pBuf(pBufSize);
 
     // Read code 0 which should be the font specifier.
-    Read(pBuf.GetData(), pBufSize);
+    if (pBufSize != Read(pBuf.GetData(), pBufSize))
+        return;
```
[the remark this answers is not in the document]
DanRod1999: This seems like a valid correctness issue, but I don't think its related to the changes in this PR. It seems like this isn't really a security concern, so I don't think I want to include these changes in this PR. Also it doesn't look like this is reachable by consumers? Maybe a low priority issue and a new PR is better?

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
```diff
@@ -324,24 +390,34 @@ static bool codeNameExists(CharCP pBytes, size_t maxNameLength) {
 //=======================================================================================
 struct ShapeConverter {
 private:
+    static constexpr size_t MAX_SUBSHAPE_DEPTH = 64;
+    static constexpr size_t MAX_DECODED_BYTES = 1024 * 1024;
```
DanRod1999, line 399: I don't think these limits are based on any real data. Just a best guess of what should be expected. But I'm not sure about the legitimacy of this guess.

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
```diff
@@ -1155,8 +1310,13 @@ void ShapeConverter::DoConversion(size_t numBytes, CharCP buff, bool doVerticals
     numBytes -= iOffset;
     buff += iOffset;
 
-    for (int currentCode = 0; currentCode < (int)numBytes; ++currentCode)
-        DecodeCodes(currentCode, buff);
+    for (int currentCode = 0; currentCode < (int)numBytes; ++currentCode) {
+        bool skipCode = m_skipNextCode;
+        m_skipNextCode = false;
+        AutoRestore<bool> skipCodes(&m_skipCodes, m_skipCodes || skipCode);
+        if (!DecodeCodes(currentCode, buff, numBytes))
```
DanRod1999, line 1322: Can we make the malformed-opcode recovery policy explicit here? Breaking out can return inconsistent partial results: geometry already moved to `m_completedLoops` is preserved, while valid geometry still in `m_activeLineString` is discarded by `GrabCurveVector()`. For example, `{0, 0x10, 8, 1}` decodes the first vector, then rejects the truncated opcode `8`, but returns no curve for that vector because no pen-state command finalized it.

We should either reject the entire top-level glyph when any opcode stream is malformed (my preference for untrusted input, by propagating a conversion failure), or deliberately finalize and preserve the valid prefix. Whichever policy we choose should also define how malformed nested subshapes affect their parent and have a focused test.

*generated with GPT-5.6 Sol*

DanRod1999: Another issue caught be copilot. I also think this is not warranted for this PR, but maybe a future fix. Resolving for now

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
jffmarker: I'm also not sure off-hand what "good" limits are... I think we'd have to throw a bunch of fonts at it to see what to expect?

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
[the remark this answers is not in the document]
jffmarker: A divide by zero is still a crash, so still a security concern? GetDescenderRatio could probably just be patched to check the descender value before dividing...

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
```diff
@@ -952,26 +1102,34 @@ void ShapeConverter::DecodePush() {
 // @bsimethod
 //---------------------------------------------------------------------------------------
 void ShapeConverter::DecodeSubShape(int& currentCode, CharCP pCodes) {
-    uint16_t subshapeCode = pCodes[++currentCode];
-    if (ShxFont::ShxType::Unicode == m_font.GetShxType())
-        subshapeCode = ((subshapeCode << 8) + pCodes[++currentCode]);
+    uint16_t subshapeCode = static_cast<Byte>(pCodes[++currentCode]);
```
jffmarker, line 1105: Why static cast to a Byte but then assign to uint16_t?

DanRod1999:
Using fonts from Jeff that I think are provided in microstation, I was able to view an imodel in dta with my version of native and compare it to our current native version. I can't see any differences between the 2 versions so I believe everything in rendering fine. Here's a few examples:
PR:
image

image

Standard addon:
image

image



Any other ways to verify this change is working as expected outside of this and the tests added to this PR?

jffmarker:
> Any other ways to verify this change is working as expected outside of this and the tests added to this PR?

Checking it visually is much appreciated, I think that's pretty good due diligence.

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
DanRod1999: ran against a battery of fonts, and appears fine. Will assume these values are safe

pmconne:
@DanRod1999 did you delete your .Tiles cache in between runs?

FYI you can [run](https://github.com/iTwin/imodel-native-internal/blob/main/iModelCore/Visualization/TilesetPublisher/docs/image-tests.md) our full ImageTests suite against any branch.

DanRod1999:
> @DanRod1999 did you delete your .Tiles cache in between runs?

I did not. I did run these in two separate instances (worktrees), so I could run 2 dta's at the same time. One with my addon and one with a regular one, but I do believe I tested this first with a stock addon, then linked mine later. So I will retry by clearing the .Tiles cache now.

DanRod1999:
Ok, I checked again and this time used 2 different imodels (copies of one another), so I can view the same content at the same time, and have 2 different .Tiles caches created for each. Still seeing the same results so I think its still good.

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
DanRod1999: pCodes is a cons char*, and if that for some reason it is a negative value coming in, you need to cast to unsigned byte first, and then you can expand it to a 16 bit uint. Otherwise, it will incorrectly calculate the value because it will pad the wrong bits when redefining the size. From everything I read it seems like this is a standard thing to do, but I can't really tell if we'd every hit this edge case

pmconne (approved):

iModelCore/iModelPlatform/DgnCore/ShxFont.cpp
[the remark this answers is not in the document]
DanRod1999: ran a quick check, and this didn't cause a crash. It seems like this will just return a pos inf value. Which although probably not what we want, but not a sec issue from what I can tell. I think making a separate issue for this, and the other comment below is still the better move for now, so we can get this sec fix in

DanRod1999:

PR #1577, opened 2026-09-01
Fixed failing iModelEvolution tests

RohitPtnkr1996:
The recent change to move from jsoncpp to rapidjson caused a few tests to fail.

Fixes: https://github.com/iTwin/itwinjs-backlog/issues/2359


RohitPtnkr1996:
Our PR pipelines do not execute the iModelEvolution tests, which is why this was caught when the bis-schemas PR checks started to fail.
I believe we should update our pipeline to execute the evolution tests, but only for the current runtime and files.

The complete evolution test suite can be deferred as it is time intensive.

In discussion with @naveedkhan8067 about this.

khanaffan (approved):

naveedkhan8067:
> Our PR pipelines do not execute the iModelEvolution tests, which is why this was caught when the bis-schemas PR checks started to fail. I believe we should update our pipeline to execute the evolution tests, but only for the current runtime and files.
> 
> The complete evolution test suite can be deferred as it is time intensive.
> 
> In discussion with @naveedkhan8067 about this.

If it is not going to increase our build time then we should consider adding it and I think it should follow the same pattern of parallel execution as we do it for others.

PR #1578, opened 2026-09-01
Allow GeoCoord tests to run in macOS

### LINKED ISSUE #1406: Low priority: GeoCoord unit tests don't run on macOS
Unit tests for GeoCoord are [configured](https://github.com/iTwin/imodel-native/blob/377e8db83058e9938a8d1f700edb5eeb38c1e57a/iModelCore/GeoCoord/GeoCoord.PartFile.xml#L76) to not run on macOS using `OnlyPlatforms` in the part file. I believe this is due to the Mac build creating static libraries. I tried enabling them and ended up with errors where it was looking for a dynamic library where only a static library existed. Tweaking the `mke` in question led to a later build failure with undefined symbols, and I stopped investigating further.

tcobbs-bentley:
Before these changes, macOS was excluded from the platforms that could run GeoCoord tests due to the fact that the build failed if they were enabled. This enables them and fixes the problems that prevented them from working before.


aruniverse (approved):

tcobbs-bentley:
FYI @ben-polinsky

pmconne (approved):

PR #1580, opened 2026-09-02
Improve native test performance

rschili:
Optimized some tests.

- Cache ECDbs created from identical inline schema XML and clone the immutable seed for each test, preserving per-test isolation.
- Reuse the immutable SchemaSync mock-hub profile seed instead of recreating an identical private seed for every hub.
- Remove unnecessarily long waits from timeout, delay and timestamp tests.
- Clear a clone's existing change-cache sidecar before copying the seed sidecar.
- Explicitly abandon changes after expected failed schema imports.

I recently introduced the `IMODEL_RUN_EXTENDED_TESTS` environment variable, and this PR moves more tests into the extended suite to increase speed of developer and CI builds.

The normal tier keeps representative coverage for each mapping, schema upgrade, remap, SchemaSync conflict and data-shape behavior. Repetitive permutation and scale cases now use `ExtendedTests` fixtures and run when `IMODEL_RUN_EXTENDED_TESTS=1` is set.

## Timings

Measured in full local Debug builds on the same machine. Timings are for the ECDb test executable.

| Run | Before | After | Saved |
| --- | ---: | ---: | ---: |
| Normal/core | 6:23 | 5:04 | 1:19 (20.5%) |
| Full, including extended | 9:24 | 9:01 | 0:23 (4.1%) |

The normal run passed 1,533 tests and skipped 255 extended tests. The full run passed all 1,788 tests.

---

Some measurements after merging this comparing the CI pipeline old vs new:

ECDbTest was faster on every platform that ran it:

| Configuration | Previous |  New | Saved |
| --- | ---: | ---: | ---: |
| Linux ARM64  | 591.9 s  | 388.9 s | 203.0 s / 34.3% 
|  Linux x64     | 309.2 s  | 282.2 s | 27.0 s / 8.7%   
| macOS ARM64   | 82.6 s   | 66.4 s  | 16.2 s / 19.6%  
| Windows MSVC  | 849.6 s  | 842.4 s | 7.2 s / 0.8%    
| Windows Clang | 339.2 s  | 322.8 s | 16.4 s / 4.8%   

 Summed across the five runners, ECDb went from 36:13 to 31:43, saving 4:30 / 12.4% of CI runner time.

 Linux ARM64 is the clearest win: its whole Build step dropped by 226 seconds, of which ECDb accounts for 203 seconds. Windows MSVC is effectively unchanged and deserves another sample before drawing conclusions.

 Why the job timings barely moved

 ECDbTest runs inside the large Azure Build step. Other variations often hid its improvement:

 - Linux x64 ECDb saved 27 seconds, but Pull and initialization became slower.
 - Windows MSVC ECDb saved 7 seconds, while CMake configure became 41 seconds slower.
 - Windows Clang ECDb saved 16 seconds, while Pull and initialization added about 34 seconds.
 - iOS became 80 seconds slower and does not run ECDb.
 - Android saved 6:37, mostly because Pull fell from 5:32 to 0:38. It does not run ECDb.

 Android remained the longest native job at 50:41, so ECDb was not on the native critical path. The change saves machine time without shortening pipeline wall time.

 ECDb is still significant within some jobs:

 - Linux ARM64: about 37% of the job.
 - Windows MSVC: about 37%.
 - Windows Clang: about 19%.
 - Linux x64: about 17%.
 - macOS: about 9%.

### DIFF
--- a/iModelCore/ECDb/Tests/BackDoor/PublicAPI/BackDoor/ECDb/ECDbTestFixture.h
+++ b/iModelCore/ECDb/Tests/BackDoor/PublicAPI/BackDoor/ECDb/ECDbTestFixture.h
@@ -91,6 +91,7 @@ private:
         {
     private:
         bmap<BeFileName, BeFileName> m_seedFilePathsBySchemaFileName;
+        bmap<Utf8String, BeFileName> m_seedFilePathsBySchemaXml;
 
         //not copyable
         SeedECDbManager(SeedECDbManager const&) = delete;
@@ -118,6 +119,24 @@ private:
             return ret.first->second;
             }
 
+        bool TryGetForXml(BeFileName& seedPath, Utf8StringCR schemaXml) const
+            {
+            auto it = m_seedFilePathsBySchemaXml.find(schemaXml);
+            if (it == m_seedFilePathsBySchemaXml.end())
+                return false;
+
+            seedPath = it->second;
+            return true;
+            }
+
+        BeFileNameCR AddForXml(Utf8StringCR schemaXml, BeFileNameCR seedPath)
+            {
+            BeAssert(m_seedFilePathsBySchemaXml.find(schemaXml) == m_seedFilePathsBySchemaXml.end());
+            auto ret = m_seedFilePathsBySchemaXml.insert(bpair<Utf8String, BeFileName>(schemaXml, seedPath));
+            return ret.first->second;
+            }
+
+        size_t GetXmlSeedCount() const { return m_seedFilePathsBySchemaXml.size(); }
         };
 
     static bool s_isInitialized;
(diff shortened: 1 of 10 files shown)


iModelCore/ECDb/Tests/BackDoor/PublicAPI/BackDoor/ECDb/ECDbTestFixture.h
```diff
@@ -118,6 +119,24 @@ struct ECDbTestFixture : public ::testing::Test
             return ret.first->second;
             }
 
+        bool TryGetForXml(BeFileName& seedPath, Utf8StringCR schemaXml) const
```
rschili: This creates 1,029 additional seed databases which consume about 549 MB on disk.
I'll change it so it only does something when the same DB is encountered at least twice.

rschili: New behavior for each schema hash:

 1. First encounter: create the requested test database normally; record only the hash.
 2. Second encounter: create the immutable seed and clone it for the test.
 3. Later encounters: clone the existing seed.

 This means schemas used once no longer produce an additional seed database. The map stores only fixed-size MD5 strings rather than complete schema XML.

khanaffan (approved):

PR #1584, opened 2026-09-08
Prevent duplicate property maps during shared-column remapping

khanaffan:
## Changes

- retain unchanged inherited property mappings when another class frees the same shared column
- scope freed-column invalidation to class/property mappings recorded for data migration
- clear caches and validate persisted mappings after schema mapping and data transforms, preventing corrupt mappings from entering a new changeset
- detect duplicate data-property mappings by class and access string, including conflicting property paths
- tolerate historical incomplete (`ECDb_0160`) and duplicate (`ECDb_0116`) data-property mappings as warnings during changeset apply while keeping schema import strict
- add regressions covering shared-column remapping, persisted-state validation, incomplete historical maps, duplicate historical maps, ECSQL preparation, value preservation, and reopen

## Related issues

- iTwin/itwinjs-backlog#2374
- iTwin/itwinjs-backlog#2403


iModelCore/ECDb/ECDb/DbMapValidator.cpp
```diff
@@ -168,7 +168,7 @@ BentleyStatus DbMapValidator::CheckDuplicateDataPropertyMap() const {
                 JOIN [ec_Column] [c] ON [c].[Id] = [pp].[ColumnId]
                 JOIN [ec_Table] [t] ON [t].[Id] = [c].[TableId]
             WHERE  [p].[AccessString] != 'ECClassId' AND [p].[AccessString] != 'ECInstanceId'
-            GROUP BY [pp].[ClassId], [pp].[PropertyPathId] HAVING COUNT (*) > 1;)");
+            GROUP BY [pp].[ClassId], [p].[AccessString] HAVING COUNT (*) > 1;)");
```
rschili, line 171: Grouping by access string introduces a new database-wide hard error for existing duplicate paths, so an unrelated schema import—or replay of a historical schema changeset containing these mappings can fail even when this import did not create the duplicates. 

We may want to keep the prevention fix, but defer this stronger validation or distinguish newly introduced duplicates from existing ones until a repair is available.

This is the same rollout concern that held up #1529, applied to a different corruption shape.

Supporting paths:

- `DbMapValidator.cpp:161–193`: scans all persisted mappings, reports `ECDb_0116`, and returns `ERROR`; there is no restriction to changed classes or exception for changeset replay.
- `SchemaManagerDispatcher.cpp:1495`: propagates that error from schema import.
- `ECDb.cpp:113–122` → `SchemaManagerDispatcher.cpp:54`: schema changeset replay runs the same validator with `ChangesetApply`, which this check does not consult.
- `TxnManager.cpp:2587–2594`: aborts the merge when that validation fails.

rschili (approved):
Looks good but left one concern as comment.

iModelCore/ECDb/ECDb/DbMapValidator.cpp
khanaffan: The issue is we cannot really repair it in place. You need a LastKnowGood iModel for reference to fix duplicate property map. The validation should at least catch and prevent this issue. While the import schema fix avoid the problem. 

Current there are two approach for existing iModels with this issue
1. Find last know good and revert to it. And then run connector with this PR merged. 
2. Use [itwin-tools](https://github.com/iTwin/itwinjs-backlog/issues/2374#issuecomment-5603284613) to repair exiting imodel. It will auto find LKG and use tip and push a changeset that will remove duplicate properties.

RohitPtnkr1996 (approved):

pmconne (approved):

iModelCore/ECDb/ECDb/ClassMap.cpp
```diff
@@ -557,10 +557,14 @@ BentleyStatus ClassMap::Update(SchemaImportContext& ctx)
             if (propMap == nullptr || !propMap->IsData() || propMap->GetType() == PropertyMap::Type::Navigation)
                 continue;
 
+            // A sibling may have freed this shared column without changing this class's mapping.
+            if (!ctx.RemapManager().HasCleanedPropertyMapping(m_ecClass.GetId(), property->GetName()))
```
rschili, line 561: https://github.com/iTwin/imodel-native/pull/1591
The test I added is fixed by this exact change. We should merge these together! This is very likely the current MstnConnector bug.

iModelCore/ECDb/ECDb/SchemaManagerDispatcher.cpp
```diff
@@ -1544,10 +1540,50 @@ SchemaImportResult MainSchemaManager::MapSchemas(SchemaImportContext& ctx, bvect
...
+    }
+
     ClearCache();
     return  SchemaImportResult::OK;
 }
 
+//---------------------------------------------------------------------------------------
+// @bsimethod
+//+---------------+---------------+---------------+---------------+---------------+------
+BentleyStatus MainSchemaManager::ValidatePersistedMappings(SchemaManager::SchemaImportOptions options) const
```
RohitPtnkr1996, line 1554: I am wondering about the performance impact of this, especially on large iModels.
Even a tiny schema import will force load everything.

Should this be limited to when remapping has occured?

rschili: Yeah I thought about this, my rationale being I think doing a full check here may cost a few seconds, but letting something slip costs us much more.
@khanaffan probably knows how much this costs, my understanding is that it should be fairly cheap even on large iModels.

khanaffan: I think performance a side we cannot skip verification of persisted map. It's causing incorrect map to be pushed to hub. Performance can be improved if it's slow. 

I am also adding following so we can invoke it without import schema and see errors in existing iModels.

```sql
PRAGMA validate_persisted_mappings OPTIONS enable_experimental_features;
```

khanaffan: I measured the new `validate_persisted_mappings` pragma, which invokes the same `ValidatePersistedMappings` path, against two large production-derived iModels:

| Database | Logical size | Cold | Warm runs | Max RSS |
|---|---:|---:|---:|---:|
| Changeset 68030 clone | 46.9 GB | 1.901 s | 1.132 s, 1.159 s | ~247 MiB |
| Pre-repair affected clone | 39 GB | 1.445 s | 1.287 s, 1.338 s | ~251 MiB |

Each measurement used a fresh `iModelConsole` process and opened the file read-only. The reported pragma time includes statement preparation, where validation actually runs, plus result iteration. Total process wall time was 1.14–1.94 s, so console startup/open overhead was negligible.

Both databases contained mapping problems and validation stopped at the first reported error (`ECDb_0197`/`ECDb_0198` in the first clone and `ECDb_0160` in the second). Since the validator is currently fail-fast, these numbers do not represent scanning past every possible inconsistency. Based on this sample, the full persisted-map check costs roughly 1–2 seconds on 39–47 GB iModels.

rschili:
https://github.com/Mergifyio backport release/5.13.x

PR #1589, opened 2026-09-09
Allow SqliteChangesetReader groups to use SQLiteDb

khanaffan:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/9708

## Changes

- allow native `SqliteChangesetReader.openGroup` to accept a generic `SQLiteDb`
- widen the TypeScript declaration from `AnyECDb` to `AnyDb`
- validate that every referenced changeset table exists and has enough columns before constructing the change group
- add multi-file grouped-output, missing-schema, `DgnDb`, and `ECDb` coverage

## Validation

The focused group suite processes 342 changes across two files and completes in 38 ms test time (457 ms including Mocha startup) on macOS ARM64.

## Cross-repository scope

This is the native prerequisite for exposing `SQLiteDb` through `SqliteChangesetReader.openGroup` in `@itwin/core-backend`. The linked core PR currently exposes `SQLiteDb` through `openFile`; `openGroup` can follow after this native change is released and consumed.

Supports iTwin/itwinjs-core#9701.


iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -1414,13 +1414,33 @@ void SqliteChangesetReader::OpenChangeStream(Napi::Env env, std::unique_ptr<Chan
...
+        for (auto change = changes.begin(); change.IsValid(); ++change) {
+            Utf8CP tableName;
+            int columnCount;
+            DbOpcode opcode;
+            int indirect;
+            if (BE_SQLITE_OK != change.GetOperation(&tableName, &columnCount, &opcode, &indirect))
+                THROW_JS_BE_SQLITE_EXCEPTION(env, "openGroup(): unable to read changeset", BE_SQLITE_ERROR);
+
+            if (checkedTables.insert(tableName).second) {
+                if (!db.TableExists(tableName))
```
rschili: TableExists seems to be case sensitive, so a changeset for bis_Element is rejected when the schema db declares BIS_ELEMENT even though SQLite resolves the table correctly. Could use a bound metadata query with SQLite's identifier lookup semantics, as the TypeScript reader does.

iModelJsNodeAddon/JsInterop.cpp
```diff
@@ -1414,13 +1414,33 @@ void SqliteChangesetReader::OpenChangeStream(Napi::Env env, std::unique_ptr<Chan
...
     DdlChanges ddlGroup;
+    bset<Utf8String> checkedTables;
     for(auto& changesetFile : changesetFiles) {
         BeFileName inputFile(changesetFile);
         if (!inputFile.DoesPathExist()) {
             THROW_JS_BE_SQLITE_EXCEPTION(env, SqlPrintfString("openGroup(): changeset file specified does not exists (%s)", inputFile.GetNameUtf8().c_str()), BE_SQLITE_CANTOPEN);
         }
 
         ChangesetFileReader reader(inputFile);
+        Changes changes(reader, false);
```
rschili, line 1549: Could we instead of iterating twice, move the checks into the grouping pass?

rschili (approved):

iModelJsNodeAddon/JsInterop.cpp
khanaffan: TableExits() is not case sensitive. It use SQLite prepare which is case insensitive. 

```cpp
bool Db::TableExists(Utf8CP tableName) const
    {
    Statement statement;
    return BE_SQLITE_OK == statement.TryPrepare(*this, SqlPrintfString("SELECT NULL FROM %s", tableName));
    }

```

iModelJsNodeAddon/JsInterop.cpp
khanaffan: You right let me improve this code

iModelJsNodeAddon/JsInterop.cpp
khanaffan: You are correct about the required behavior, but I rechecked the current implementation and `Db::TableExists()` queries `sqlite_master.name = ?`, which is case-sensitive. I replaced it with `Db::GetColumns()`, whose `PRAGMA table_info` lookup follows SQLite identifier semantics, and added coverage using uppercase physical table names. Fixed in bc1f6aa9d.

iModelJsNodeAddon/JsInterop.cpp
khanaffan: I investigated combining these passes, but the existing `Changes` iterator does not expose terminal stream errors: malformed or truncated input is indistinguishable from normal end-of-stream. Manually calling `ChangeGroup::AddChange()` during that iteration could therefore silently accept a corrupt changeset. Keeping `AddToChangeGroup()` preserves `sqlite3changegroup_add_strm` error propagation, so I retained the two passes for correctness.

pmconne (approved):

PR #1590, opened 2026-09-09
(bug): Repair spatial-index trigger replay with profile 2.0.0.9

hl662:
## TL;DR

The DgnDb 2.0.0.8 upgrade changed spatial-index triggers locally without recording those changes in its changesets. Other briefcases received the new profile version but kept old triggers, causing the core schema synchronization integration tests to fail.

Profile 2.0.0.9 replaces both triggers using `ExecuteDdl`, so the changeset carries the updated triggers. It repeats the replacement even when the upgrading file already has correct triggers, because other briefcases that were on 2.0.0.8 may still need the repair.

The investigation also found that splitting SQL at every semicolon broke trigger bodies and could alter quoted values. Statement splitting now uses `sqlite3_complete()` to preserve complete SQL statements.

We also found some mapped-table DDL failures are intentionally tolerated so the receiving briefcase can reconstruct the physical tables from `ec_*` metadata. This PR preserves that recovery behavior rather than introducing a broader replay-error policy...

## What

| Asset | Why it exists |
|---|---|
| `iModelCore/iModelPlatform/DgnCore/DgnDbSchema.cpp` | `ExecuteSql` did not record the trigger DDL for changesets; `ExecuteDdl` records all four statements when tracking is active. |
| `iModelCore/iModelPlatform/Tests/DgnProject/NonPublished/DgnProject_Test.cpp` | The existing test checked only the locally upgraded file; it now checks the serialized DDL and compares both trigger definitions after replay into a pre-upgrade copy. |

The shared trigger-creation helper always uses `ExecuteDdl`. Tracking is disabled during ordinary new-file creation, so those definitions are not recorded as changeset DDL. No global SQL or replay APIs changed.

> IMPORTANT: This repairs trigger definitions in already-stamped 2.0.0.8 files through profile upgrade or changeset replay. We can only do that through another DgnDb profile upgrade to 2.0.0.9.

## Tests

<details>
<summary>New and updated test coverage</summary>

| Test | What it covers |
|---|---|
| `DgnDbTest.UpgradeSpatialIndexTriggers` | Both trigger DROP/CREATE statements appear in the generated changeset, and replay produces the same profile version and trigger definitions as the upgraded source. |

</details>

---
*Nam and Nambot 🤖 (powered by gpt-6-astra)*


hl662:

khanaffan (approved):

tcobbs-bentley (approved):

pmconne (approved):

aruniverse:
Why was this bug/regression not caught in our profile version tests?

cc @RohitPtnkr1996 @rschili

RohitPtnkr1996:
> Why was this bug/regression not caught in our profile version tests?

The compatibility/iModelEvolution tests in imodel-native run upgrade permutations on local files only, and no changesets are ever produced or applied.
There are no round-trip tests here, and no DDL assertions as well.
The tests only assert version numbers, open/upgrade modes, and schema contents.
I will update the coverage to include the above.

Also, the profile version tests (over in their own repo) are terribly outdated (and have been for quite some time), all the supposed versions have reached their EOL.
I've got an active PR to restructure the entire repo and update the test coverage.

@aruniverse @rschili

rschili (approved):

hl662:
I added 2 more fixes to pre existing bugs uncovered while working on the PR:

 1. Existing DDL processing was splitting SQL at every semicolon. That broke complete trigger bodies and could silently alter quoted values such as DEFAULT 'a; b'. The integration tests for schemaSync caught this problem. So now changeset grouping, Schema synchronization replay, and index-DDL rewriting now uses a native `sqlite3_complete()` method to do splitting, that does preserve statement contents.

2.  Applying a changeset could report success despite failed DDL schema changes. For example, dropping an index could succeed while creating its replacement failed. Native replay logged the error but returned success, allowing the caller to continue and mark the changeset as applied with an incomplete Sqlite schema **(Not ECSchema this doesn't have anything to do with our current ec schema work).** Replay now propagates unrecovered failures so our TypeScript apply workflow in itwinjs-core can catch the error and roll back the partial changes via `abandonChanges()`, as it's already doing

PR #1596, opened 2026-09-15
Fix SQLite bug applying UPDATE with a unique constraint.

kring:
This is a patch from upstream: https://sqlite.org/src/info/8b3da5d6cea2e3b3

Here's how I originally described it when I reported it to the SQLite developers:

> Hi SQLite team,
>
> I recently ran into what I think is a bug in the Session extension.
>
> The scenario is: we’re applying a changeset that includes a row update, and that update legitimately causes a UNIQUE constraint violation. The expectation is that this will be raised to the conflict handler.
>
> Instead, here is what happens.
>
> The process ends up in sessionRetryConstraints, as expected. Step (1) concludes without being able to make progress, because this is a true violation. Step (2) tries to delete the row and re-insert it, but that doesn’t work either.
>
> The attempted insert calls sqlite3_step/sqlite3_finalize, and then checks the return code to see if we still have a constraint violation. It checks like this:
>
> ```
>       sqlite3_step(pInsert);
>       rc = sqlite3_finalize(pInsert);
>       if( rc==SQLITE_CONSTRAINT ){
>         rc = sqlite3_exec(db, "ROLLBACK TO update_op", 0, 0, 0);
>         sqlite3_free(pApply->constraints.aBuf);
>         pApply->constraints = cons;
>         memset(&cons, 0, sizeof(cons));
>       }
> ```
>
> The bug is that comparison to SQLITE_CONSTRAINT. In the case of a UNIQUE constraint violation, at least, the value of rc is SQLITE_CONSTRAINT_UNIQUE. So I believe the check should instead be:
>
> ```
>       if( (rc&0xff)==SQLITE_CONSTRAINT ){
> ```
>
> I’ve confirmed this works in my use-cases, at least.

The developers confirmed my analysis and adopted the proposed fix.

A test is coming, but I wanted to get this open right away.


rschili (approved):

aruniverse:
We'll want to cherry pick this to imodel02. cc @khanaffan @rschili

kring:
Any idea what's going on with the Android build? This is the error, as far as I can tell:

> 2026-09-15T11:59:58.6695069Z FAILURE: Build failed with an exception.
2026-09-15T11:59:58.6695205Z 
2026-09-15T11:59:58.6695426Z * What went wrong:
2026-09-15T11:59:58.6695753Z Execution failed for task ':app:packageDebug'.
2026-09-15T11:59:58.6696080Z > A failure occurred while executing com.android.build.gradle.tasks.PackageAndroidArtifact$IncrementalSplitterRunnable
2026-09-15T11:59:58.6696401Z    > java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError

DanRod1999:
re queued android build. I think its transient failure since everything else is passing

kring:
> re queued android build.

For my future reference, is that something I can do myself?

aruniverse:

PR #1604, opened 2026-09-17
Add `GeoServices.getListOfVerticalCRS` and implement Vertical CRS imodel persistence

tcobbs-bentley:
## Summary

- Add `GeoServices.getListOfVerticalCRS` with optional point and extent filtering.
- Support named Vertical CRS interpretation, conversion, and iModel persistence while retaining legacy Type 66 compatibility and rejecting stale named metadata.
- Add coverage for workspace-backed definitions and grids, filtering and validation, round-tripping, conversion, persistence, and missing resources.

## Testing

- GeoServices TypeScript tests
- GeoCoord and iModelPlatform builds/tests

### DIFF
--- a/iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
+++ b/iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
@@ -9207,6 +9207,23 @@ public:
     }
 };
 
+static Utf8String GetVerticalGridFilePath(WStringCR gridFile, WStringCR dataDirectory)
+{
+    BeFileName fileName(gridFile);
+    if (!fileName.IsAbsolutePath())
+    {
+        BeFileName resourcePath(L"assets");
+        resourcePath.AppendToPath(gridFile.c_str());
+        return resourcePath.GetNameUtf8();
+    }
+
+    WString resolvedGridFilePath;
+    if (SUCCESS != BeFileName::ResolveRelativePath(resolvedGridFilePath, gridFile.c_str(), dataDirectory.c_str()))
+        return "";
+
+    return Utf8String(resolvedGridFilePath.c_str());
+}
+
 class VerticalGeoidSeparationGridTransform : public VerticalTransform
 {
     friend class VerticalTransform;
@@ -9407,11 +9424,9 @@ public:
         {
             if (0 != gridFile.length())
             {
-                WString resolvedGridFilePath;
-                
-                if (SUCCESS == BeFileName::ResolveRelativePath(resolvedGridFilePath, gridFile.c_str(), dataDirectory.c_str()))
+                Utf8String resolved = GetVerticalGridFilePath(gridFile, dataDirectory);
+                if (!resolved.empty())
                 {
-                    Utf8String resolved(resolvedGridFilePath.c_str());
                     csDatumCatalogEntry_* entry = CSnewDatumCatalogEntry2 (resolved.c_str(), // fullpath
                                                                         0,      // path is not relative
                                                                         0,      // buffer size used when streaming, will be overloaded by size of record entry if available and that is larger
@@ -9638,11 +9653,9 @@ public:
         {
             if (0 != gridFile.length())
             {
-                WString resolvedGridFilePath;
-                if (SUCCESS == BeFileName::ResolveRelativePath(resolvedGridFilePath, gridFile.c_str(), dataDirectory.c_str()))
+                Utf8String resolved = GetVerticalGridFilePath(gridFile, dataDirectory);
+                if (!resolved.empty())
                 {
-                    Utf8String resolved(resolvedGridFilePath.c_str());
-
                     csDatumCatalogEntry_* entry = CSnewDatumCatalogEntry2 (resolved.c_str(), // fullpath
                                                                         0,      // path is not relative
                                                                         0,      // buffer size used when streaming, will be overloaded by size of record entry if available and that is larger
@@ -10574,21 +10587,33 @@ StatusInt VerticalDatumDictionary::AddVerticalDatumsFromFile(const WString& file
     if (0 == filepath.length())
         return GEOCOORDERR_BadArg;
 
-    BeFile dictionaryFile;
-    if (BeFileStatus::Success != dictionaryFile.Open(filepath.c_str(), BeFileAccess::Read))
+    Utf8String filePathUtf8(filepath.c_str());
+    csFILE* dictionaryFile = CS_fopen(filePathUtf8.c_str(), "rb");
+    if (nullptr == dictionaryFile)
         return GeoCoordParse_MissingFile;
 
     s_verticalDatumDictionary->m_dictionaryPath = filepath;
 
-    bvector<Byte> dictionary;
-    if (BeFileStatus::Success != dictionaryFile.ReadEntireFile(dictionary) 
-        || (0 == dictionary.size())
-        || (nullptr == dictionary.data()))
+    if (0 != CS_fseek(dictionaryFile, 0, SEEK_END))
+    {
+        CS_fclose(dictionaryFile);
+        return GeoCoordParse_ReadError;
+    }
+
+    long dictionarySize = CS_ftell(dictionaryFile);
+    if (dictionarySize <= 0 || 0 != CS_fseek(dictionaryFile, 0, SEEK_SET))
+    {
+        CS_fclose(dictionaryFile);
+        return GeoCoordParse_ReadError;
+    }
+
+    bvector<Byte> dictionary((size_t)dictionarySize + 1);
+    size_t bytesRead = CS_fread(dictionary.data(), 1, (size_t)dictionarySize, dictionaryFile);
+    CS_fclose(dictionaryFile);
+    if (bytesRead != (size_t)dictionarySize)
         return GeoCoordParse_ReadError;
 
-    // Ensure dictionary is null terminated
-    dictionary.resize(dictionary.size() + 1);
-    dictionary[dictionary.size() - 1] = '\0';
+    dictionary[(size_t)dictionarySize] = '\0';
 
     return AddVerticalDatumsFromJsonString(reinterpret_cast<char*>(dictionary.data()));
 }
@@ -12252,6 +12277,16 @@ struct WorkspaceDb : RefCountedBase, NonCopyableClass {
         }
         return m_db->IsDbOpen() ? m_db : nullptr;
     }
+    bool ContainsResource(Utf8CP resourceName) {
+        auto db = GetDb();
+        if (nullptr == db)
+            return false;
+
+        Statement stmt;
+        stmt.Prepare(*db, "SELECT 1 FROM blobs WHERE id=? COLLATE NOCASE");
+        stmt.BindText(1, resourceName, Statement::MakeCopy::No);
+        return BE_SQLITE_ROW == stmt.Step();
+    }
     WorkspaceDb(int priority, Utf8StringCR dbName, CloudContainerP container) : m_priority(priority), m_dbName(dbName), m_container(container) {
         if (m_container) {
             // set up a listener for the container being disconnected and remove this entry (closes Db)
@@ -12403,10 +12438,14 @@ bool BaseGCS::AddWorkspaceDb(Utf8String dbName, CloudContainerP container, int p
     for (auto it=s_workspaceDbs.begin(); it != s_workspaceDbs.end(); ++it) {
         if (priority > (*it)->m_priority) {
             s_workspaceDbs.emplace(it, newDb);
+            if (VerticalDatumDictionary::Get().IsValid() && newDb->ContainsResource("VerticalDatumDefinitions.json"))
+                VerticalDatumDictionary::ClearAndReinitialize();
             return true;
         }
     }
     s_workspaceDbs.emplace_back(newDb);
+    if (VerticalDatumDictionary::Get().IsValid() && newDb->ContainsResource("VerticalDatumDefinitions.json"))
+        VerticalDatumDictionary::ClearAndReinitialize();
     return true;
 }
 
@@ -12427,14 +12466,6 @@ StatusInt BaseGCS::Initialize(Utf8CP dataDirectory) {
     ::CS_altdr(s_assetsDirPrefix.c_str());
     s_assetsDir = dataDirectory;
 
-    // Initialize vertical datum dictionary
-    if (!VerticalDatumDictionary::Get().IsValid())
-        {
-        BeFileName vdatumFileName(dataDirectory);
-        vdatumFileName.AppendToPath(L"VerticalDatumDefinitions.json");
-        VerticalDatumDictionary::Initialize(vdatumFileName, WString(dataDirectory, true));
-        }
-
 #if defined (BENTLEY_WIN32)||defined (BENTLEY_WINRT)
     if (s_assetsDir.StartsWith("\\\\?\\")) // fopen doesn't work correctly with long path prefix on Windows
         s_assetsDir = s_assetsDir.substr(4);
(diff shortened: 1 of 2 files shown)


iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
```diff
@@ -9629,9 +9623,6 @@ class VerticalOffsetGridTransform : public VerticalTransform
         if (nullptr != m_vertconUS)
             return SUCCESS;
 
-        if (0 != CSvrtconInit())
-            return REPROJECT_CSMAPERR_VerticalDatumConversionError;
```
AlainRobertAtBentley, line 9633: I do not understand why the CSvrtconInit() initialization was removed.

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
```diff
@@ -5346,6 +5347,28 @@ DgnGCSP         DgnGCS::FromProject(DgnDbR project)
...
+    Utf8String verticalCrsJson;
+    if (BeSQLite::BE_SQLITE_ROW == project.QueryProperty(verticalCrsJson, DgnProjectProperty::DgnGCSVerticalCRS()))
+        {
+        BeJsDocument verticalCrs(verticalCrsJson);
+        Utf8String type66VerticalDatum;
+        gcs->GetVerticalDatumName(type66VerticalDatum);
+        MD5 type66Hasher;
+        Utf8String type66Hash = type66Hasher(buffer.GetData(), propSize);
+        // An older writer can update Type 66 without updating this property, so only use named metadata written with the current Type 66 payload.
+        if (verticalCrs["id"].isString() && 0 == type66VerticalDatum.CompareToI(verticalCrs["id"].asString()) &&
```
AlainRobertAtBentley, line 5359: The main identifier for a VerticalCRS is the property 'crsName'. The property 'id' is only to back support previous definitions. The call to FromVerticalJson() will only use the 'id' property if the rest of the properties do not define a valid VerticalCRS.

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
```diff
@@ -5430,6 +5453,28 @@ StatusInt       DgnGCS::Store(DgnDbR project)
...
+            Utf8String crsName;
+            Utf8String id;
+            GetFullVerticalDatumName(crsName);
+            GetVerticalDatumName(id);
+            verticalCrs["crsName"] = crsName;
+            verticalCrs["id"] = id;
+            // Correlate this named metadata with the Type 66 payload saved above so readers can detect an independent Type 66 update.
+            MD5 type66Hasher;
+            verticalCrs["type66Hash"] = type66Hasher(type66AppData, type66AppDataBytes);
+            status = project.SavePropertyString(DgnProjectProperty::DgnGCSVerticalCRS(), verticalCrs.Stringify()) == BeSQLite::BE_SQLITE_OK ? SUCCESS : ERROR;
```
AlainRobertAtBentley, line 5470: We will want to support user-defined VerticalCRS so storing the identifier properties is insufficient. The complete definition should be stored at all time. The FromVerticalJson() will make use of the 'crsName', 'epsg' or 'id' properties to locate a match in the predefined list, but if not then the full definition must remain valid and useable. NOTE: I am not sure the library currently fully supports user-defined VerticalCRS but it is being implemented in PowerPlatform so it will be supported soon.

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
```diff
@@ -5346,6 +5347,28 @@ DgnGCSP         DgnGCS::FromProject(DgnDbR project)
...
+    if (BeSQLite::BE_SQLITE_ROW == project.QueryProperty(verticalCrsJson, DgnProjectProperty::DgnGCSVerticalCRS()))
+        {
+        BeJsDocument verticalCrs(verticalCrsJson);
+        Utf8String type66VerticalDatum;
+        gcs->GetVerticalDatumName(type66VerticalDatum);
+        MD5 type66Hasher;
+        Utf8String type66Hash = type66Hasher(buffer.GetData(), propSize);
+        // An older writer can update Type 66 without updating this property, so only use named metadata written with the current Type 66 payload.
+        if (verticalCrs["id"].isString() && 0 == type66VerticalDatum.CompareToI(verticalCrs["id"].asString()) &&
+            verticalCrs["type66Hash"].isString() && type66Hash.Equals(verticalCrs["type66Hash"].asString()))
```
AlainRobertAtBentley, line 5360: Just a thought and this can be implemented in a later Pull Request. Currently the type66 binary storage has many disadvantages including the fact it is binary, opaque and non-flexible. There exists HorizontalCRS definitions that can be represented using the JSON definition that cannot be converted to a type66. Since it is possible to set the CRS from itwinjs using a JSON definition but we cannot guarantee this definition can be stored in the existing mechanism I wonder if we should not migrate the whole CRS definition storage to a full JSON storage, the type66 being used only if the new property does not exist (and storing the type66 as we can for backward compatibilty only)

iModelJsNodeAddon/IModelJsNative.h
```diff
@@ -660,10 +660,22 @@ struct CRSListResponseProps
     Utf8String m_unit;
     };
 
+struct VerticalCRSListResponseProps
+    {
+    Utf8String m_crsName;
```
AlainRobertAtBentley, line 665: I would add the 'epsg' property since it is an alternate identifier and users may want to select based on this value.

iModelJsNodeAddon/GeoServicesInterop.cpp
```diff
@@ -120,3 +134,125 @@ bvector<CRSListResponseProps> GeoServicesInterop::GetListOfCRS(DRange2dCP extent
...
+            return status;
+
+        VerticalCRSListResponseProps verticalCrs;
+        verticalCrs.m_crsName = name;
+        info->GetDescription(verticalCrs.m_description);
+        verticalCrs.m_deprecated = info->IsDeprecated();
+        info->GetType(verticalCrs.m_type);
+        info->GetUnits(verticalCrs.m_unit);
+        info->GetExtent(verticalCrs.m_extent);
+        verticalCrs.m_id = verticalCrs.m_type;
```
AlainRobertAtBentley, line 253: id and type are not the same. id is the former identifier that was limited to 5 distinct value (NGVD29, NAVD88, GEOID,...) while type indicated the fact the VerticalCRS is mathematically based (ELLIPSOID) or gravitic (GEOID). We want to eventually get rid of the 'id' property.
Since the id property is mandatory in the itwinjs definition (for backward compatibility) I would propagate it regardless it will not be used in the next version.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
```diff
@@ -856,11 +856,30 @@ export declare namespace IModelJsNative {
...
+  interface VerticalCRSListProps {
+    point?: { longitude: number, latitude: number };
+    extent?: Range2dProps;
+    includeIntersecting?: boolean;
+  }
+
+  /** Expected to migrate to `@itwin/core-common` with `VerticalCRSListProps`. */
+  interface VerticalCRSListEntry {
+    crsName: string;
+    id: "GEOID" | "ELLIPSOID" | "NGVD29" | "NAVD88" | "LOCAL_ELLIPSOID";
```
AlainRobertAtBentley, line 869: See other comments about adding the 'epsg' property.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
```diff
@@ -0,0 +1,344 @@
...
+    const statement = new iModelJsNative.SqliteStatement();
+    statement.prepare(workspaceDb, "CREATE TABLE blobs(id TEXT PRIMARY KEY NOT NULL, value BLOB)");
+    statement.step();
+    statement.dispose();
+
+    statement.prepare(workspaceDb, "INSERT INTO blobs(id,value) VALUES(?,?)");
+    const egm96Grid = fs.readFileSync(path.join(csMapDataDir, "WW15MGH._96"));
+    const resources = new Map<string, Buffer>([
+      ["VerticalDatumDefinitions.json", Buffer.from(verticalDatumDictionary)],
+      ["World/WW15MGH.GRD", egm96Grid],
```
AlainRobertAtBentley, line 91: The WW15MGH._96 file is distributed but the file WW15MGH.GRD should not be distributed. It may be in previous csmap_data packages but it is only for backward compatibility. The system first checks the presence of the file of WW15MGH._96 and will not need the other if located. Consider this file(...GRD) does not exist.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
```diff
@@ -0,0 +1,344 @@
...
+    expect(egm96.extent).to.deep.equal({ low: [-180, -90], high: [180, 90] });
+    expect(verticalSystems.some((entry) => entry.crsName === "Regional test height")).to.be.true;
+  });
+
+  it("interprets a named vertical coordinate reference system with its fallback id", () => {
+    const response = iModelJsNative.GeoServices.getGeographicCRSInterpretation({
+      format: "JSON",
+      geographicCRSDef: JSON.stringify({
+        horizontalCRS: { id: "LL84" },
+        verticalCRS: { crsName: "EGM96 height", id: "GEOID" },
```
AlainRobertAtBentley, line 144: the id property is mandatory but only not to break the former API. If crsName is specified then the value of id can be ignored. (not tested below).

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
```diff
@@ -0,0 +1,344 @@
...
+        true,
+      );
+      if (typeof verticalCrsProperty !== "string")
+        assert.fail("Expected the named vertical CRS file property to exist as a string");
+
+      iModelDb.updateIModelProps({
+        rootSubject: { name: "Stale named Vertical CRS: IGNORED BY updateIModelProps" },
+        geographicCoordinateSystem: {
+          horizontalCRS: { id: "LL84" },
+          verticalCRS: { id: "ELLIPSOID" },
```
AlainRobertAtBentley, line 242: Using this test is a good way to test backward compatibility. The equivalent new identifier would have 'crsName' set to 'WGS84'. Just mentioning since testing with only the id property is a valid.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
```diff
@@ -0,0 +1,344 @@
...
+      fs.rmSync(iModelPath, { force: true });
+    }
+  });
+
+  it("rejects an unknown named vertical coordinate reference system", () => {
+    const response = iModelJsNative.GeoServices.getGeographicCRSInterpretation({
+      format: "JSON",
+      geographicCRSDef: JSON.stringify({
+        horizontalCRS: { id: "LL84" },
+        verticalCRS: { crsName: "Unknown test height" },
```
AlainRobertAtBentley, line 268: The id property is mandatory and must be specified whatever it contains.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
AlainRobertAtBentley: I located the comment indicating this call was redundant but I only located calls in code paths concerning converting to XYZ and using the Vertical Legacy codes. Are we sure the initialization will be performed in all cases?

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
tcobbs-bentley: This check is here to guard against an earlier addon without Vertical CRS support updating the type 66 record in an iModel after tha iModel has been given vertical CRS data. If that happens, this will cause it to ignore the vertical CRS data entirely and simply use the data in the type 66 record. This is detected in two ways:

- If `id` doesn't match, then the type 66 must have been updated
- If `id` does match and the hash doesn't match, then the type 66 must have been updated

Either of the above detections cause the vertical CRS code to ignore the vertical CRS record. We might want to explicitly delete the vertical CRS record in that case, but we don't currently do that. If you feel that this is not the correct behavior, please let me know what it should do instead.

tcobbs-bentley: I'll update the code to skip the `id` check and simply compare the hashes. If they don't match, that means that the type 66 record was updated by an addon without vertical CRS support.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
tcobbs-bentley: Agreed. I’ll retain the id-only WGS84 test as legacy compatibility coverage.

iModelCore/GeoCoord/BaseGeoCoord/basegeocoord.cpp
tcobbs-bentley: Legacy vertical conversions use global CS-Map VERTCON state through `CSvrtconInit` and `CSvrtcon29To88`. Named vertical transforms instead construct their own catalog-backed `csVertconUS_` instance through `CSnewVertconUSFromCatalog` and evaluate it with `CScalcVertconUS`.

The named Vertical CRS meaningful-elevation test exercises this catalog-backed path. A clarifying code comment or more focused test may improve confidence, since this is deleted code, the comment might be confusing. Let me know if you feel that a more focused test is needed.

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
tcobbs-bentley: I created a new Issue to track this and left it with "Needs Triage".

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
tcobbs-bentley: Updates committed.

iModelCore/iModelPlatform/DgnGeoCoord/DgnGeoCoord.cpp
tcobbs-bentley: Fixed.

iModelJsNodeAddon/IModelJsNative.h
tcobbs-bentley: Fixed.

iModelJsNodeAddon/GeoServicesInterop.cpp
tcobbs-bentley: Fixed.

iModelJsNodeAddon/api_package/ts/src/NativeLibrary.ts
tcobbs-bentley: Fixed.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
tcobbs-bentley: Fixed.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
tcobbs-bentley: Fixed.

iModelJsNodeAddon/api_package/ts/src/test/GeoServices.test.ts
tcobbs-bentley: Fixed.

AlainRobertAtBentley (approved):

pmconne (approved):

